PR #15, opened 2016-02-16
NetworkParametersCreator: Create .ihmc config directory if it does not exist

iamwolf:
Currently upon a fresh install on a new machine, one needs to manually run ``mkdir ~/.ihmc`` in order for the NetworkParametersCreator to succeed. This automatically creates the directory if it does not exist without showing a generic error message to the user.


dljsjr:
Thanks @iamwolf!

PR #19, opened 2016-03-04
[Valkyrie] Changed the logic for checking a new ArmTrajectoryPacket

iamwolf:
Don't merge yet, we'd like to keep a PR open though to track conversation and see when new commits come in on the branch.

~~We uncovered another related bug while testing on the robot and will report in a second~~


mauricefallon:
@dljsjr or @SylvainBertrand can you merge this PR? We have just tested this on Val-D and it works fine.

The other issue (https://github.com/ihmcrobotics/ihmc-open-robotics-software/issues/22) was unrelated.

Issue #26, opened 2016-04-01
Only claim required sliderboard channels


wxmerkt:
As bespoken in Slack with @dljsjr:

 * Have a boolean DEBUG or TUNING flag that determines whether all sliderboard channels are mapped or just 7 and 8 (for CoM height and smoothing between standing and walking control), i.e. ``if (DEBUG_OR_TUNING) { claim_all_sliders_and_knobs }``

dljsjr:
@SylvainBertrand This would be nice for Valkyrie; the robot currently cannot stand without using the WalkingSliderBoard and we grab all the channels since this is usually used for tuning. In the case of Valkyrie it would be nice to allow for other stuff to be on the traditionally more diagnostic channels.

dljsjr:
@wxmerkt We're still looking in to automating the control gain transition but we need physical robot access. If the functionality you contributed is satisfying your use cases currently, I can close this issue. Let me know your thoughts.

wxmerkt:
Yeah it's sufficient if it is carried through to the new controller
interface. Happy to close this issue :)

PR #27, opened 2016-04-04
Don't claim all sliders if not in tuning mode

wxmerkt:
This is a first temporary go at it - is there a global variable for tuning? The debug message should give a hint to the user to change the variable if tuning is desired and the sliders are mystically not claimed.

I've seen in other places that sliders might be claimed (e.g. ValkyrieSliderBoard) - is the code there active?

Addresses https://github.com/ihmcrobotics/ihmc-open-robotics-software/issues/26

cc: @dljsjr @SylvainBertrand


dljsjr:
I'd be okay merging this in but I talked about it w/ @SylvainBertrand and Jerry, they want to consider automating the gain ratios/CoM transitions before modifying the common tuning slider board. Code-wise I have no problem with this PR but I want to make sure everybody's on board first.

wxmerkt:
Thanks - let me know what the feedback is or drop me a line on Slack to discuss further if there are any open questions.

PR #70, opened 2016-09-13
Added missing message packet checking

edbot:
FootTrajectoryMessage, GoHomeMessage, PelvisOrientationTrajectoryMessage and PelvisTrajectoryMessage don't check the packet validity although the packet checking routine exist for these messages.

In the case of PelvisTrajectoryMessage, NaN values get passed into the trajectory generator and crash the controller with an exception during execution.

This PR adds the checking routine into the respective message classes.


nmertins:
This looks pretty straight forward but I'll defer to @SylvainBertrand for final judgement before merging.

PR #101, opened 2016-11-18
component-based desired footstep calculator fix

mahopkins:
This pull request includes modifications to the component-based footstep calculator to scale the step length based on step duration in order to track the desired input velocity. It also includes a fix to compute the footstep positions based on the predicted velocity assuming a constant desired heading rate. Note that some related interfaces have been modified.


SylvainBertrand:
Thanks for that!
I merged your branch on Stash.

PR #105, opened 2017-01-12
Print name of joints outside of joint limits in a motion plan

VladimirIvan:
Added a couple of print commands when the trajectory is rejected due to the waypoints being outside of joint limits.


wxmerkt:
@dljsjr further to our slack from yesterday, can we get this one into 0.8.3 as well?

dljsjr:
@wxmerkt Yessir.

PR #141, opened 2018-09-07
changes to get gazebo working

jjholley:
@dljsjr @SylvainBertrand this is a couple of minor changes we made to get val walking correctly in gazebo.  We are interested in getting the gazebo walking in the regular val-develop branch, so I'm passing these changes along to you.

Make sure code is formatted with the relevant template(s) at IHMCJavaToolkit/CodeFormatTemplates.

### DIFF
--- a/valkyrie/src/main/java/us/ihmc/valkyrie/ValkyrieRobotModel.java
+++ b/valkyrie/src/main/java/us/ihmc/valkyrie/ValkyrieRobotModel.java
@@ -142,7 +142,7 @@ public class ValkyrieRobotModel implements DRCRobotModel, SDFDescriptionMutator
       jointMap = new ValkyrieJointMap();
       contactPointParameters = new ValkyrieContactPointParameters(jointMap, simulationContactPoints);
       sensorInformation = new ValkyrieSensorInformation(target);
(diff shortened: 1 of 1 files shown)


valkyrie/src/main/java/us/ihmc/valkyrie/ValkyrieRobotModel.java
```diff
@@ -142,7 +142,7 @@ public ValkyrieRobotModel(RobotTarget target, boolean headless, String model, Fo
       jointMap = new ValkyrieJointMap();
       contactPointParameters = new ValkyrieContactPointParameters(jointMap, simulationContactPoints);
       sensorInformation = new ValkyrieSensorInformation(target);
-      highLevelControllerParameters = new ValkyrieHighLevelControllerParameters(target == RobotTarget.REAL_ROBOT, jointMap);
+      highLevelControllerParameters = new ValkyrieHighLevelControllerParameters(target == RobotTarget.REAL_ROBOT || target == RobotTarget.GAZEBO, jointMap);
```
dljsjr: Like I mentioned before, this is a hack, we have a plan to fix this properly. So we'll probably reject this PR but I'm going to leave it open right now because we need to port over your other commits for the F/T sensors.

jjholley: 👍 
Just getting this on your radar so you are able to reproduce.

PR #142, opened 2018-09-07
added ros1 to ros2 bridge script

jjholley:
just a bash script to make using the ros1 to ros2 bridge easier.


dljsjr:
@jjholley I'm going to merge this in to `develop` instead of `val-develop` because of the current state of the robot branches and how far behind `val-develop` is. In general `val-develop` is only updated by slowly pulling in commits from `develop` to establish stable checkpoints between releases anyway.

PR #197, opened 2024-01-05
Added TransferFunctionDiscretizer and YoFilteredDouble

conconman:
This pull request allows a user to transform a causal transfer function (represented by the ContinuousTransferFunction class) into a set of digital coefficients using a new class, TransferFunctionDiscretizer. The TransferFunctionDiscretizer can solve for any causal transfer function and results in the set of input/output digital filter coefficients using a software friendly form of Tustin's Bilinear Transform. When passed into the YoFilteredDouble, this allows for easy digital filtering to take place behind the scenes.

The YoFilteredDoubleSimulation contains an empty robot simulation with several active filters being passed a chirp signal so that the filtering effects can be seen. The SimplePendulumSimulation is an example where the original PID controller has been cast as a causal transfer function with the same gains and displays steady control behavior.

A paper about this work has been published to arXiv titled "Software Implementation of Digital Filtering via Tustin's Bilinear Transform" (https://arxiv.org/abs/2401.03071) with a bunch of details about the algorithm (which isn't originally mine) and performance.

Example Simulations:
- YoFilteredDoubleSimulation (SCS)
- SimplePendulumSimulation (SCS2 using newer code from ihmc-open-robotics-software-tutorials)

Tests:
- YoFilteredDoubleTest
- TransferFunctionDiscretizerTest
- ContinuousTransferFunctionTest


SylvainBertrand:
Ok I see what's going on, we rewrote the history of the repo at some point and your branch is missing that. I need to figure out how to handle it.

PR #202, opened 2024-06-03
Add Height Map to Perception and Autonomy Process

calvertdw:


ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/RDXHeightMapVisualizer.java
```diff
@@ -68,21 +69,33 @@ public class RDXHeightMapVisualizer extends RDXVisualizer
...
+      activeHeartbeat = new ROS2Heartbeat(ros2, PerceptionAPI.REQUEST_OUSTER_HEIGHT_MAP);
    }
 
    public void setupForImageMessage(ROS2Helper ros2)
    {
       ros2.subscribeViaCallback(PerceptionAPI.HEIGHT_MAP_CROPPED, this::acceptImageMessage);
-      activeHeartbeat = new ROS2Heartbeat(ros2, PerceptionAPI.REQUEST_HEIGHT_MAP);
+   }
+
+   /** @deprecated Use RDXPerceptionVisualizerPanel which supports demand heartbeats.*/
```
calvertdw: @TomaszTB I think this is right?

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
```diff
@@ -246,6 +266,9 @@ public PerceptionAndAutonomyProcess(ROS2Helper ros2Helper,
 
       planarRegionsExtractorThread = new RestartableThrottledThread("PlanarRegionsExtractor", 10.0, this::updatePlanarRegions);
       planarRegionsExtractorThread.start();
+
+      heightMapExtractorThread = new RestartableThrottledThread("HeightMapExtractor", 10.0, this::updateHeightMap);
```
calvertdw: What is the best way to thread this guy? @TomaszTB @ds58 I think in the continuous hiking process it gets triggered on each realsense frame. Should we do that here and how is the best way to do that?

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
```diff
@@ -567,6 +594,51 @@ public void updatePlanarRegions()
...
 
+   public void updateHeightMap()
+   {
+      if (realsenseDepthImage != null && realsenseDepthImage.isAvailable() && heightMapDemandNode.isDemanded())
+      {
+         RawImage latestRealsenseDepthImage = realsenseDepthImage.get();
+
+         if (rapidHeightMapExtractor == null)
+         {
+            rapidHeightMapExtractor = new RapidHeightMapExtractor(openCLManager, null); // TODO: Fix up for resetting to feet Z
```
calvertdw: The only thing the height map needs the reference frames for is to reset the height based on mid feet Z (or I think we might want lowest foot Z) Anyway, it's happy with null and we'll need to add a UI button somewhere to reset it.

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
```diff
@@ -567,6 +594,51 @@ public void updatePlanarRegions()
...
+            rapidHeightMapExtractor.setDepthIntrinsics(latestRealsenseDepthImage.getIntrinsicsCopy());
+            rapidHeightMapExtractor.create(realsenseDepthImage, 1); // TODO: Can't use same image for whole run?
```
calvertdw: @PotatoPeeler3000 The height map extractor wants to use the same depth image for the whole time, but I think @TomaszTB has it so there's a newly allocated RawImage with the data each frame, so we'll need to change that.

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
```diff
@@ -567,6 +594,51 @@ public void updatePlanarRegions()
...
+         }
+
+         ReferenceFrame d455SensorFrame = realsenseFrameSupplier.get(); // TODO: Can we do this in this thread?
+         ReferenceFrame d455ZUpSensorFrame = realsenseZUpFrameSupplier.get(); // TODO: Can we do this in this thread?
```
calvertdw: @TomaszTB @ds58 Is this thread safe? Looks like we do it in other places, seems sketch but maybe it's okay?

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
```diff
@@ -614,6 +687,7 @@ private void initializeDependencyGraph(ROS2PublishSubscribeAPI ros2)
       zedColorDemandNode.addDependents(zedPointCloudDemandNode, centerposeDemandNode, ballDetectionDemandNode);
       planarRegionsDemandNode.addDependents(yoloZEDDemandNode); // Planar region used for door detection
       realsenseDemandNode.addDependents(yoloRealsenseDemandNode);
+      realsenseDemandNode.addDependents(heightMapDemandNode);
```
calvertdw, line 662: @TomaszTB Double check me here

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -587,6 +587,11 @@ public void createSteppabilityMapImages(int height, int width)
       steppabilityConnectionsImage.createOpenCLImage(openCLManager, OpenCL.CL_MEM_READ_WRITE);
    }
 
+   public void destroy()
+   {
+      // TODO What do we need to destroy in here?
```
calvertdw: Yeah this thing never destroyed anything, we need to watch that

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/RDXHeightMapVisualizer.java
TomaszTB: Yeah, seems right to me. I'd have to look at the context of this method to really make sure. Is it simply used to assign a heartbeat to this visualizer? If so, the comment is correct.

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
TomaszTB: Hmm there are many ways to do it. An `ExecutorService` comes to my mind. You can create an executor, and in the realsense thread submit the call to `updateHeightMap` and pass in the depth image, reference frame, or whatever data it requires. That way the processing for the height map doesn't block the realsense thread, and it runs each time you get an image from the realsense. 

You could also make a separate class which handles the `updateHeightMap` method and whatever logic surrounds it if you want to keep the `PerceptionAndAutonomyProcess` class shorter.

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
TomaszTB: Looks correct to me! When the `heightMapDemandNode` is demanded it will also demand the realsense node.

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
TomaszTB: The `RapidHeightMapExtractor` takes a `BytedecoImage` not a `RawImage`. Could it be easier to keep a `BytedecoImage` object as a field and simply update its contents each time we get a new `RawImage`?

ihmc-communication/src/main/java/us/ihmc/communication/PerceptionAPI.java
```diff
@@ -251,7 +251,8 @@ public final class PerceptionAPI
    public static final ROS2Topic<ImageMessage> HEIGHT_MAP_LOCAL = PERCEPTION_MODULE.withOutput().withTypeName(ImageMessage.class).withSuffix("height_map_local");
 
    public static final ROS2Topic<Empty> REQUEST_LIDAR_SCAN = PERCEPTION_MODULE.withSuffix("request_lidar_scan").withType(Empty.class);
-   public static final ROS2Topic<Empty> REQUEST_HEIGHT_MAP = PERCEPTION_MODULE.withSuffix("request_height_map").withType(Empty.class);
+   public static final ROS2Topic<Empty> REQUEST_OUSTER_HEIGHT_MAP = PERCEPTION_MODULE.withSuffix("request_ouster_height_map").withType(Empty.class);
```
PotatoPeeler3000: We only use one height map, can we leave this how it was? This adds complexity when its not needed right now

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/RDXHeightMapVisualizer.java
PotatoPeeler3000: Why are we deprecating stuff? And also what does demand heartbeats do? I ask because if using demand heartbeats is the thing we should be doing I would like to know what that is lol

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
PotatoPeeler3000: Nice! Since this is assumed to be flat ground it won't matter which foot. One of the future tasks I have is to reset the height map to the lowest foot (I'm trying to leave some easy tasks for future interns which is why I haven't done this)

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
PotatoPeeler3000: Why do we need to change that? I'm unsure of the problems that will cause

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
PotatoPeeler3000: Did you add an empty destroy() method?

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
ds58: Just create a BytedecoImage each update to keep it more self contained. I wouldn't want that as a field. This is what I did for planar regions.

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
ds58: I don't immediately see the problem

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
calvertdw: @ds58 Can't do that because the API is different than planar regions. I have to store as a field and copy into it each tick

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
calvertdw: Yep, needs to be filled in sooner than later

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/RDXHeightMapVisualizer.java
```diff
@@ -61,28 +61,31 @@ public class RDXHeightMapVisualizer extends RDXVisualizer
...
-      super("Height Map");
+      this("Height Map");
+   }
+
+   public RDXHeightMapVisualizer(String title)
+   {
+      super(title);
 
-      executorService = Executors.newFixedThreadPool(4);
+      executorService = MissingThreadTools.newSingleThreadExecutor("Height Map Visualizer Subscription", true, 1);
```
calvertdw, line 73: This was used to stop a bunch of execution rejected exceptions and is the better strategy anyways.

ds58 (approved):

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
```diff
@@ -644,12 +682,16 @@ public static void main(String[] args)
       ROS2Node ros2Node = ROS2Tools.createROS2Node(CommunicationMode.INTERPROCESS.getPubSubImplementation(), "perception_autonomy_process");
       ROS2Helper ros2Helper = new ROS2Helper(ros2Node);
 
+      RapidHeightMapExtractor.reloadHeightMapParameters("GPUBuildingExploration"); // FIXME: This should be redesigned.
```
calvertdw, line 685: @PotatoPeeler3000 This was the only way to load a different parameter set unfortunately. Why the heck are they statically accessed everywhere :cry:

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
```diff
@@ -644,12 +682,16 @@ public static void main(String[] args)
       ROS2Node ros2Node = ROS2Tools.createROS2Node(CommunicationMode.INTERPROCESS.getPubSubImplementation(), "perception_autonomy_process");
       ROS2Helper ros2Helper = new ROS2Helper(ros2Node);
 
+      RapidHeightMapExtractor.getHeightMapParameters().setCropWindowSize(501);
```
calvertdw, line 685: Okay actually I think this is a better strategy where the parameters don't have to get duplicated.

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/RDXHeightMapVisualizer.java
TomaszTB: Doesn't this reduce the number of threads for this executor from 4 to 1?

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
PotatoPeeler3000, line 1: All the changes in this file don't seem relevant to this PR, could this be left out? In the future we aren't going to use the midFeetUnderPelvisFrame.

ihmc-perception/src/main/java/us/ihmc/perception/headless/HumanoidPerceptionModule.java
```diff
@@ -280,7 +280,7 @@ public void initializePerspectiveRapidRegionsExtractor(CameraIntrinsics cameraIn
    public void initializeHeightMapExtractor(HumanoidReferenceFrames referenceFrames, CameraIntrinsics cameraIntrinsics)
    {
       LogTools.info("Rapid Height Map: {}", cameraIntrinsics);
-      rapidHeightMapExtractor = new RapidHeightMapExtractor(openCLManager, referenceFrames);
+      rapidHeightMapExtractor = new RapidHeightMapExtractor(openCLManager, referenceFrames.getMidFeetUnderPelvisFrame());
```
PotatoPeeler3000: It seems like this doesn't need to be changed for this PR, can we leave this out?

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
TomaszTB: Probably should fill this out before merging, right?

ihmc-perception/src/main/java/us/ihmc/perception/headless/HumanoidPerceptionModule.java
calvertdw: It was needed because we don't have an instance of HumanoidReferenceFrames to pass in, only the single frame. Since you only use that frame I didn't think it would be an issue.

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
calvertdw: Can you guys help me find stuff to destroy? I'm not sure what would need to be destroyed

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/RDXHeightMapVisualizer.java
calvertdw: Yes, which is what you want, right? What's the point in having more than one here?

ihmc-perception/src/main/java/us/ihmc/perception/headless/HumanoidPerceptionModule.java
calvertdw: The reason you don't always have a HumanoidReferenceFrames is that PerceptionAndAutonomyProcess supports being run standalone in order to test sensor drivers and algorithms for development and debugging. In this case all the frames are just world frame and there is no robot model with which to create a HumanoidReferenceFrames.

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
calvertdw: See my other comments https://github.com/ihmcrobotics/ihmc-open-robotics-software/pull/202#discussion_r1628206361

ihmc-perception/src/main/java/us/ihmc/perception/headless/HumanoidPerceptionModule.java
PotatoPeeler3000: We don't use that frame though, you've switch the reference frames, its meant to be midZUpFrame, not the under pelvis one

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/RDXHeightMapVisualizer.java
PotatoPeeler3000: It might be good to leave it how it was, is there a problem with having 4? Does that cause problems that make us want to use 1?

PotatoPeeler3000:
Also how do you get the reset height map button to work? In the SimulationUI its not working for me

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/RDXHeightMapVisualizer.java
calvertdw: Yes, there were rejected execution exceptions blowing up the console

calvertdw:
> Also how do you get the reset height map button to work? In the SimulationUI its not working for me

@PotatoPeeler3000 The RapidHeightMapManager implements the subscriber. The RapidHeightMapExtractor does not have any networking in it. So if you want it, you'd have to add the line in HumanoidPerceptionModule.

calvertdw:
@PotatoPeeler3000 I fixed that
Peek 2024-06-05 16-40

PotatoPeeler3000:
Cool this looks good! The reset to lowest food is nice too

PotatoPeeler3000:
Well I can't find the approve button lmao what

PotatoPeeler3000 (approved):

TomaszTB (approved):
Assuming the `destroy()` method didn't break anything, I approve!

PR #205, opened 2024-06-05
Feature/stance generation - Open Robotics

PotatoPeeler3000:
- Add stance panel to the UI
- Update StancePoseCalculator to get the best possible stances from a simple cost function
- Pass the stance pose into the Continuous Hiking thread so it can walk to that goal
- Update the ContinuousPlannerSchedulingTask to allow for walking to a goal rather then just straight forward
- Other cleanup


ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/polygonSnapping/PolygonSnapperTools.java
PotatoPeeler3000, line 1: Removed duplicated code

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
```diff
@@ -118,59 +114,16 @@ public RDXContinuousHikingPanel(ROS2Helper ros2Helper,
          defaultFoothold.update();
          defaultContactPoints.put(robotSide, defaultFoothold);
       }
-      stancePoseCalculator = new StancePoseCalculator(0.5f, 0.5f, 0.1f, defaultContactPoints);
+      StancePoseCalculator stancePoseCalculator = new StancePoseCalculator(defaultContactPoints);
+      stancePoseSelectionPanel = new RDXStancePoseSelectionPanel(ros2Helper, stancePoseCalculator);
+      addChild(stancePoseSelectionPanel);
```
PotatoPeeler3000, line 119: Add the stancePoseSelectionPanel to the UI

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
```diff
@@ -118,59 +114,16 @@ public RDXContinuousHikingPanel(ROS2Helper ros2Helper,
...
-                                                                    robotModel.getContactPointParameters().getControllerFootGroundContactPoints());
-
-      ros2Helper.subscribeViaCallback(ContinuousWalkingAPI.START_AND_GOAL_FOOTSTEPS, this::onStartAndGoalPosesReceived);
-      ros2Helper.subscribeViaCallback(ContinuousWalkingAPI.PLANNED_FOOTSTEPS, this::onPlannedFootstepsReceived);
-      ros2Helper.subscribeViaCallback(ContinuousWalkingAPI.MONTE_CARLO_FOOTSTEP_PLAN, this::onMonteCarloPlanReceived);
-      ros2Helper.subscribeViaCallback(HumanoidControllerAPI.getTopic(WalkingControllerFailureStatusMessage.class, robotModel.getSimpleRobotName()), message ->
-      {
-         terrainPlanningDebugger.reset();
-      });
+                                                               robotModel.getContactPointParameters().getControllerFootGroundContactPoints());
```
PotatoPeeler3000, line 123: GitHub does a terrible job but there were 2 constructors for this class and I combined them into one

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
```diff
@@ -239,6 +192,11 @@ public void renderImGuiWidgets()
       publishInputCommandMessage();
    }
 
+   public void processImGui3DViewInput(ImGui3DViewInput input)
+   {
+      stancePoseSelectionPanel.processImGui3DViewInput(input);
```
PotatoPeeler3000, line 197: Pass the mouse input down to the calculator

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXStancePoseSelectionPanel.java
```diff
@@ -161,13 +169,32 @@ else if (dScroll < 0.0)
...
    {
-      placed = true;
+      List<Pose3D> poses = new ArrayList<>();
+      poses.add(new Pose3D(stancePoses.get(RobotSide.LEFT)));
+      poses.add(new Pose3D(stancePoses.get(RobotSide.RIGHT)));
+
+      PoseListMessage poseListMessage = new PoseListMessage();
+      MessageTools.packPoseListMessage(poses, poseListMessage);
+
+      ros2Helper.publish(ContinuousWalkingAPI.PLACED_GOAL_FOOTSTEPS, poseListMessage);
```
PotatoPeeler3000, line 181: Once we have the poses, we send it to the thread via this topic

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXStancePoseSelectionPanel.java
```diff
@@ -178,11 +205,17 @@ public void renderImGuiWidgets()
...
+      ImGui.text("World Point: " + latestPickPoint.getTranslation().toString("%.3f"));
       if (terrainMapData != null)
       {
-         ImGui.text("Height: " + terrainMapData.getHeightInWorld(latestPose.getTranslation().getX32(), latestPose.getTranslation().getY32()));
-         ImGui.text("Contact Score: " + terrainMapData.getContactScoreInWorld(latestPose.getTranslation().getX32(), latestPose.getTranslation().getY32()));
+         ImGui.text("Height: " + terrainMapData.getHeightInWorld(latestPickPoint.getTranslation().getX32(), latestPickPoint.getTranslation().getY32()));
+         ImGui.text("Contact Score: " + terrainMapData.getContactScoreInWorld(latestPickPoint.getTranslation().getX32(), latestPickPoint.getTranslation().getY32()));
+      }
+
+      ImGui.checkbox(labels.get("Calculate Stance Pose"), calculateStancePose);
```
PotatoPeeler3000, line 215: Render parameters to the UI for tuning, once its tuned its fine but nice to have available

ihmc-high-level-behaviors/src/main/generated-java/us/ihmc/behaviors/activeMapping/StancePoseParametersBasics.java
PotatoPeeler3000, line 1: Cause we love StoredPropertySets!!!!!!!!!!

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanner.java
```diff
@@ -326,6 +330,21 @@ public void setGoalWaypointPoses()
                                                                    (float) continuousHikingParameters.getGoalPoseForwardDistance(),
                                                                    xRandomMargin,
                                                                    (float) continuousHikingParameters.getGoalPoseUpDistance(), nominalStanceWidth);
+         case WALK_TO_GOAL:
```
PotatoPeeler3000, line 337: Allows for Continuous Hiking to walk to a goal as well

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanner.java
```diff
@@ -610,4 +629,20 @@ public void requestMonteCarloPlannerReset()
    {
       resetMonteCarloFootstepPlanner = true;
    }
+
+   public void addWayPointToList(Pose3D leftFootGoalPose, Pose3D rightFootGoalPose)
```
PotatoPeeler3000, line 633: Not only can we walk to a goal, we can add waypoints which it will walk too in the order they were placed, still requires holding control to walk forward though

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanningTools.java
```diff
@@ -18,6 +21,22 @@
 
 public class ContinuousPlanningTools
 {
+   public static double getDistanceFromRobotToGoalPoseOnXYPlane(Vector3DBasics robotPositionInWorld, SideDependentList<FramePose3D> goalPoses)
```
PotatoPeeler3000: Need to know how far away we are from a goal pose "waypoint" to see if we should start going to wards the next one or not. Maybe there is a better method that already does this

PotatoPeeler3000 (commented):
Gets the job done but isn't perfect, there are still bugs around which won't be fixed in this PR

PotatoPeeler3000:
[Uploading Screencast from 06-04-2024 02:51:01 PM.webm…]()

PotatoPeeler3000:
Currently this thing wants to only go forward because of the way the footstep parameters are tuned. Will address that in a future PR after talking with Stephen about reducing the number of parametes

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXStancePoseSelectionPanel.java
```diff
@@ -6,53 +6,72 @@
...
-   private ArrayList<ModelInstance> rightSpheres = new ArrayList<>();
+   private final ArrayList<ModelInstance> leftSpheres = new ArrayList<>();
+   private final ArrayList<ModelInstance> rightSpheres = new ArrayList<>();
 
-   private SideDependentList<FramePose3D> stancePoses = new SideDependentList<>();
-   private final FramePose3D latestPose = new FramePose3D(ReferenceFrame.getWorldFrame());
-   private final SideDependentList<RDXFootstepGraphic> footstepGraphics;
+   public SideDependentList<FramePose3D> stancePoses = new SideDependentList<>();
+   private final FramePose3D latestPickPoint = new FramePose3D(ReferenceFrame.getWorldFrame());
+   public final SideDependentList<RDXFootstepGraphic> footstepGraphics;
```
rjgriffin42: are tehse meant to be public?

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXStancePoseSelectionPanel.java
```diff
@@ -65,35 +84,25 @@ public RDXStancePoseSelectionPanel(StancePoseCalculator stancePoseCalculator)
...
-         selectionActive = false;
-      }
+      
       updatePoses();
    }
 
    private void updatePoses()
    {
-      if (environmentHandler.hasTerrainMapData())
+      if (environmentHandler.hasTerrainMapData() && environmentHandler.hasHeightMap())
```
rjgriffin42, line 96: is there ever a case where you won't have height map data if you have terrain map data? should this be an error thrown in the get height in world method?

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXStancePoseSelectionPanel.java
```diff
@@ -106,8 +115,8 @@ private void updatePoses()
             {
                for (int i = 0; i < leftPoses.size(); i++)
                {
-                  leftSpheres.add(RDXModelBuilder.createSphere(0.015f, Color.BLUE));
-                  rightSpheres.add(RDXModelBuilder.createSphere(0.015f, Color.RED));
+                  leftSpheres.add(RDXModelBuilder.createSphere(0.01f, Color.WHITE));
+                  rightSpheres.add(RDXModelBuilder.createSphere(0.01f, Color.BLACK));
```
rjgriffin42, line 118: should colors match the red/green we use for left/right?

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXStancePoseSelectionPanel.java
```diff
@@ -6,53 +6,72 @@
...
-   private ArrayList<ModelInstance> leftSpheres = new ArrayList<>();
-   private ArrayList<ModelInstance> rightSpheres = new ArrayList<>();
+   private final ArrayList<ModelInstance> leftSpheres = new ArrayList<>();
+   private final ArrayList<ModelInstance> rightSpheres = new ArrayList<>();
 
-   private SideDependentList<FramePose3D> stancePoses = new SideDependentList<>();
-   private final FramePose3D latestPose = new FramePose3D(ReferenceFrame.getWorldFrame());
-   private final SideDependentList<RDXFootstepGraphic> footstepGraphics;
+   public SideDependentList<FramePose3D> stancePoses = new SideDependentList<>();
+   private final FramePose3D latestPickPoint = new FramePose3D(ReferenceFrame.getWorldFrame());
```
rjgriffin42, line 52: does this name convention change imply that it will always be selected? How does the pose get set programmatically (like from a message)?

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanner.java
rjgriffin42: do you know what frontier expansion and active search modes do?

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanner.java
```diff
@@ -81,6 +83,8 @@ public enum PlanningMode
    private boolean active;
    private double previousContinuousHikingSwingTime = 0.0;
 
+   private List<SideDependentList<Pose3D>> walkToGoalWayPointList = new ArrayList<>();
+
    float xRandomMargin = 0.2f;
    float nominalStanceWidth = 0.22f;
```
rjgriffin42, line 92: isn't this a property of the robot and out of scope for this class?

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanningTools.java
rjgriffin42: Pass in robotPositionInWorld as a Point2DReadOnly or a Point3DReadOnly. Then you don't need lines 34 and 35. You can then use the `distanceXY` method, instead of `distance.`

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanningTools.java
```diff
@@ -18,6 +21,22 @@
...
+   {
+      FramePose3D leftGoalPose = goalPoses.get(RobotSide.LEFT);
+      FramePose3D rightGoalPose = goalPoses.get(RobotSide.RIGHT);
+
+      // Get point halfway between the left and right goal poses
+      leftGoalPose.interpolate(rightGoalPose, 0.5);
+      FramePose3D middleDistanceBetweenGoalPoses = leftGoalPose;
+      middleDistanceBetweenGoalPoses.getPosition().setZ(0.0);  // We want to check distance in XY Plane so zero the Z
```
rjgriffin42: these few lines are actually pretty problematic, I think. You're changing leftGoalPose Pose for no real reason, which I'm guessing is going to mess up some of the extgernal stuff.

I would change to:

```
Point3D middleDistanceBetweenGoalPoses = new Point3D();
middleDistanceBetweenGoalPoses.interpolate(leftGoalPose.getPosition(), rightGoalPose.getPosition(), 0.5);
```

rjgriffin42: then setting the height to zero is irrelevant if you use `distanceXY` instead of `distance`

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
```diff
@@ -3,57 +3,43 @@
...
    {
-      this.maxWidth = maxWidth;
-      this.maxLength = maxLength;
-      this.maxYaw = maxYaw;
       this.footPolygons = footPolygons;
       this.heightMapPolygonSnapper = new HeightMapPolygonSnapper();
    }
 
-   public SideDependentList<FramePose3D> getStancePoses(FramePose3D goalPose, TerrainMapData terrainMap, FootstepPlannerEnvironmentHandler environmentHandler)
+   public SideDependentList<FramePose3D> getStancePoses(FramePose3D pickPoint, TerrainMapData terrainMap, FootstepPlannerEnvironmentHandler environmentHandler)
```
rjgriffin42: I wouldn't call this a pickpoint, I'd call it the `midstancePose` or something like that. PickPoint implies this only comes form selection.

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
```diff
@@ -71,13 +57,14 @@ public void insertCandidatePoses(ArrayList<FramePose3D> poses, FramePose3D goalP
       poses.clear();
```
rjgriffin42: makes more sense to call `populateCandidatePoses` and to rename poses as `posesToPack` so it's clear it's a list getting populated. 

Also, can change goalPose to `FramePose3DReadOnly`

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
```diff
@@ -87,10 +74,10 @@ public void insertCandidatePoses(ArrayList<FramePose3D> poses, FramePose3D goalP
       }
    }
 
-   public void searchForOptimalGoalStance(ArrayList<FramePose3D> leftPoses, ArrayList<FramePose3D> rightPoses, FramePose3D goalPose, TerrainMapData terrainMap)
+   public void searchForOptimalGoalStance(ArrayList<FramePose3D> leftPoses, ArrayList<FramePose3D> rightPoses, FramePose3D pickPoint, TerrainMapData terrainMap)
```
rjgriffin42: You can make pickpoint (one: a new name), and then make it `FramePose3DReadOnly`

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
```diff
@@ -99,16 +86,21 @@ public void searchForOptimalGoalStance(ArrayList<FramePose3D> leftPoses, ArrayLi
...
-            double contactCostRight = Math.abs(MAX_CONTACT_VALUE - terrainMap.getContactScoreInWorld(rightPose.getPosition().getX32(), rightPose.getPosition().getY32()));
+            double contactCostLeft = Math.abs(stancePoseParameters.getMaxContactValue() - terrainMap.getContactScoreInWorld(leftPose.getPosition().getX32(), leftPose.getPosition().getY32()));
+            double contactCostRight = Math.abs(stancePoseParameters.getMaxContactValue() - terrainMap.getContactScoreInWorld(rightPose.getPosition().getX32(), rightPose.getPosition().getY32()));
 
-            cost = Math.abs(NOMINAL_STANCE_DISTANCE - leftPose.getPositionDistance(rightPose));
-            cost += CONTACT_COST_WEIGHT * (contactCostLeft + contactCostRight);
+            cost = Math.abs(stancePoseParameters.getNominalStanceDistance() - leftPose.getPositionDistance(rightPose));
+            cost += Math.abs((stancePoseParameters.getNominalStanceDistance() / 2) - leftPose.getPositionDistance(pickPoint));
+            cost += Math.abs((stancePoseParameters.getNominalStanceDistance() / 2) - rightPose.getPositionDistance(pickPoint));
```
rjgriffin42: don't you want the distance to be just an XY distance? Otherwise this also has height changes incorporated, which skews it. I'd add a separate cost on the height delta between the  two snapped feet.

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
```diff
@@ -99,16 +86,21 @@ public void searchForOptimalGoalStance(ArrayList<FramePose3D> leftPoses, ArrayLi
...
+            cost += stancePoseParameters.getContactCostWeight() * (contactCostLeft + contactCostRight);
 
             leftPose.setZ(heightLeft);
             rightPose.setZ(heightRight);
 
-            if (cost < minCost)
+            // In order to not jump back and forth between two good poses, only accept a pose if its at least epsilon better for the cost
+            double epsilon = 0.02;
+            double difference = Math.abs(cost - minCost);
+            if (cost < minCost && difference > epsilon)
```
rjgriffin42: this could instead be written as `cost < minCost - epsilon`, which would be cleaner and do the same thing.

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
rjgriffin42: This also inherently doesn't consider partial support, does it?

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
rjgriffin42: You can also make this a `FramePose3DReadOnly`, which I think would be a lot safer. Additionally, it's highly likely the returns could be `SideDependentList<? extends FramePose3DReadOnly>`

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanner.java
calvertdw: Are you able to change the mode via the API? Is the intention that the mode is assumed to get set on construction and never change? Otherwise, it seems like the mode could dynamically change based on the user's request.

ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/communication/ContinuousWalkingAPI.java
```diff
@@ -28,6 +28,7 @@ public class ContinuousWalkingAPI
    public static final ROS2Topic<ContinuousWalkingCommandMessage> CONTINUOUS_WALKING_COMMAND = IHMC_ROOT.withModule("continuous_walking").withType(ContinuousWalkingCommandMessage.class).withSuffix("command");
    public static final ROS2Topic<ContinuousWalkingStatusMessage> CONTINUOUS_WALKING_STATUS = IHMC_ROOT.withModule("continuous_walking").withType(ContinuousWalkingStatusMessage.class).withSuffix("status");
 
+   public static final ROS2Topic<PoseListMessage> PLACED_GOAL_FOOTSTEPS = IHMC_ROOT.withModule("continuous_walking").withType(PoseListMessage.class).withSuffix("placed_goal_footsteps");
```
calvertdw, line 31: Why not include this in the CONTINUOUS_WALKING_COMMAND? There's joystick inputs in there too, this is another form of telling it where to go. Would be nice for the user reading the API to see all the options in one place.

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanner.java
PotatoPeeler3000: Currently frontier expansion and active search don't do anything, I believe that was part of a future work Bhavyansh wanted to do.

PotatoPeeler3000: I don't know about changing the mode via the API

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXStancePoseSelectionPanel.java
PotatoPeeler3000: I'm not sure, in this case where `updatePoses()` is called we are setting both the terrain map and the height map so I think it makes sense to check that its not null

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXStancePoseSelectionPanel.java
PotatoPeeler3000: They were red and green but I found that when the height map is one of those colors the grid's are really difficult to see. I don't have a preference on the color just wanted to make it obvious to see

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanningTools.java
PotatoPeeler3000: Ya thats nice, passed in a Point3DReadOnly and removed those lines

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXStancePoseSelectionPanel.java
PotatoPeeler3000: No, but there is a warning when its made private but thats fine for now, made them private

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanningTools.java
PotatoPeeler3000: True thats very good

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
PotatoPeeler3000: Yep all good points

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
PotatoPeeler3000: Like that too, nice

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
PotatoPeeler3000: Ya I've changed the distance to be just XY and then added a cost value for Z. Right now it doesn't consider partial support inherently

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
PotatoPeeler3000: Also true

ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/communication/ContinuousWalkingAPI.java
PotatoPeeler3000: I didn't want to mess with the CONTINUOUS_WALKING_COMMAND right now. Isn't everything contained within  ContinuousWalkingAPI relavent to the API? What makes just the CONTINUOUS_WALKING_COMMAND the entire API?

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXStancePoseSelectionPanel.java
PotatoPeeler3000: Its only selected when the RDXStanceSelectionPanel is active, the latestPickPoint pose gets set programmatically from the pickPointSphere, and when the left click on the mouse happens, its sent as a message to the ContinuousHiking process

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanner.java
PotatoPeeler3000: I'm not sure if its a property of the robot, I agree it shouldn't be in this class, will have to think about where this should live. Some how I'm looking for a way to tell ContinuousHiking to move on from one goal waypoint and onto the next one if it exists. That allows for clicking multiple goals like waypoints and going to all of them, then it will be easy to make a patrol feature.

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/StancePoseCalculator.java
PotatoPeeler3000: Couldn't get the method to return SideDependentList<? extends FramePose3DReadOnly>

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousPlanner.java
PotatoPeeler3000: I'm planning on moving this up a class, or some how creating a remote continuous hiking manager (not really sure yet) to handle things like this. Like what ever class started the process may want to keep track of "waypoints" and then ContinuousHiking just handles the immediate goal, for example if I want to add a Patrol feature, somehow I have to remember the start and end, and go back and forth. A higher level manager could track the start and goal, and check the distance, and when ever something new needs to be done, pass that down to the ContinuousHiking Module, open to ideas about this

PR #208, opened 2024-06-05
Filled the destroy method with code

TomaszTB:
Filled destroy method! This should be tested.


ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -588,7 +588,37 @@ public void createSteppabilityMapImages(int height, int width)
 
    public void destroy()
    {
-      // TODO What do we need to destroy in here?
+      if (cropWindowRectangle != null) // this indicates everything was initialized, and should be destroyed
```
TomaszTB: I thought it'd be easier to check one variable to see whether everything was initialized. Otherwise a null check on every variable could work.

calvertdw (approved):

PR #210, opened 2024-06-06
Support Kinematics Simulation in SCS 2 Factory

calvertdw:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/scs2/SCS2AvatarSimulationFactory.java
```diff
@@ -332,10 +334,35 @@ private void setupYoVariableServer()
...
+   private SCS2JointDesiredOutputWriterFactory getDefaultOutputWriterFactory()
+   {
+      return (controllerInput, controllerOutput) ->
+      {
+         boolean writeBeforeEstimationTick = true;
+         if (kinematicsSimulation.get())
+            return new SCS2KinematicsSimulationOutputWriter(controllerInput, controllerOutput, writeBeforeEstimationTick);
+         else
+            return new SCS2OutputWriter(controllerInput, controllerOutput, writeBeforeEstimationTick);
+      };
```
calvertdw, line 346: I added this since you last looked to make it easier to use the new feature. This just sets the default output writer appropriately, but the user is still free to set the outputwriter to something custom.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/scs2/SCS2AvatarSimulationFactory.java
```diff
@@ -332,10 +334,35 @@ private void setupYoVariableServer()
...
+
+   private void setupKinematicsSimulationOutputWriter()
+   {
+      if (kinematicsSimulation.get() && simulationOutputWriter instanceof SCS2KinematicsSimulationOutputWriter kinematicsSimulationOutputWriter)
+      {
+         RootJointDesiredConfigurationDataReadOnly outputForRootJoint = highLevelHumanoidControllerFactory.get()
+                                                                                                          .getWholeBodyControllerCoreFactory()
+                                                                                                          .getWholeBodyControllerCore()
+                                                                                                          .getOutputForRootJoint();
+         kinematicsSimulationOutputWriter.setOutputForRootJoint(outputForRootJoint);
```
calvertdw, line 364: This has to be done later so had to be a separate setup method.

SylvainBertrand (approved):
I wish it didn't have to be that intrusive into the controller, but I don't have a good counter proposal for now.
Other than that, it looks good to me

rjgriffin42:
I was about to make basically the same comment as Sylvain, in that I really hate the idea of injecting knowledge of simulation into the controller, but am not going to let my fundamental distaste for that in all parts of the code base be a blocker.

calvertdw:
If it helps, I don't like it either, but I think you'd have to make another factory or something.

PR #213, opened 2024-06-07
Cleanup/footstep parameters

PotatoPeeler3000:
Changes in this PR:
- Removed 10 parameters shown in image below
- Organized the setters, getters, and .json files so they all appear in the same order
- Renamed to DefaultFootStepPlannerParameters
- Added all the keys to the parameters file because it makes sense (its how we do it in other places)


ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/graphSearch/parameters/DefaultFootstepPlannerParameters.java
```diff
@@ -1,17 +1,91 @@
 package us.ihmc.footstepPlanning.graphSearch.parameters;
 
+import us.ihmc.tools.property.BooleanStoredPropertyKey;
+import us.ihmc.tools.property.DoubleStoredPropertyKey;
+import us.ihmc.tools.property.IntegerStoredPropertyKey;
+import us.ihmc.tools.property.StoredPropertyKeyList;
 import us.ihmc.tools.property.StoredPropertySet;
 
-public class DefaultFootstepPlannerParameters extends StoredPropertySet implements FootstepPlannerParametersBasics
+public class DefaultFootstepPlannerParameters extends StoredPropertySet implements DefaultFootstepPlannerParametersBasics
```
PotatoPeeler3000, line 9: I added the keys to this file, I found it confusing to put the parameters in a different file when this file is named DefaultFootStepPlannerParameters

PotatoPeeler3000 (commented):
image

PotatoPeeler3000:
Removed 10 parameters from the planner, tests still pass, when walking in sim I don't see any real difference when planning around and walking to goals. Did change the name to Default... because I was constantly getting confused between what was Default and not

stephenmcc (approved):

ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/graphSearch/stepChecking/FootstepPoseHeuristicChecker.java
```diff
@@ -146,64 +143,24 @@ else if (Math.abs(stepHeight.getValue()) > maximumStepZ)
...
-      maxXFromPitchContraint.set(InterpolationTools.linearInterpolate(Math.abs(parameters.getMaximumStepReach()), parameters.getMaximumStepXWhenFullyPitched(), alphaPitchedBack));
-      double stepDownFraction = - stepHeight.getValue() / minZFromPitchContraint.getValue();
-      double stepForwardFraction = stepLength.getValue() / maxXFromPitchContraint.getValue();
+      double alphaPitchedBack = Math.max(0.0, - stanceFootPose.getPitch() / parameters.getMinSurfaceIncline());
 
       stepIsPitchedBack.set(alphaPitchedBack > 0.0);
-      stepTooLow.set(stepLength.getValue() > 0.0 && stepDownFraction > 1.0);
-      stepTooForward.set(stepHeight.getValue() < 0.0 && stepForwardFraction > 1.0);
+      stepTooLow.set(stepLength.getValue() > 0.0 && stepHeight.getValue() > 1.0);
+      stepTooForward.set(stepHeight.getValue() < 0.0 && stepLength.getValue() > 1.0);
```
rjgriffin42: do these do anything anymore? As written, "stepTooLow" is now only true if you're stepping forward and up a meter in the air, which wasn't the intent of that. Also stepTooForward is a similar thing. So this should be either fixed, or these fields should be deleted and the `STEP_TOO_LOW_AND_FORWARD_WHEN_PITCHED` check should be removed.

PR #220, opened 2024-06-12
FrequencyCalculator fixes/optimization

ds58:
FrequencyCalculator was pretty inaccurate. It also did not update when you queried the frequency, which isn't the desired behavior, in my opinion.

I wrote new tests for it and also Throttler. These tests reveal Throttler is also inaccurate.


ihmc-java-toolkit/src/main/java/us/ihmc/tools/time/FrequencyCalculator.java
```diff
@@ -1,59 +1,78 @@
...
 
-      double totalElapsed = 0.0;
-      for (Double delta : deltas)
-      {
-         totalElapsed += delta;
-      }
-      frequency = UnitConversions.secondsToHertz(totalElapsed / deltas.size());
+      pingCount++;
+      elapsedNanos += currentTime - lastPingTimeNanos;
+      lastPingTimeNanos = currentTime;
```
SylvainBertrand: The previous approach would do the average over a moving window while this new version does it seems the average over the entire run, which I'd argue is not desirable. I think using a moving window is desired as it discards the initial JIT compilation slowdowns rapidly and captures if they are changes in the code being ran. When running for 10min, if the monitored code running changes for more or less computing, it'll take several minutes before being able to identify the change,

SylvainBertrand (changes requested):
I'm suggesting re-introducing the use of a moving window when computing the average.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/time/FrequencyCalculator.java
ds58: I reset the `elapsedNanos` every 500ms, so in this way it is similar to the rolling average. Would you suggest I not do that and reintroduce an additional data structure like before?

ds58:
Oops didn't mean to re-request a review from you @SylvainBertrand

ihmc-java-toolkit/src/main/java/us/ihmc/tools/time/FrequencyCalculator.java
ds58: I changed the implementation, let me know what you think about it

ds58:
There is some problem with Throttler that maybe I should try and address as well. Test results:

| method                   | desired hz    | actual hz
| -------------            | ------------- | ---------		|
| ThreadTools.sleep()      | 0.5		   | 0.4999			|
| ThreadTools.sleep()      | 1  		   | 0.9998			|
| ThreadTools.sleep()      | 10  		   | 9.9832         |
| ThreadTools.sleep()      | 100  		   | 99.118			|
| ihmc Throttler           | 0.5  		   | 0.7503			|
| ihmc Throttler           | 1  		   | 1.2510			|
| ihmc Throttler           | 10  		   | 9.9852         |
| ihmc Throttler           | 100  		   | 99.076			|

ihmc-java-toolkit/src/main/java/us/ihmc/tools/time/FrequencyCalculator.java
```diff
@@ -1,59 +1,96 @@
...
+      int pings = pingTimes.size();
 
-      while (deltas.size() > history)
+      if (first != null && last != null && pings > 1)
       {
-         deltas.removeFirst();
+         long elapsedNanos = last - first;
+         long elapsedNanosAverage = elapsedNanos / (pings - 1);
+         double elapsedSecondsAverage = elapsedNanosAverage / NANOS_IN_A_SECOND;
+         return UnitConversions.secondsToHertz(elapsedSecondsAverage);
```
TomaszTB, line 65: Coud do 
```
         long elapsedNanos = last - first;
         double elapsedSeconds = Conversions.nanosecondsToSeconds(elapsedNanos);
         double elapsedSecondsAverage = elapsedSeconds / (pings - 1);
         return UnitConversions.secondsToHertz(elapsedSecondsAverage);
```
Just to avoid the integer division of `elapsedNanos / (pings - 1)`. I'd only make the change if you want to be super pedantic.

TomaszTB (approved):
Looks great and works well!

ihmc-java-toolkit/src/main/java/us/ihmc/tools/time/FrequencyCalculator.java
ds58: Good idea, I've applied that change

ihmc-java-toolkit/src/main/java/us/ihmc/tools/time/FrequencyCalculator.java
```diff
@@ -1,59 +1,97 @@
...
-      double totalElapsed = 0.0;
-      for (Double delta : deltas)
+      return 0.0;
+   }
+
+   public void ping() {
+      pingTimes.add(System.nanoTime());
+
+      double frequency = calculateFrequency(false);
+      while (frequency > 0.0 && pingTimes.size() > (frequency * 10))
```
SylvainBertrand, line 75: That `fequency * 10` threshold is throwing me off, I don't really make sense of it. But if it works for you, then go for it

SylvainBertrand (approved):
Feel free to merge. I added a comment that there's something that I don't quite get, but it's fine if that works

ihmc-java-toolkit/src/main/java/us/ihmc/tools/time/FrequencyCalculator.java
ds58: It was basically a way to get a dynamically sized history. I didn't want to have to specify the history size when I think there is some relationship between frequency and desired history size.

A scalar of 10 seemed good to me. For example consider these values
| freq (hz) | history size |
|-----------|--------------|
| 1         | 10           |
| 10        | 100          |
| 100       | 1000         |

PR #221, opened 2024-06-12
Kick door message

christiandebuys:
Adds a kick door message to the controller and in RDX land.


ds58 (approved):

rjgriffin42:
We should test to make sure the existing behaviors still function on the robot. Once that's done, and merge conflicts are resolved, I'm fine with merging.

PR #222, opened 2024-06-13
Feature/quicker mpc stepping new

Beomyeong:
Adding some tools for MPC-based touchdown calculator


rjgriffin42:
You should add some tests in here before this gets merged. They would be very simple.

SylvainBertrand (changes requested):
I'd suggest adding the tests for these new methods.

Beomyeong:
> simple
I added two tests: one for MatrixMissingTools.power and one for MatrixMissingTools.elementWiseLessThan.

Beomyeong:
> I'd suggest adding the tests for these new methods.

I added two tests: one for MatrixMissingTools.power and one for MatrixMissingTools.elementWiseLessThan.

ihmc-robotics-toolkit/src/test/java/us/ihmc/robotics/MatrixMissingToolsTest.java
```diff
@@ -313,4 +313,61 @@ public void testSetSelectedMatrixDiagonals()
...
+         DMatrixRMaj matrixCompareTemporary = new DMatrixRMaj(matrixSize, matrixSize);
+         DMatrixRMaj matrixCompareResult = new DMatrixRMaj(matrix);
+
+         if (powerNumber > 1)
+         {
+            for (int j = 0; j < powerNumber - 1; j++)
+            {
+               CommonOps_DDRM.mult(matrix, matrixCompareResult, matrixCompareTemporary);
+               matrixCompareResult.set(matrixCompareTemporary);
+            }
```
SylvainBertrand, line 357: This is not really a test because you re-implemented the method here and compare it against the method, besides testing that the method runs properly it doesn't do much more.
To test I suggest the following:
- test simple examples
- test properties on the resulting matrix (try with identity, skew matrix, etc)
In general, testing requires some thinking about what the method should satisfy and the type of user errors that it should catch

ihmc-robotics-toolkit/src/test/java/us/ihmc/robotics/MatrixMissingToolsTest.java
```diff
@@ -313,4 +313,61 @@ public void testSetSelectedMatrixDiagonals()
...
+         DMatrixRMaj matrixMore = new DMatrixRMaj(rowSize, columnSize);
+         for (int j = 0; j < rowSize; j++)
+         {
+            for (int k = 0; k < columnSize; k++)
+            {
+               matrixMore.set(j, k, matrixLess.get(j, k) + random.nextDouble(0.01, 0.05));
+            }
+         }
+
+         assertTrue(MatrixMissingTools.elementWiseLessThan(matrixLess, matrixMore));
```
SylvainBertrand, line 422: Here I could swap the current implementation of your method with the following and your test would still pass:
```
public static boolean elementWiseLessThan(DMatrixRMaj a, DMatrixRMaj b)
{
   return true;
}
```
You're on the right track for the test, it is just incomplete.

SylvainBertrand (changes requested):

stefanfasano (approved):

ihmc-robotics-toolkit/src/test/java/us/ihmc/robotics/MatrixMissingToolsTest.java
```diff
@@ -313,4 +313,123 @@ public void testSetSelectedMatrixDiagonals()
...
+         assertFalse(MatrixMissingTools.elementWiseLessThan(matrixMore, matrixLess));
+
+         for (int j = 0; j < matrixLess.getNumElements(); j++)
+         {
+            DMatrixRMaj matrixLessNotAll = new DMatrixRMaj(matrixLess);
+            matrixLessNotAll.set(j, matrixMore.get(j) + 0.001);
+            assertFalse(MatrixMissingTools.elementWiseLessThan(matrixLess, matrixMore));
```
rjgriffin42: it looks like you didn't use the new `matrixLessNotAll` object

Beomyeong: OH, you are right.

Beomyeong: Fixed

SylvainBertrand (approved):

PR #225, opened 2024-06-17
Fixed YOLOv8 Checkbox

TomaszTB:
The recent visualizer changes caused YOLO heartbeat to only be activated when the visualizer was expanded (due to the logic being in the render loop.. my bad). This PR fixes that!


ds58:
Nice, ty for the fix

ds58 (approved):

PR #243, opened 2024-06-19
Cleanup RTPSCommunicationFactory.

calvertdw:


ihmc-communication/src/main/java/us/ihmc/communication/RTPSCommunicationFactory.java
```diff
@@ -123,7 +115,6 @@ public RTPSCommunicationFactory()
       defaultAddressRestriction = foundAddressRestriction;
       if (defaultAddressRestriction != null)
          LogTools.info("Setting IP restriction: {}", defaultAddressRestriction.getHostAddress());
-      createParticipant(rtpsDomainID);
```
calvertdw, line 126: A participant was getting created here I noticed when collecting stats. I don't think it's used anywhere. It seems this class was part before we had unit tests for ROS 2 and Pub Sub stuff. Now that we've got those tests. I think this stuff can get removed.

ihmc-communication/src/main/java/us/ihmc/communication/RTPSCommunicationFactory.java
```diff
@@ -14,26 +13,19 @@
...
- * Creates and Manages participants
+ * Loads the Domain ID and Subnet Restriction from ~/.ihmc/IHMCNetworkParameters.ini
  */
 public class RTPSCommunicationFactory
 {
    private static final int START_OF_RANDOM_DOMAIN_RANGE = 200;
    private static final List<InterfaceAddress> MACHINE_INTERFACE_ADDRESSES = findInterfaceAddresses();
 
-   private final Domain domain = DomainFactory.getDomain(DomainFactory.PubSubImplementation.FAST_RTPS);
-   private final TIntObjectHashMap<Participant> participants = new TIntObjectHashMap<>();
```
calvertdw, line 36: I think these were not used.

ds58 (approved):

PR #252, opened 2024-06-21
Feature/point cloud viz optimization

ds58:
I put the kernel and mesh building on a separate thread and throttled it to 15hz. This has significant effect when using a GTX 1080ti. When rendering a ZED 2 point cloud, FPS goes from ~35->250. I do notice less of a performance jump on an RTX 3080ti however, probably because of the larger memory bandwidth available.


TomaszTB:
Does this mean all point clouds will be rendered at 15hz?

ds58:
> Does this mean all point clouds will be rendered at 15hz?

Yes

ds58:
I think I'm gonna just wait until we can do gl shared vertex buffers

PR #263, opened 2024-06-25
Remove unecessary parameterization and make InstantDetection not abst…

calvertdw:
…ract.


TomaszTB:
I had made `InstantDetection` abstract since the `DetectionManager` code needs to differentiate between, say, CenterPose and YOLO detections. I'm not entirely sure how it would handle generic `InstantDetection`s from different sources.

TomaszTB:
Also, could you run the tests and make sure they still work?

PR #264, opened 2024-06-25
feature/ground-plane-yaw-support

stefanfasano:
Can now add a yaw component to the ground plane orientation estimate


SylvainBertrand (approved):
I'm not super familiar with these classes but that looks good to me

PR #266, opened 2024-06-26
ZED SVO Recording UI

ds58:


calvertdw (approved):

ihmc-perception/src/main/java/us/ihmc/sensors/ZEDColorDepthImageRetriever.java
```diff
@@ -408,6 +410,16 @@ public int getCameraID()
       return cameraID;
    }
 
+   public void setPaused(boolean paused)
```
TomaszTB: I'm not sure if it would work, but couldn't this functionality be achieved by stopping the thread? And RestartableThread has `isRunning()` method which could be used in `isPaused()`. Could you look into whether this would work?

ds58: I changed it to use the start() and stop() of the parent class

PR #273, opened 2024-06-28
added some better prints to some of the wbcc commands

rjgriffin42:


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controllerCore/command/feedbackController/SpatialFeedbackControlCommand.java
```diff
@@ -1326,6 +1326,7 @@ else if (object instanceof SpatialFeedbackControlCommand)
    public String toString()
    {
       String ret = getClass().getSimpleName() + ": ";
+      ret += "link = " + spatialAccelerationCommand.getEndEffector().getName() + ", ";
       ret += "base = " + spatialAccelerationCommand.getBase() + ", ";
       ret += "endEffector = " + spatialAccelerationCommand.getEndEffector() + ", ";
```
SylvainBertrand: I think we should just change the base and endEffector lines to only return the name of the bodies:
```
      ret += "base = " + spatialAccelerationCommand.getBase().getName() + ", ";
      ret += "endEffector = " + spatialAccelerationCommand.getEndEffector().getName() + ", ";
```

SylvainBertrand (approved):

PR #285, opened 2024-07-07
Clarify up Timer code, improve sleepUntilExpiration, add sleepAtLeast method with sleep guarantee, and update Throttler to use it too.

calvertdw:
image
image
image

### DIFF
--- a/ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/MissingThreadTools.java
+++ b/ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/MissingThreadTools.java
@@ -13,12 +13,17 @@ import java.util.concurrent.locks.LockSupport;
 public class MissingThreadTools
 {
    /**
-    * {@link ThreadTools#sleepSeconds} can actually return early.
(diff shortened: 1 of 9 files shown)


ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/MissingThreadTools.java
```diff
@@ -10,6 +11,24 @@
 
 public class MissingThreadTools
 {
+   /**
+    * {@link ThreadTools#sleepSeconds} can actually return early.
```
TomaszTB: How did you realize this? Looking at the code for `ThreadTools#sleep()` it seems to do essentially the same thing as this method (sleep, see if it slept long enough, if not sleep again). I'd be surprised if it somehow  returned early.

calvertdw: Yup it was really the other sleepSeconds method cutting off the subnanosecond part, allowing it to undersleep by a nanosecond at most.

TomaszTB (approved):

ds58 (approved):

PR #286, opened 2024-07-08
Minor Change to RawImage

TomaszTB:
Little change I've been wanting to make when working with RawImages. 
Removed image width, height, and opencv type from constructor parameters as those are obtainable from the `Mat`/`GpuMat` itself.


ds58 (approved):

ihmc-perception/src/main/java/us/ihmc/perception/RawImage.java
```diff
@@ -105,6 +106,30 @@ public RawImage(RawImage other)
       this.orientation = other.orientation;
    }
 
+   public static RawImage fromMessage(ImageMessage imageMessage)
```
TomaszTB, line 109: I think this'll be useful for the YOLO detections, specifically getting the object mask image.

calvertdw (approved):

PR #289, opened 2024-07-10
fixed shift fraction - bug

LaughingCommit:
Checked the math in the CombineTwoShifts function, made tests to see if all values will pass the function.

### DIFF
--- a/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionTools.java
+++ b/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionTools.java
@@ -1,5 +1,7 @@
 package us.ihmc.commonWalkingControlModules.capturePoint.splitFractionCalculation;
 
+import static us.ihmc.commons.MathTools.clamp;
+
 public class SplitFractionTools
 {
    public static double appendSplitFraction(double desiredSplitFraction, double currentSplitFraction, double nominalSplitFraction)
@@ -17,6 +19,8 @@ public class SplitFractionTools
       if (currentShift == -1.0)
          return desiredShift;
 
+      //transfer = desired, default = nominal
+
       // TODO apply some clamps on the shifts
 
       if (desiredShift > nominalShift)
@@ -25,7 +29,7 @@ public class SplitFractionTools
          double desiredShiftForward = desiredPercentShiftForward * (1.0 - currentShift);
 
          // TODO clamp the output
-         return currentShift + desiredShiftForward;
+         return clamp ((currentShift + desiredShiftForward),0.01,0.99);
       }
       else
       {
@@ -33,7 +37,7 @@ public class SplitFractionTools
          double desiredShiftBackward = desiredPercentShiftBackward * currentShift;
 
          // TODO clamp the output
-         return currentShift - desiredShiftBackward;
+         return clamp((currentShift - desiredShiftBackward),0.01,0.99);
       }
    }
 }
--- a/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionTools.java
+++ b/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionTools.java
@@ -17,16 +17,22 @@ public class SplitFractionTools
       if (currentShift == -1.0)
          return desiredShift;
 
+      // TODO apply some clamps on the shifts
+
       if (desiredShift > nominalShift)
       {
          double desiredPercentShiftForward = (desiredShift - nominalShift) / (1.0 - nominalShift);
          double desiredShiftForward = desiredPercentShiftForward * (1.0 - currentShift);
+
+         // TODO clamp the output
          return currentShift + desiredShiftForward;
       }
       else
       {
          double desiredPercentShiftBackward = (nominalShift - desiredShift) / nominalShift;
          double desiredShiftBackward = desiredPercentShiftBackward * currentShift;
+
+         // TODO clamp the output
          return currentShift - desiredShiftBackward;
       }
    }
--- a/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionTools.java
+++ b/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionTools.java
@@ -4,6 +4,9 @@ import static us.ihmc.commons.MathTools.clamp;
 
 public class SplitFractionTools
 {
+
+   private final static double minimumFraction = 1e-5;
+
    public static double appendSplitFraction(double desiredSplitFraction, double currentSplitFraction, double nominalSplitFraction)
    {
       return combineTwoShifts(desiredSplitFraction, currentSplitFraction, nominalSplitFraction);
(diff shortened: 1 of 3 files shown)


ihmc-common-walking-control-modules/src/test/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionToolsTest.java
```diff
@@ -0,0 +1,162 @@
...
+      if (transferSplitFraction > defaultSplitFraction)
+      {
+         assertEquals(expectedForwardShift, actualSplitFraction, epsilon);
+      }
+      else
+      {
+         assertEquals(expectedBackwardShift, actualSplitFraction, epsilon);
+      }
+   }
+   //   @Test
```
LaughingCommit, line 117: commented this out as above test checks the function with all possible input values

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionTools.java
```diff
@@ -14,20 +16,31 @@ public static double appendWeightDistribution(double desiredWeightDistribution,
...
+      //transfer = desired, default = nominal
+
       if (desiredShift > nominalShift)
       {
          double desiredPercentShiftForward = (desiredShift - nominalShift) / (1.0 - nominalShift);
          double desiredShiftForward = desiredPercentShiftForward * (1.0 - currentShift);
-         return currentShift + desiredShiftForward;
+
+         //clamping the output to stay within bounds
+         return clamp((currentShift + desiredShiftForward), 0.00001, 0.99999);
```
rjgriffin42: I'd recommend making 0.00001 a static variable 
```private final static double minimumFraction = 1e-5;```
and then changing these to 
```clamp(a, minimumFraction, 1.0 - minimumFraction);```
for consistency and ease of change.

File: ihmc-common-walking-control-modules/src/test/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionToolsTest.java [hunk 3]
```diff
@@ -0,0 +1,162 @@
...
+
+   @Test
+   public void testRandom()
+   {
+      Random random = new Random(66);
+      defaultSplitFraction = random.nextDouble();
+      currentSplitFraction = random.nextDouble();
+      transferSplitFraction = random.nextDouble();
+      System.out.printf("Values: %.2f, %.2f, %.2f\n", defaultSplitFraction, currentSplitFraction, transferSplitFraction);
+      calculateValues();
```
rjgriffin42: Please change to `LogTools.info` so you get the trace of where it's printed from. You'd also probably  want to disable when it's running on Bamboo.

ihmc-common-walking-control-modules/src/test/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionToolsTest.java
```diff
@@ -0,0 +1,162 @@
...
+      if (transferSplitFraction > defaultSplitFraction)
+      {
+         assertEquals(expectedForwardShift, actualSplitFraction, epsilon);
+      }
+      else
+      {
+         assertEquals(expectedBackwardShift, actualSplitFraction, epsilon);
+      }
```
rjgriffin42, line 103: This is a good start. However, random will only return values that are between 0 and 1. do you want to test values that are out of bounds? 

Also we should wrap this in a for loop. Testing one random double instance is good, testing 1e5 instances is better.

ihmc-common-walking-control-modules/src/test/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionToolsTest.java
```diff
@@ -0,0 +1,162 @@
+package us.ihmc.commonWalkingControlModules.capturePoint.splitFractionCalculation;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.Random;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static us.ihmc.commons.MathTools.clamp;
+
+public class SplitFractionToolsTest
```
rjgriffin42, line 11: Can you add a few tests for right at the epsilon bounds? To make sure that the epsilon is enforced. So see what happens if the return is just above the minimum epsilon, right at the minimum epsilon, and just below the minimum epsilon. THat's quite literally the definition of your edge case here.

ihmc-common-walking-control-modules/src/test/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionToolsTest.java
```diff
@@ -0,0 +1,162 @@
...
+      else
+      {
+         percentShift = Math.abs(transferSplitFraction - defaultSplitFraction) / defaultSplitFraction;
+      }
+
+      actualSplitFraction = SplitFractionTools.appendSplitFraction(transferSplitFraction, currentSplitFraction, defaultSplitFraction);
+      double expectedForwardShiftedSplitFraction = percentShift * (1 - currentSplitFraction);
+      double expectedBackwardShiftedSplitFraction = percentShift * currentSplitFraction;
+      expectedForwardShift = currentSplitFraction + expectedForwardShiftedSplitFraction;
+      expectedBackwardShift = currentSplitFraction - expectedBackwardShiftedSplitFraction;
```
rjgriffin42: Is there a way to rearrange or re-do the math here? Right, it's the exact same math in the method. As the easiest example, if in the method your calculating 
y = a * (x + b),

in the test, you could check
y = a * x + a * b
just so it's slightly different.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionTools.java
```diff
@@ -14,20 +16,31 @@ public static double appendWeightDistribution(double desiredWeightDistribution,
 
    private static double combineTwoShifts(double desiredShift, double currentShift, double nominalShift)
    {
+      //clamping input values, incase we get negative values
+      desiredShift = clamp(desiredShift, 0.0, 1);
+      currentShift = clamp(currentShift, 0.0, 1);
+      nominalShift = clamp(nominalShift, 0.0, 1);
+
       if (currentShift == -1.0)
          return desiredShift;
```
rjgriffin42, line 24: can you move this before the clamp? It's meant to be an indicator, I think. Where in some methods you would pass in null as a way of getting a certain result, you can do that with primitives, so you use like -1.0. Right now, if you're clamping the value before getting here, this if statement will never return true.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/splitFractionCalculation/SplitFractionTools.java
```diff
@@ -14,20 +16,31 @@ public static double appendWeightDistribution(double desiredWeightDistribution,
 
    private static double combineTwoShifts(double desiredShift, double currentShift, double nominalShift)
    {
+      //clamping input values, incase we get negative values
+      desiredShift = clamp(desiredShift, 0.0, 1);
+      currentShift = clamp(currentShift, 0.0, 1);
+      nominalShift = clamp(nominalShift, 0.0, 1);
```
rjgriffin42: I know it's silly, but please use 1.0 to make it explicit you want a double an not an integer.

rjgriffin42:
This is good work, but I think there's just a few more things that need to be done. SHould be pretty quick.

PR #291, opened 2024-07-11
Feature/more demo poses

rjgriffin42:


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controlModules/foot/MoveViaWaypointsState.java
```diff
@@ -214,6 +214,8 @@ public FeedbackControlCommand<?> createFeedbackControlTemplate()
    @Override
    public void onExit(double timeInState)
    {
+      controlManager.getTaskspaceControlState().onExit(timeInState);
+      controlManager.getJointspaceControlState().onExit(timeInState);
```
rjgriffin42, line 218: @SylvainBertrand if we don't call this, the desired state doesn't necessarily get reset. Then, if you put the foot down, awlk forward, and lift it again, it starts with the first waypoint as the old terminal position. I set up a better test in EndToEndFootTrajectoryMessageTest that showed this.

SylvainBertrand: wow I wonder if it was always like that and just never triggered the issue. I wonder if we should add anything in the `onEntry`

SylvainBertrand (approved):

PR #292, opened 2024-07-11
Minimize Number of Fields in Behavior Tree Messages

calvertdw:


calvertdw:
This is working on real robot for door and exploration behaviors and I got verbal approval from both Dex and Luigi.

PR #295, opened 2024-07-12
Add stop all trajectories boolean in order to have this safety on the sliderboard.

calvertdw:


rjgriffin42:
Can we make a test or two for this? We honestly should have had one for the StopAllTrajectoryCommand, and it's embarassing that we don't.

PR #298, opened 2024-07-13
Feature/persistent detection tests

rjgriffin42:
This adds some unit tests, and makes things a little more reference safe, so that we do more pose copying rather than modifying.


TomaszTB (approved):
Looks great to me!

calvertdw (approved):

ds58 (approved):

PR #299, opened 2024-07-15
small logic correction in VelocityBasedSteppingPlugin

stefanfasano:
This feature ensures that the scaling of the physical joystick input matches that of our max/min desired velocity parameters


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/plugin/VelocityBasedSteppingPlugin.java
```diff
@@ -98,7 +98,7 @@ else if (startWalkingMessenger != null && walk.getValue() != walkPreviousValue.g
       double desiredVelocityY = desiredVelocity.getY();
       double turningVelocity = desiredTurningVelocityProvider.getTurningVelocity();
 
-      if (desiredVelocityProvider.isUnitVelocity())
```
stefanfasano, line 101: this was incorrect because the condition below it assumed the desired velocity was a [0, 1] scaling factor that you then had to multiply by the max/min desired velocities

PR #300, opened 2024-07-15
YOLO Detection Optimization

TomaszTB:
Small optimization so the annotated image is computed faster

### DIFF
--- a/ihmc-perception/src/main/java/us/ihmc/perception/detections/yolo/YOLOv8DetectionExecutor.java
+++ b/ihmc-perception/src/main/java/us/ihmc/perception/detections/yolo/YOLOv8DetectionExecutor.java
@@ -35,9 +35,9 @@ import us.ihmc.ros2.ROS2PublisherBasics;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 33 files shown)


ds58:
Can this be merged into `feature/multiple-yolo-models`?

ihmc-perception/src/main/java/us/ihmc/perception/detections/yolo/YOLOv8DetectionExecutor.java
```diff
@@ -63,14 +63,14 @@ public class YOLOv8DetectionExecutor
 
    // TODO: temp hack
    private int lastRunDetectorIndex = 0;
-   private final Map<Integer, YOLOv8DetectionResults> yoloDetectionResults = new HashMap<>();
+   private final Map<Integer, YOLOv8DetectionResults> yoloDetectionResults = new ConcurrentHashMap<>();
```
TomaszTB, line 66: Not sure if this is necessary but technically this could be accessed concurrently so I thought it'd be safer this way.

ihmc-perception/src/main/java/us/ihmc/perception/detections/yolo/YOLOv8DetectionResults.java
```diff
@@ -17,17 +15,13 @@
...
-   private double floatTime = 0.0;
-   private final static Stopwatch booleanWatch = new Stopwatch();
-   private double booleanTime = 0.0;
-
    private final Set<YOLOv8DetectionOutput> detections;
    private final MatVector outputBlobs;
    private final RawImage detectionImage;
    private final FloatIndexer outputMasksIndexer;
-   private final Map<String, Mat> objectMasks = new HashMap<>();
+   private final Map<YOLOv8DetectionOutput, Mat> objectMasks = new HashMap<>();
```
TomaszTB, line 22: When a mask gets calculated, it's stored internally so we don't have to re-calculate if it's requested again.

PR #302, opened 2024-07-15
Make full footstep planner parameter set tunable per behavior footste…

calvertdw:
…p plan action.

TODO: 

- Sort out reducing the number of things saved with the JSON, possibly saving only certain ones
- Add widgets for key parameters


calvertdw:
Tested on real robot with Luigi.

PR #312, opened 2024-07-19
Show disabled properties when property set waiting for upated values from remote

TomaszTB:
I think it's nice to see the adjustable parameters, even if the values may be incorrect and they cannot be adjusted immediately. 
Screenshot from 2024-07-19 17-24-30
vs
Screenshot from 2024-07-19 17-28-01


calvertdw (approved):
Definitely agree 👍

ds58 (approved):

PR #313, opened 2024-07-22
Fix bugs that ensure creation of the wrong folder.

calvertdw:
Fixes the useless creation of these folders:


PotatoPeeler3000 (approved):
Nice! Saw that the other day and I was like lol what is this

PR #335, opened 2024-08-07
More CUDAImageEncoder (now CUDAJPEGProcessor) Improvements

TomaszTB:
This PR:
* simplifies method calls to encode images
* adds method to encode gray scale image
* adds decoding ability

### DIFF
--- a/ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/pointCloud/RDXROS2ColoredPointCloudVisualizerColorChannel.java
+++ b/ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/pointCloud/RDXROS2ColoredPointCloudVisualizerColorChannel.java
@@ -2,13 +2,13 @@ package us.ihmc.rdx.ui.graphics.ros2.pointCloud;
 
 import org.bytedeco.opencl.global.OpenCL;
 import org.bytedeco.opencv.global.opencv_core;
-import org.bytedeco.opencv.global.opencv_imgcodecs;
 import org.bytedeco.opencv.global.opencv_imgproc;
 import org.bytedeco.opencv.opencv_core.Mat;
 import perception_msgs.msg.dds.ImageMessage;
 import us.ihmc.log.LogTools;
 import us.ihmc.perception.BytedecoImage;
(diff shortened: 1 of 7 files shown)


ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/pointCloud/RDXROS2ColoredPointCloudVisualizerColorChannel.java
TomaszTB, line 1: This seems like the perfect class to use the CUDA jpeg decoding. Let me know if there's any other places we should use it.

ds58: Is CUDA a requirement of RDX already? Should we expect anyone running the UI to need a CUDA device?

ds58 (approved):

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/pointCloud/RDXROS2ColoredPointCloudVisualizerColorChannel.java
TomaszTB: Oh that's a great point. I don't recall RDX requiring CUDA as of yet. Perhaps I should remove the RDX changes from this PR?

TomaszTB: Ultimately I'd like to have something like a generic `ImageProcessor` which internally checks whether CUDA is available or not, and defaults to OpenCV if no CUDA devices are found.

ds58: Perhaps we should do that so the UI stuff doesn't have to require CUDA

PR #341, opened 2024-08-13
added the ability to use all controller API methods without the topic…

rjgriffin42:
…s being delcared in the m essage calss types list


ihmc-communication/src/main/java/us/ihmc/communication/controllerAPI/ControllerAPI.java
```diff
@@ -142,15 +142,11 @@ public static ROS2Topic<?> getBaseTopic(String controlModuleName, String robotNa
...
-      }
-      if (outputMessageClasses.contains(messageClass))
-      {
+      else if (outputMessageClasses.contains(messageClass))
          return baseTopic.withOutput().withTypeName(messageClass).withQoS(getQoS(messageClass));
-      }
-
-      throw new RuntimeException("Topic does not exist: " + messageClass);
+      else
+         return baseTopic.withTypeName(messageClass).withQoS(getQoS(messageClass));
```
rjgriffin42, line 149: without this else, it is not usable for messages that aren't included in the input/output message class lists.

calvertdw (approved):

PR #344, opened 2024-08-14
Feature/realsensestuff

PotatoPeeler3000:


ihmc-perception/src/main/java/us/ihmc/sensors/RealSensePublisher.java
```diff
@@ -0,0 +1,77 @@
...
+   private static final String REALSENSE_SERIAL_NUMBER = System.getProperty("d455.serial.number", "213522252883");
+   private static final ROS2Topic<ImageMessage> REALSENSE_COLOR_TOPIC = PerceptionAPI.D455_COLOR_IMAGE;
+   private static final ROS2Topic<ImageMessage> REALSENSE_DEPTH_TOPIC = PerceptionAPI.D455_DEPTH_IMAGE;
+
+   private final RealsenseColorDepthImageRetriever realsenseImageRetriever;
+   private final RealsenseColorDepthImagePublisher realsenseImagePublisher;
+
+   private final ROS2DemandGraphNode realsenseDemandNode;
+
+   protected final ScheduledExecutorService executorService = ExecutorServiceTools.newScheduledThreadPool(1,
```
TomaszTB: This guys unused

ihmc-perception/src/main/java/us/ihmc/sensors/RealSensePublisher.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.function.Supplier;
+
+public class RealSensePublisher
+{
+   private static final String REALSENSE_SERIAL_NUMBER = System.getProperty("d455.serial.number", "213522252883");
+   private static final ROS2Topic<ImageMessage> REALSENSE_COLOR_TOPIC = PerceptionAPI.D455_COLOR_IMAGE;
+   private static final ROS2Topic<ImageMessage> REALSENSE_DEPTH_TOPIC = PerceptionAPI.D455_DEPTH_IMAGE;
+
+   private final RealsenseColorDepthImageRetriever realsenseImageRetriever;
+   private final RealsenseColorDepthImagePublisher realsenseImagePublisher;
```
TomaszTB: The retriever and publisher both have `destroy` methods which should be called when the process ends. Perhaps a shutdown hook is the best way to call it? I can't think of a better method.

ihmc-perception/src/main/java/us/ihmc/sensors/RealSensePublisher.java
```diff
@@ -0,0 +1,77 @@
...
+      realsenseImageRetriever = new RealsenseColorDepthImageRetriever(new RealsenseDeviceManager(),
+                                                                      REALSENSE_SERIAL_NUMBER,
+                                                                      RealsenseConfiguration.D455_COLOR_720P_DEPTH_720P_30HZ,
+                                                                      realsenseFrameSupplier,
+                                                                      realsenseDemandNode);
+
+      realsenseImagePublisher = new RealsenseColorDepthImagePublisher(REALSENSE_DEPTH_TOPIC, REALSENSE_COLOR_TOPIC);
+
+      RestartableThread realsenseProcessAndPublishThread = new RestartableThread("RealsenseProcessAndPublish", this::processAndPublishRealsense);
+      realsenseProcessAndPublishThread.start();
```
TomaszTB: I don't think you need to start a separate thread here. Could just do a `while (true)` thing if you want.

TomaszTB (changes requested):

ihmc-perception/src/main/java/us/ihmc/sensors/RealSensePublisher.java
```diff
@@ -0,0 +1,77 @@
...
+         realsenseDepthImage.release();
+         realsenseColorImage.release();
+      }
+      else
+         ThreadTools.sleep(500);
+   }
+
+   public static void main(String[] args)
+   {
+      ROS2Node ros2Node = ROS2Tools.createROS2Node(CommunicationMode.INTERPROCESS.getPubSubImplementation(), "perception_autonomy_process");
```
TomaszTB: Oh, shouldn't the `ROS2Node` be destroyed too?

ihmc-perception/src/main/java/us/ihmc/sensors/RealSensePublisher.java
calvertdw, line 1: Could you think of a more specific name and lower case the S in Realsense to match the others. i.e. This class has a `RealsenseColorDepthImageRetriever` and `RealsenseColorDepthImagePublisher`. I would expect this class to be called StandaloneRealsenseRetrieverAndPublisher or something.

ds58: I was thinking the same thing about the name. However Intel actually stylizes it as RealSense. So, in this case it would just be Realsense to pattern match the rest of the classes.

calvertdw: Yeah homogeneity is more important than "best" in this case in my opinion. I think converting everywhere else to capital S seems a bit silly as well.

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/StandAloneRealsenseProcess.java
```diff
@@ -0,0 +1,89 @@
...
+   public void destroy()
+   {
+      realsenseImagePublisher.destroy();
+      realsenseImageRetriever.destroy();
+      realsenseDemandNode.destroy();
+   }
+
+   public static void main(String[] args)
+   {
+      ROS2Node ros2Node = ROS2Tools.createROS2Node(DomainFactory.PubSubImplementation.FAST_RTPS, "nadia_realsense_process");
```
TomaszTB: This ROS2Node should be destroyed too, right?

PotatoPeeler3000: Nice yep got it

PR #346, opened 2024-08-15
Fixing the issue for orientation / point FeedbackControl

Beomyeong:
Mostly, the issues come from the non-matching frame.


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/feedbackController/taskspace/PointFeedbackController.java
```diff
@@ -610,34 +610,39 @@ private void computeIntegralTerm(FrameVector3D feedbackTermToPack)
...
-      feedbackTermToPack.clipToMaxNorm(maximumIntegralError);
-      yoErrorPositionIntegrated.setIncludingFrame(feedbackTermToPack);
-      yoErrorPositionIntegrated.changeFrame(trajectoryFrame);
-      yoErrorPositionIntegrated.setCommandId(currentCommandId);
+         // If the trajectory frame changed reset the integration.
+         if (yoErrorPositionIntegrated.getReferenceFrame() != trajectoryFrame)
+         {
+            yoErrorPositionIntegrated.setToZero(trajectoryFrame);
+         }
+         if (yoErrorPosition.getReferenceFrame() != trajectoryFrame)
```
PotatoPeeler3000, line 622: Do you think its worth adding a comment saying that "if the trajectory frame changed, reset the position, or maybe adding to the other comment"

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/feedbackController/taskspace/SpatialFeedbackController.java
```diff
@@ -10,20 +10,12 @@
 import us.ihmc.commonWalkingControlModules.controllerCore.command.inverseKinematics.SpatialVelocityCommand;
 import us.ihmc.commonWalkingControlModules.controllerCore.command.virtualModelControl.VirtualModelControlCommand;
 import us.ihmc.commonWalkingControlModules.controllerCore.command.virtualModelControl.VirtualWrenchCommand;
-import us.ihmc.commonWalkingControlModules.controllerCore.data.FBPose3D;
-import us.ihmc.commonWalkingControlModules.controllerCore.data.FBQuaternion3D;
-import us.ihmc.commonWalkingControlModules.controllerCore.data.FBRateLimitedVector6D;
-import us.ihmc.commonWalkingControlModules.controllerCore.data.FBVector3D;
-import us.ihmc.commonWalkingControlModules.controllerCore.data.FBVector6D;
+import us.ihmc.commonWalkingControlModules.controllerCore.data.*;
```
PotatoPeeler3000, line 13: You might not have the most up to date version of the formatter, I would hae thought it would not use the wildcard (*) modifier for such a little amount of imports

PotatoPeeler3000 (approved):
Do we wannt test this on the robot? I'm not an expert in this material but it seems like if we are changing controller code it would be good to test on robot before merging

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/feedbackController/taskspace/PointFeedbackController.java
Beomyeong: Oh. it sounds good. I'll add that comments into the code

Beomyeong:
> Do we wannt test this on the robot? I'm not an expert in this material but it seems like if we are changing controller code it would be good to test on robot before merging

Before test it on the robot, I'll check which robot tests are needed. 
Just walking might not be enough to test this.

rjgriffin42:
@Beomyeong yea walking isn't enough. Should also test moving the pelvis around and the chest around, as well as flamingo mode. Those shoudl all help

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/feedbackController/taskspace/OrientationFeedbackController.java
```diff
@@ -631,6 +631,12 @@ public void computeIntegralTerm(FrameVector3D feedbackTermToPack)
          yoErrorRotationVectorIntegrated.setToZero(trajectoryFrame);
       }
 
+      if (yoErrorOrientation.getReferenceFrame() != trajectoryFrame)
+      {
+         yoErrorOrientation.setToZero(trajectoryFrame);
+         yoErrorRotationVectorIntegrated.setToZero(trajectoryFrame);
+      }
```
rjgriffin42, line 639: I'm confused by this. How would it ever reach this point? If the error was in a different frame, how did the error get calculated in that frame in the PD controller portion?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/feedbackController/taskspace/PointFeedbackController.java
rjgriffin42: I have the same comment here as before.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/feedbackController/taskspace/OrientationFeedbackController.java
Beomyeong: I'm not sure though, this if statement is only triggered through the orientationTest: comapreAgainstSpatialController. 
Fist of all, the structure of the orientationfeedbackController is differ with orientation part of the spatialController. 
With the robot simulation, I've never seen this if statement is triggered. 
For now, I just pushed. But later, I'll look into it again when I have spare time.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/feedbackController/taskspace/SpatialFeedbackController.java
Beomyeong: I tried again , but that wildcard has been used again. 
Has the ihmcformatter been changed?
If so, I'll update this later.

PR #347, opened 2024-08-20
ZED ArUco Demo

TomaszTB:
Simple RDX demo for ArUco marker detection using a ZED camera. 

https://github.com/user-attachments/assets/02225142-954a-438e-b7d9-4b1fd5be9608


calvertdw (approved):
Nice!

PR #353, opened 2024-08-23
properly deprecated everything

rjgriffin42:


ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/BehaviorModule.java
```diff
@@ -18,6 +18,7 @@
  *
  * @deprecated Replaced by ROS2BehaviorTreeExecutor
  */
+@Deprecated
```
calvertdw, line 21: This should be covered by the Javadoc, no?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/ArrayTools.java
```diff
@@ -32,7 +32,7 @@ public static double[] parseDoubleArray(String stringSource) throws IOException
       StringTokenizer tokenizer = new StringTokenizer(line.replace("{", "").replace("}", "").replace(" ", "").replace(";", ""), ",");
       while (tokenizer.hasMoreElements())
       {
-         retArray.add(new Double(Double.parseDouble((String) tokenizer.nextElement())));
+         retArray.add(Double.valueOf(Double.parseDouble((String) tokenizer.nextElement())));
```
calvertdw, line 35: I would imagine this would be sufficient?
```suggestion
         retArray.add(Double.parseDouble((String) tokenizer.nextElement()));
```

ihmc-model-file-loader/src/main/java/us/ihmc/modelFileLoaders/SdfLoader/RobotDescriptionFromSDFLoader.java
```diff
@@ -636,7 +637,7 @@ private void addLidarMounts(SDFSensor sensor, JointDescription scsJoint, SDFLink
             }
             else
             {
-               System.err.println("Unknown noise model: " + sdfNoise.getType());
+               LogTools.info("Unknown noise model: " + sdfNoise.getType());
```
calvertdw, line 640: Not the case here, but it reminded me, careful doing this more broadly. Anything that executes as part of a Runtime shutdownHook can't use LogTools anymore, because log4j2 is one of the things that gets shutdown in that phase.

calvertdw (approved):

ihmc-model-file-loader/src/main/java/us/ihmc/modelFileLoaders/SdfLoader/RobotDescriptionFromSDFLoader.java
rjgriffin42: No clue, if that's the case I can cahnge it back.

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/BehaviorModule.java
rjgriffin42: If you look at the warnings shown by the CI runner, it generates a warning for everywhere where deprecated is declared in the javadoc but not provided as an annotation.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/ArrayTools.java
rjgriffin42: I'll probably shelve this PR and do anotehr pass on cleaning up deprecations. They make the build log a pain.

PR #357, opened 2024-08-29
added some useful extra methods

rjgriffin42:


ihmc-humanoid-robotics/src/main/java/us/ihmc/humanoidRobotics/communication/controllerAPI/command/EuclideanTrajectoryControllerCommand.java
```diff
@@ -296,6 +297,7 @@ public ReferenceFrame getTrajectoryFrame()
 
    public void setTrajectoryFrame(ReferenceFrame trajectoryFrame)
    {
+      trajectoryPointList.setReferenceFrame(trajectoryFrame);
```
rjgriffin42: I think this one could be debated.

Beomyeong: Isn't it good? if we use the same reference frame for all the trajectory point.

Beomyeong (approved):

ihmc-humanoid-robotics/src/main/java/us/ihmc/humanoidRobotics/communication/controllerAPI/command/EuclideanTrajectoryControllerCommand.java
rjgriffin42: I removed this. The trajectory frmae is held in the command, but we call the frame in the point list the data frame elsewhere in this class, so I want to keep it consistent. We can definitely revisit, but this is an unnecessary change, so I don't want to make it without being really thorough on the class.

PR #359, opened 2024-09-03
Simplify RealSense Serial Numbers

TomaszTB:
This https://ihmcrobotics.slack.com/archives/CBPNP7ZNU/p1724680812321409


ihmc-perception/src/main/java/us/ihmc/perception/realsense/RealsenseDeviceManager.java
```diff
@@ -22,6 +21,8 @@
  */
 public class RealsenseDeviceManager
 {
+   public static final String[] REALSENSE_SERIAL_NUMBERS = {"213522252883", "215122253249"};
```
TomaszTB: Perhaps there's a better place for this list? Also, let me know if you know any other realsense serial numbers. These two are the numbers for the robot, and the one at my desk.

calvertdw: One on Nehar's desk: 215122254074

calvertdw: However, can't we just remove the serial numbers entirely and just connect to the first one connects?

TomaszTB: Oh yeah, we absolutely can

ihmc-perception/src/main/java/us/ihmc/perception/realsense/RealsenseDevice.java
```diff
@@ -83,6 +84,9 @@ public RealsenseDevice(rs2_context context, rs2_device device, String serialNumb
       realsense2.rs2_config_enable_stream(config, realsense2.RS2_STREAM_DEPTH, DEPTH_STREAM_INDEX, depthWidth, depthHeight, realsense2.RS2_FORMAT_Z16, fps, error);
       checkError("Failed to enable stream.");
 
+      BytePointer serialNumber = realsense2.rs2_get_device_info(device, RS2_CAMERA_INFO_SERIAL_NUMBER, error);
```
TomaszTB, line 101: Instead of passing in the serial number (which can be null), we get it from the device

ihmc-perception/src/main/java/us/ihmc/perception/realsense/RealsenseDeviceManager.java
```diff
@@ -116,7 +117,7 @@ public rs2_device createDevice(String serialNumberToFind)
             if (deviceRecommendedFirmwareVersion != null)
                LogTools.info("Realsense device recommended firmware version: {}", deviceRecommendedFirmwareVersion);
 
-            if (deviceSerialNumber.contains(serialNumberToFind.toLowerCase()))
+            if (serialNumberToFind == null || deviceSerialNumber.contains(serialNumberToFind.toLowerCase()))
```
TomaszTB: Serial number can be specified, or left null

ihmc-high-level-behaviors/src/main/java/us/ihmc/PerceptionAndAutonomyProcess.java
```diff
@@ -20,15 +20,11 @@
 import us.ihmc.communication.ros2.ROS2PublishSubscribeAPI;
 import us.ihmc.euclid.referenceFrame.ReferenceFrame;
 import us.ihmc.log.LogTools;
-import us.ihmc.perception.BallDetectionManager;
-import us.ihmc.perception.BytedecoImage;
-import us.ihmc.perception.IterativeClosestPointManager;
-import us.ihmc.perception.RapidHeightMapManager;
-import us.ihmc.perception.RawImage;
+import us.ihmc.perception.*;
```
PotatoPeeler3000: Do you have the most up to date formatter? I was under the impression that we increased the number of imports before the wildcard would get used?

ihmc-perception/src/main/java/us/ihmc/perception/realsense/RealsenseDevice.java
PotatoPeeler3000: Is it possible to not get the serial number? If its coming from the device itself what would that mean?

ihmc-perception/src/main/java/us/ihmc/perception/realsense/RealsenseDeviceManager.java
PotatoPeeler3000: Would there ever be a case where you would want to do this?

PotatoPeeler3000 (commented):
A few questions, would it not be better to remove the serial number all together? I guess that begs the question if its possible to run two realsenses at the same time.

ihmc-perception/src/main/java/us/ihmc/perception/realsense/RealsenseDeviceManager.java
TomaszTB: I think we'd want the serial numbers if we have multiple realsense devices so we can specify which one to connect to. This seems very likely with Alexander

ihmc-perception/src/main/java/us/ihmc/perception/realsense/RealsenseDevice.java
```diff
@@ -3,7 +3,21 @@
 import org.bytedeco.javacpp.BytePointer;
 import org.bytedeco.javacpp.FloatPointer;
 import org.bytedeco.librealsense2.global.realsense2;
-import org.bytedeco.librealsense2.*;
```
calvertdw, line 6: Probably better with the *

ds58:
What if you want to use multiple RealSense devices? How can you handle that?

ihmc-perception/src/main/java/us/ihmc/perception/realsense/RealsenseDevice.java
calvertdw: I think it's useful to keep it. Puts it in the log.

calvertdw (approved):

calvertdw:
> What if you want to use multiple RealSense devices? How can you handle that?

You don't because we don't ever need to. Why wouldn't we just build that if we need it

ihmc-perception/src/main/java/us/ihmc/perception/realsense/RealsenseDevice.java
PotatoPeeler3000: Isn't this just the formatter? Shouldn't we stick to those guidlines so its universal?

PotatoPeeler3000 (approved):

ihmc-perception/src/main/java/us/ihmc/perception/realsense/RealsenseDevice.java
calvertdw: I thought that was the guideline but apparently it's 20 lines before it collapses

TomaszTB:
> What if you want to use multiple RealSense devices? How can you handle that?

We'll have to reintroduce serial numbers haha

PR #364, opened 2024-09-11
Remove skully

ds58:


calvertdw (approved):
Goodbye Skully, see ya never

PR #372, opened 2024-09-24
Remove standalone realtime ros2node from pointcloud viz

ds58:
This seemed unnecessary to me to have an extra ros2node in the point cloud visualizer

### DIFF
--- a/ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/pointCloud/RDXROS2ColoredPointCloudVisualizer.java
+++ b/ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/pointCloud/RDXROS2ColoredPointCloudVisualizer.java
@@ -8,13 +8,11 @@ import imgui.type.ImBoolean;
 import imgui.type.ImFloat;
 import imgui.type.ImInt;
 import perception_msgs.msg.dds.ImageMessage;
-import us.ihmc.communication.ROS2Tools;
 import us.ihmc.euclid.transform.RigidBodyTransform;
 import us.ihmc.log.LogTools;
 import us.ihmc.perception.CameraModel;
 import us.ihmc.perception.opencl.OpenCLFloatBuffer;
 import us.ihmc.perception.opencl.OpenCLManager;
-import us.ihmc.pubsub.DomainFactory.PubSubImplementation;
 import us.ihmc.rdx.RDXPointCloudRenderer;
 import us.ihmc.rdx.imgui.ImGuiAveragedFrequencyText;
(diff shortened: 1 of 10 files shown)


ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/pointCloud/RDXROS2ColoredPointCloudVisualizer.java
```diff
@@ -70,32 +67,30 @@ public class RDXROS2ColoredPointCloudVisualizer extends RDXROS2MultiTopicVisuali
 
 
    public RDXROS2ColoredPointCloudVisualizer(String title,
-                                             PubSubImplementation pubSubImplementation,
+                                             ROS2Node ros2Node,
```
calvertdw, line 70: I would recommend accepting ROS2NodeInterface instead.

calvertdw (changes requested):
Otherwise looks good. This was added as a precaution because nobody knew what was going on and things weren't working. (years ago)

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/pointCloud/RDXROS2ColoredPointCloudVisualizer.java
ds58: I don't believe we do that anywhere else in the visualizers. Why should we do that here?

ds58: I personally don't like the way the inheritance works and would like to avoid using ROS2NodeInterface and ROS2NodeBasics at all

calvertdw: In general throughout the codebase it's the practice to use the least interface needed. I think unifying ROS2Node and RealtimeROS2Node would be good but that's another thing

calvertdw (approved):

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/pointCloud/RDXROS2ColoredPointCloudVisualizer.java
calvertdw: In this case, it's never gonna mater though

PR #375, opened 2024-09-26
Feature/delete unused controller things

rjgriffin42:
- Deletes the push recovery controller
- deletes the simple linear momentum rate
- makes quickster its own enum


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/factories/HighLevelHumanoidControllerFactory.java
```diff
@@ -576,7 +554,6 @@ public FrameVector3DReadOnly getCenterOfMassVelocity()
                                                                                   initialControllerState,
                                                                                   highLevelControllerParameters,
                                                                                   walkingControllerParameters,
-                                                                                  pushRecoveryControllerParameters,
```
stefanfasano, line 579: is it worth passing in the fast walking parameters this way (and with that adding fast walking parameters to WholeBodyControllerParameters and all that stuff), or does it make more sense to just keep it as a custom control state? Seems like the first option is a bit more invasive

james-p-foster (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/captureRegion/AchievableCaptureRegionCalculatorWithDelay.java
stefanfasano, line 1: Is this something that could be used by anything else? Or not really/do we have other calculators that do similar things

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/captureRegion/ReachableFootholdsCalculator.java
stefanfasano, line 1: Same here-Is this something that could be used by anything else? Or not really/do we have other calculators that do similar things

stefanfasano:
For the most part all looks good. Only question is would it be worth keeping some of the infrastructure surrounding the push recovery state, so that when we develop a new one we can easily swap out the old one for the new one? Or are we doing it in such a way that from the WBCC perspective, there will be no push recovery state, everything will be contained within the MPC state, and then at the whole body MPC level there will be a push recovery state or push recovery functionality? (assuming it is option 2 but just clarifying)

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/factories/HighLevelHumanoidControllerFactory.java
rjgriffin42: It's not possible because of the dependency tree. fast walking parameters is in internal, which depends on open source, where walking parameters is defined.

stephenmcc (approved):

PR #379, opened 2024-10-02
Feature/fix dynamic matrix calculator

rjgriffin42:


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controllerCore/WholeBodyInverseDynamicsSolver.java
```diff
@@ -200,40 +200,44 @@ public WholeBodyInverseDynamicsSolver(WholeBodyControlCoreToolbox toolbox, YoReg
       parentRegistry.addChild(registry);
    }
 
-   public void reset()
+   public void initialize()
```
rjgriffin42, line 203: The diff here is pretty bad between this method and reset. I switched the two methods in the ordering of the class,  because reset is always called right before compute, and initialize is only called on request. However, the diff makes it look like there were a lot more changes than in reality.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/optimization/DynamicsMatrixCalculator.java
```diff
@@ -2,99 +2,149 @@
...
+ * <p>
+ *    They can be further partition into the floating base component, demarcated by (.)<sub>base</sub>, and the actuated component, demarcated by
+ *    (.)<sub>body</sub>. These correspond to the top rows in the above equation for the base and the bottom rows for the actuated portion.
+ * </p>>
+ * <p>
+ *    In inverse dynamics classes, the joint accelerations (qDdot) and contact forces (&rho;) are typically solved for, and then the joint torques (&tau;) are
+ *    computed. This class is designed to provide the affine terms to compute these values.
+ * </p>
+ */
 public class DynamicsMatrixCalculator
```
rjgriffin42, line 28: This class was a travesty and makes me embarrassed that I wrote it. There's some weird stuff at the end, still, that's only used in the tests.

ihmc-common-walking-control-modules/src/test/java/us/ihmc/commonWalkingControlModules/momentumBasedController/optimization/DynamicsMatrixCalculatorTest.java
```diff
@@ -358,9 +362,13 @@ private void setupTest()
       wrenchMatrixCalculator = toolbox.getWrenchMatrixCalculator();
       jointIndexHandler = toolbox.getJointIndexHandler();
 
+      for (ContactablePlaneBody contactablePlaneBody : toolbox.getContactablePlaneBodies())
+         wrenchMatrixCalculator.submitPlaneContactStateCommand(nextPlaneContactStateCommand(random, contactablePlaneBody));
```
rjgriffin42, line 366: Without this, all the propagation of the wrench forces were zero. That means that the handling of contact forces between the inverse dynamics calculator and the dynamics matrix calculators weren't actually being tested. This then means that the bug on the sign of the contact forces wasn't caught. Adding this in caused all the tests to fail, and then pass again when the sign was fixed.

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/screwTheory/FloatingBaseRigidBodyDynamicsCalculator.java
```diff
@@ -4,133 +4,281 @@
...
    private static final double tolerance = 0.0001;
 
-   private static final LinearSolverDense<DMatrixRMaj> pseudoInverseSolver = LinearSolverFactory_DDRM.pseudoInverse(true);
+   private final LinearSolverDense<DMatrixRMaj> pseudoInverseSolver = LinearSolverFactory_DDRM.pseudoInverse(true);
 
    private final DMatrixRMaj matrixTranspose = new DMatrixRMaj(large, large);
-   private final DMatrixRMaj localVector = new DMatrixRMaj(large, 1);
+   private final DMatrixRMaj tempObjective = new DMatrixRMaj(large, large);
 
    public FloatingBaseRigidBodyDynamicsCalculator()
```
rjgriffin42, line 20: This class was also a travesty and could have been used as a "how not to write code" example.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/optimization/DynamicsMatrixCalculator.java
```diff
@@ -2,99 +2,149 @@
...
    private final GravityCoriolisExternalWrenchMatrixCalculator coriolisMatrixCalculator;
    private final ContactWrenchMatrixCalculator contactWrenchMatrixCalculator;
 
-   private final DynamicsMatrixCalculatorHelper helper;
-
-   private final DMatrixRMaj coriolisMatrix;
+   /** H(q) **/
+   private final DMatrixRMaj massMatrix;
+   /** C(q, qDdot) **/
+   private final DMatrixRMaj coriolisAndGravityMatrix;
```
stephenmcc: Rename field/getter to coriolisAndGravityVector? comment should be C(q, qDot) + g

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/optimization/DynamicsMatrixCalculator.java
```diff
@@ -2,99 +2,149 @@
...
+   /** J(q) **/
    private final DMatrixRMaj contactForceJacobian;
 
+   /** H<sub>base</sub>(q) **/
    private final DMatrixRMaj floatingBaseMassMatrix;
-   private final DMatrixRMaj floatingBaseCoriolisMatrix;
+   /** C<sub>base</sub>(q, qDdot) **/
+   private final DMatrixRMaj floatingBaseCoriolisAndGravityMatrix;
```
stephenmcc: Could also be renamed matrix -> vector

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/optimization/DynamicsMatrixCalculator.java
stephenmcc: Looking at some comments below it seems like the convention here is C contains both. The field is also self explanatory in that sense, so I'd be ok leaving the g out

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/screwTheory/FloatingBaseRigidBodyDynamicsCalculator.java
```diff
@@ -4,133 +4,281 @@
...
 
-   private void computeJacobianTranspose(DMatrixRMaj jacobian, DMatrixRMaj jacobianTransposeToPack)
-   {
-      jacobianTransposeToPack.reshape(jacobian.getNumCols(), jacobian.getNumRows());
-      jacobianTransposeToPack.zero();
-      CommonOps_DDRM.transpose(jacobian, jacobianTransposeToPack);
+      return equalsZero(tempObjective, tolerance);
    }
 
    private static boolean equalsZero(DMatrixRMaj matrix, double tolerance)
```
james-p-foster: Can you not use `MatrixFeatures_DDRM.isZeros`?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/optimization/DynamicsMatrixCalculator.java
```diff
@@ -2,99 +2,149 @@
...
       bodyContactForceJacobianTranspose = new DMatrixRMaj(bodyDoFs, rhoSize);
 
       torqueMinimizationObjective = new DMatrixRMaj(bodyDoFs, 1);
    }
 
-   public void reset()
+   public void setGravity(double gravityZ)
    {
-      coriolisMatrixCalculator.setExternalWrenchesToZero();
+      coriolisMatrixCalculator.setGravitionalAcceleration(-Math.abs(gravityZ));
```
james-p-foster, line 131: Dunno how much this suggestion would ripple over the code, but imo it's a bit sketch to have a negate in here. I see elsewhere we have `gravityZ = 9.81` which is then negated here. I'd personally much rather us have `gravityZ=  -9.81` up top which is just passed straight through here. I'll point out where I see it.

ihmc-common-walking-control-modules/src/test/java/us/ihmc/commonWalkingControlModules/momentumBasedController/optimization/DynamicsMatrixCalculatorTest.java
james-p-foster, line 1: Positive `gravityZ` values showing up in a lot of these tests

james-p-foster (changes requested):

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/MatrixMissingTools.java
```diff
@@ -465,11 +465,41 @@ public static void unsafe_add(DMatrixRMaj matrix, int row, int col, double value
    /**
     * Negates the input matrix in-place.
     *
-    * @param matrix the matrix to be negated.
+    * @param matrixToNegate the matrix to be negated. Modified.
```
stephenmcc: Could use CommonOps.changeSign instead

ihmc-robotics-toolkit/src/test/java/us/ihmc/robotics/MatrixMissingToolsTest.java
```diff
@@ -48,6 +50,33 @@ else if (row == 3 && col == 5)
...
+   {
+      DMatrixRMaj matrixToNegate = new DMatrixRMaj(0, 0);
+      int iters = 100;
+      Random random = new Random(1738L);
+      for (int i = 0; i < iters; i++)
+      {
+         // reinitialize the matrix to random size and value
+         int initializeRows = RandomNumbers.nextInt(random, 100, 1000);
+         int initializeCols = RandomNumbers.nextInt(random, 100, 1000);
+         matrixToNegate.set(RandomMatrices_DDRM.rectangle(initializeRows, initializeCols, random));
```
stefanfasano, line 64: what's the point of this (and the two lines before it) if it just gets overwritten on line 70, are they needed?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controllerCore/WholeBodyInverseDynamicsSolver.java
```diff
@@ -200,40 +200,44 @@ public WholeBodyInverseDynamicsSolver(WholeBodyControlCoreToolbox toolbox, YoReg
...
+      planeContactWrenchProcessor.initialize();
 
-      if (updateDynamicMatrixCalculator.getValue())
+      if (updateDynamicMatrixCalculator.getValue() || useDynamicMatrixCalculatorForInverseDynamics.getBooleanValue())
       {
          dynamicsMatrixCalculator.reset();
       }
       if (!useDynamicMatrixCalculatorForInverseDynamics.getValue())
       {
          inverseDynamicsCalculator.setExternalWrenchesToZero();
```
stefanfasano: should this be inverseDynamicsCalculator.compute()? (I'm operating under the assumption that the logic of this initialize() method is supposed to mostly be identical to before)

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controllerCore/WholeBodyInverseDynamicsSolver.java
```diff
@@ -200,40 +200,44 @@ public WholeBodyInverseDynamicsSolver(WholeBodyControlCoreToolbox toolbox, YoReg
...
+      jointAccelerationIntegrationCalculator.resetJointParameters();
 
-      if (updateDynamicMatrixCalculator.getValue())
+      if (updateDynamicMatrixCalculator.getValue() || useDynamicMatrixCalculatorForInverseDynamics.getBooleanValue())
       {
          dynamicsMatrixCalculator.reset();
       }
       if (!useDynamicMatrixCalculatorForInverseDynamics.getValue())
       {
-         inverseDynamicsCalculator.compute();
```
stefanfasano, line 231: This is what I'm referring to in my comment above

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/screwTheory/FloatingBaseRigidBodyDynamicsCalculator.java
rjgriffin42: I didn't know that was a thing.

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/MatrixMissingTools.java
```diff
@@ -465,11 +465,41 @@ public static void unsafe_add(DMatrixRMaj matrix, int row, int col, double value
...
     */
-   public static void negate(DMatrixRMaj matrix)
+   public static void negate(DMatrixRMaj matrixToNegate)
    {
-      CommonOps_DDRM.scale(-1.0, matrix);
```
stefanfasano, line 472: what's the reasoning for not using this in the new negate_unsafe() method?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/optimization/DynamicsMatrixCalculator.java
rjgriffin42: well it's to force the gravity vector to be less than 0.0, since it's coupled with a Math.abs. It's to counter the problem you're talking about, where sometimes it's stored as a positive value in the code, which doesn't make sense. 

I think that's likely o ut of scope of this PR, because to do it the right way, I'd want to make sure everywhere was sign dependent.

james-p-foster: Thought as much -- happy to disregard this then

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/optimization/DynamicsMatrixCalculator.java
rjgriffin42: I went through and did a naming cleanup on it to make sure everywhere it says gravity and coriolis to match the calculator, and also changed it to a vector per your next suggestion.

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/MatrixMissingTools.java
rjgriffin42: I didn't know that was a thing, so that's pretty cool. I went through and replaced the base use of negate with that. I did keep (and rename) negate_unsafe, because that's useful when you don't want to do it in-place.

ihmc-robotics-toolkit/src/test/java/us/ihmc/robotics/MatrixMissingToolsTest.java
rjgriffin42: it's to initialize the backing data array to something of a different size, and with random data. It may be that it ends up being much larger. overwriting with mroe information only overwrites part of the data array.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controllerCore/WholeBodyInverseDynamicsSolver.java
rjgriffin42: This didn't actually change, it just looks like it did in the diff because git went stupid.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controllerCore/WholeBodyInverseDynamicsSolver.java
rjgriffin42: see below.

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/MatrixMissingTools.java
rjgriffin42: Efficiency, in this case. 

The scale operator performs a multiply, (a = -b * a) while the unsafe_changeSign just says (a = -a). Also, it's "unsafe" because it doesn't check that the results matrix has the same size, which is a minor speed up, but does make a big of a difference. This can result in an out of bounds exception, though, which is why we called it unsafe, similar to the standard EJML unsafe_get and unsafe_set.

stefanfasano (approved):

ihmc-robotics-toolkit/src/test/java/us/ihmc/robotics/MatrixMissingToolsTest.java
```diff
@@ -48,6 +49,36 @@ else if (row == 3 && col == 5)
       }
    }
 
+   @Test
+   public void testNegateUnsafe()
```
james-p-foster, line 53: Probably want to rename the test now it's change sign?

james-p-foster (approved):

PR #382, opened 2024-10-03
better privileged configuration parameters

LuigiPenco93:
reverted changes in parameters done for boxing. Decreased null space projection alpha to what we were using before  boxing


rjgriffin42 (commented):
This looks good. Only comment is as to whether we have some nadia specific values?

LuigiPenco93:
> This looks good. Only comment is as to whether we have some nadia specific values?

yes those have been left unchanged. I don't know why Sylvain had changed these generic values to be honest. My guess is that he changed these when testing stuff with high speed without Nadia and forgot to discard

Beomyeong (approved):

PR #387, opened 2024-10-04
Feature/extract parameter tuner

rjgriffin42:
This pushes some of the code into ihmc-javafx-toolkit. it then allows us to restructure the code to make ihmc-parameter-tuner only dependent on jars, so it can be extracted.


calvertdw (approved):
Sure, this code is basically dead to me personally anyway.

ds58 (approved):

PotatoPeeler3000 (approved):

PR #400, opened 2024-10-10
Remove ihmc-perception dependency on whole body, common walking.

calvertdw:


ihmc-perception/src/main/java/us/ihmc/perception/realsense/example/L515SimpleSnapperExample.java
```diff
@@ -34,7 +33,7 @@
 import us.ihmc.yoVariables.euclid.referenceFrame.YoFramePoseUsingYawPitchRoll;
 import us.ihmc.yoVariables.registry.YoRegistry;
 
-public class L515SimpleSnapperExample implements FootstepAdjustment, L515DepthImageReceiver
```
calvertdw, line 37: This whole class is unused.

rjgriffin42:
I'm thinking this isn't the solution that we want.

rjgriffin42 (changes requested):
We can't remove the implementation of FootstepAdjustment. See slack conversation.

PR #404, opened 2024-10-11
Bugfix/fixing branches

rjgriffin42:


ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/humanoid/kinematicsBasedStateEstimation/IMUBasedPelvisRotationalStateUpdater.java
```diff
@@ -41,7 +41,7 @@ public class IMUBasedPelvisRotationalStateUpdater implements PelvisRotationalSta
    private final YoFrameVector3D yoRootJointAngularVelocity;
    private final YoFrameVector3D yoRootJointAngularVelocityInWorld;
 
-   private final BooleanParameter zeroYawAtInitialization = new BooleanParameter("zeroEstimatedRootYawAtInitialization", registry, true);
+   private final BooleanParameter zeroYawAtInitialization = new BooleanParameter("zeroEstimatedRootYawAtInitialization", registry, false);
```
rjgriffin42, line 44: This and the other state estimator variable got changed by default. This in turn messed up every test of atlas at a non-zero initialization. You should probably go in and in yoru parameter file for the state estimator that uses this, change it back to true, but not edit the default value for your specific change.

stefanfasano (approved):

christiandebuys:
Not sure if they are issues that are supposed to be considered in this PR, but I noticed 2 things:

1. AtlasFlatGroundQuickWalkingTest errors out for me because there is no fast walking parameters XML at the specified path. Is there supposed to be?
Caused by: java.lang.IllegalArgumentException: resources\us\ihmc\atlas\fast_walking_parameters.xml (The system cannot find the path specified)

2. NadiaFlatGroundQuickWalking test fails because the robot falls.

christiandebuys (approved):
The missing xml and failing test are out of scope for this PR.

Beomyeong (approved):

PR #412, opened 2024-10-17
Ensure LevelbergMarquard test doesnt create a JFrame if visualize is …

ds58:
…set to false


rjgriffin42:
I've actually got anotehr PR that fixxes this test, including the algorithm side

PR #425, opened 2024-10-23
Add LoopingThread, Deprecate Restartable Threads

TomaszTB:
Restartable threads have a few issues with them: 
* The name is misleading. It doesn't even hint at the fact that the thread runs the `Runnable` in a loop, and the thread isn't truly "restartable;" it just starts a new thread when it's stopped and restarted.
* Duplicate code. `RestartableThread` and `RestartableThrottledThread` are nearly identical, except one has a `Throttler`. Not sure why `RestartableThrottledThread` doesn't just extend `RestartableThread`.
* The `runOnce()` method is handy, but it runs the `Runnable` synchronously, where I'd expect it to run on the thread. 
* There is thread creation overhead every time `stop()` + `start()` is called. This is a minor performance issue.

Some silly person must've written this code. 

`LoopingThread` improves upon the restartable threads by: 
* Having a better name, more indicative of what it does
* Consolidating throttling functionality into one class
* Having a `runOnce()` method that actually runs asynchronously. 
* Only starting one thread, and simply pausing the loop until `start` is called again.

This PR deprecates the restartable threads, and replaces the tests with `LoopingThreadTest`.


ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
```diff
@@ -0,0 +1,193 @@
...
+ * A thread that executes code in a loop.
+ * Optionally, you may set a limit to the loop frequency.
+ * This thread does not finish running until {@link #close()} or {@link #blockingClose()} is called.
+ * Once started, make sure to close this thread.
+ */
+public class LoopingThread extends Thread
+{
+   private final Runnable target;
+   private final Throttler throttler = new Throttler();
+   private volatile double loopPeriodLimit = -1.0;
```
calvertdw: loopPeriodLowerLimit?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
```diff
@@ -0,0 +1,193 @@
...
+
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * A thread that executes code in a loop.
+ * Optionally, you may set a limit to the loop frequency.
+ * This thread does not finish running until {@link #close()} or {@link #blockingClose()} is called.
+ * Once started, make sure to close this thread.
+ */
+public class LoopingThread extends Thread
```
calvertdw: I would recommend not extending thread. It makes a bunch of public erasure and methods that are not part of the API of this class. It should be kept private internally.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
```diff
@@ -0,0 +1,193 @@
...
+   }
+
+   public LoopingThread(Runnable target, double loopFrequencyLimit, String name)
+   {
+      super(target, name);
+      this.target = target;
+      limitLoopFrequency(loopFrequencyLimit);
+   }
+
+   public LoopingThread(RunnableThatThrows targetThatThrows, ExceptionHandler exceptionHandler, String name)
```
calvertdw: runnableThatThrows?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
```diff
@@ -0,0 +1,193 @@
...
+         runInLoop();
+      }
+   }
+
+   public boolean isLooping()
+   {
+      return runCommand.get() == RunCommand.LOOP;
+   }
+
+   private enum RunCommand
```
calvertdw: Can you declare this at the top of the file?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
```diff
@@ -0,0 +1,193 @@
...
+   /**
+    * Pauses re-executing the {@link #runInLoop()} method upon its completion.
+    */
+   public void pause()
+   {
+      runCommand.set(RunCommand.PAUSE);
+   }
+
+   /**
+    * Finalizes the execution of {@link #runInLoop()} and stops the thread.
```
calvertdw: What does finalize mean here?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
```diff
@@ -0,0 +1,193 @@
...
+      catch (InterruptedException ignored) {}
+   }
+
+   /**
+    * The method that is executed in a loop.
+    * You may either provide this class a
+    * {@link Runnable}/{@link RunnableThatThrows} to run,
+    * or override this method.
+    */
+   public void runInLoop()
```
calvertdw: I think this method name is pretty ambiguous for the public API. If some methods do stuff in the user's thread, the ones that don't should have *Async. Like runOnceAsync()

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
calvertdw: What you're going for seems a lot like what I had done with PausablePeriodicThread. That name seems like it would be a better choice than LoopingThread.

ds58: I prefer that it extends Thread. It makes expectations clearer and you can treat it as standard Java API.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
```diff
@@ -0,0 +1,193 @@
...
+   }
+
+   public boolean isLooping()
+   {
+      return runCommand.get() == RunCommand.LOOP;
+   }
+
+   private enum RunCommand
+   {
+      INITIALIZE, LOOP, PAUSE, RUN_ONCE, FINALIZE
```
calvertdw: Can we remove INITIALIZE and use PAUSE at the beginning?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
```diff
@@ -0,0 +1,193 @@
...
+   public void pause()
+   {
+      runCommand.set(RunCommand.PAUSE);
+   }
+
+   /**
+    * Finalizes the execution of {@link #runInLoop()} and stops the thread.
+    * The thread cannot be re-started after calling this method.
+    */
+   public void close()
```
calvertdw: What about destroy which makes it clear in one word that it can't be used anymore after calling?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
ds58: I don't understand this comment

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
calvertdw: If you want to remove PausablePeriodicThread and replace it with this, I would be in favor. I don't like that that one uses the ExceptionHandlingThreadScheduler.

ds58: Yeah this is pretty similar to PausablePeriodicThread. Should we deprecate that one in favor of this one. I personally don't have a problem with the name LoopingThread.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
ds58: close() is convention in Java. I.e. Closable, AutoClosable

calvertdw: Neither Thread, nor this class extends Closable. This is just using "close" as a word.  That word holds no meaning in this context, like it would for a file or stream.

calvertdw: I would be okay with others, like dispose, endThread, etc

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: Which methods are you thinking of? I looked through the methods that `Thread` offers and I didn't see any that I wouldn't want this class to have, besides the deprecated ones. I'm not seeing the downside to extending `Thread`

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: I was trying to convey that he `runInLoop()` method will finish executing for the last time

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: Are you suggesting `runInLoopAsync()`? I was trying to keep it close to the `Thread` API (`run()`) so it'd be familiar to the user of this class

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
calvertdw: Well you can't guarantee the thing works because the user can call all those things like yield and interrupt, etc. at any time, and it'll break the assumptions made by the API you added. I would say for the advanced user case, you could accept a user created Thread in the constructor so they could operate on it externally.

TomaszTB: Hmmm I'm not sure if I'd call this "periodic" since it can free spin the loop. I suppose there's still some period that the loop runs in, but it wouldn't be defined by the programmer.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: INITIALIZE indicates that the thread has to be created, PAUSE indicates the thread is paused. I'd need some sort of indication that the thread needs to be created if I remove INITIALIZE

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
calvertdw: The main one here is Thread.run(). That's really confusing that that would be available as a method on this class, but you really don't want to be calling that.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: I think within the context of Java `close()` also indicates that the object cannot be used anymore (like `Closable`, `AutoClosable`). I've been exploring using `close()` over `destroy()` to stick with Java convention. If you prefer `destroy()` I can change it to that

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: Yeah, I agree that having `run()` is a bit weird. But you don't really want to call that even with a normal `Thread`. You probably don't want to call `runInLoop()` of this class either. I think it's just a price to pay for enabling the user to extend the class to add custom functionality.

calvertdw: I think Java's Thread is pretty badly designed in this respect, but it can't be fixed because it would break too much. I feel pretty strongly that this class should not extend thread. What about providing getThread() to access the underlying thread? Is there a use case I'm missing?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
calvertdw: I think what you just said would be way more clear

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
calvertdw: Having thought more, I'm suggesting first to make this method private.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
calvertdw: Fair. Would PausableLoopingThread be better than LoopingThread?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
calvertdw: Why do you need to delay the creation of the thread? Can't you do it on construction?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: Sure, I'm cool with that

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: That wouldn't allow extending the method, which is part of my goal with this class.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: You're right, I could. I was trying to keep the semantics of `start()` similar to `Thread`, where it would create the thread when `start()` is called. But I suppose there's no harm in creating the thread upon construction.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
calvertdw: What convention are you citing? The documentation of Closable and AutoClosable talk about releasing resources like files and socket handles. I think you're confusing it with dispose (Disposable) which is more for this purpose. 

Regardless, Java's Thread doesn't require any cleanup, it just ends. So the name of this method just needs to convey that the class can't be used anymore if this is called. What about exitForever()?

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/Closeable.html
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/AutoCloseable.html
https://docs.oracle.com/middleware/12211/coherence/java-reference/com/oracle/common/base/Disposable.html

calvertdw: Okay, so a Thread is a resource until it ends, right? So what about adding implements Disposable and using `dispose()` here?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: I think that'd be fine functionality wise, but I'd find it weird that `LoopingThread` isn't a `Thread`. I'll make that change for now.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
calvertdw: Oh, I think I missed that as a goal of this class. I think you should require the runnableThatThrows be passed in at construction and make it final.

TomaszTB: Do you still mean to remove the `runInLoop()` method?

calvertdw: Yes, in that case it wouldn't be needed?

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: Sure, I can make it a `Disposable`

TomaszTB: Oh, we don't actually have the `Disposable` in the dependencies of `ihmc-java-toolkit`. 

I'm down to implement `AutoClosable` though. `Disposable` is just an `AutoClosable` which adds a method to call instead of `close()` when releasing resources.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/LoopingThread.java
TomaszTB: I just realized why it's needed. Without a `runInLoop()` method it's difficult to extend `PausableLoopingThread`. 

E.g. a common way to use `Thread` is like: 
```
public class MyClass extends Thread
{
   private final Param myParam;

   public MyClass(/* parameters */)
   {
      this.myParam = myParam;
      ...
   }

   @Override
   public void run()
   {
      // Do something with myParam
   }
}
```

I thought perhaps it'd be possible to do something like:
```
public class MyClass extends PausableLoopingThread
{
   private final Param myParam;
 
   public MyClass(/* params */)
   {
      super(this::run, /* other params */);
      this.myParam = myParam;
   }

   private void run()
   {
      // Do something with myParam
   }
}
```

BUT I don't have access to `this` before the call to `super()`. The natural way to extend `PausableLoopingThread` would be like `Thread`, to override the method that it runs on a thread. 

Perhaps there's a different hack to going around this, but now I feel that the class should have a `runInLoop()` method which can be overridden.

ihmc-java-toolkit/src/main/java/us/ihmc/tools/thread/PausableLoopingThread.java
```diff
@@ -0,0 +1,181 @@
...
+
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * A thread that executes code in a loop.
+ * Optionally, you may set a limit to the loop frequency.
+ * This thread does not finish running until {@link #destroy()} or {@link #blockingDestroy()} is called.
+ * Once started, make sure to destroy this thread.
+ */
+public class PausableLoopingThread
```
ds58: I still think this should extend thread

ds58 (changes requested):

TomaszTB:
Just a quick summary of what was discussed on Slack:

- `PausableLoopingThread` should extend `Thread`
  - `run()` method of `Thread` will not be used. I'll add a stern warning against using it in Javadoc
- `runInLoop()` method will be protected, to allow overriding it, but not calling it
- Remove the `RunCommand` enum, and instead use `int` for how many times to run, and `isDestroyed` boolean to signal the loop to end. 
  - N > 0 -> loop N times
  - N = -1 -> keep looping until told otherwise
  - N = 0 -> pause
- Add Javadoc explaining the reasoning for design choices

ds58 (approved):

TomaszTB:
Going to move this to `ihmc-commons`

PR #427, opened 2024-10-24
Remove boofcv

ds58:


ihmc-humanoid-robotics/src/main/java/us/ihmc/humanoidRobotics/communication/packets/HumanoidMessageTools.java
```diff
@@ -2416,38 +2418,40 @@ public static void checkRobotSide(HumanoidBodyPart bodyPart)
          throw new RuntimeException("Need to provide robotSide for the bodyPart: " + bodyPart);
    }
 
-   public static IntrinsicParametersMessage toIntrinsicParametersMessage(CameraPinholeBrown intrinsicParameters)
+   @Deprecated
+   public static IntrinsicParametersMessage toIntrinsicParametersMessage(Object intrinsicParameters)
```
calvertdw, line 2422: I think these are importantly used? Would be worth switching it to the CameraIntrinsics

ihmc-perception/build.gradle.kts
```diff
@@ -26,11 +26,6 @@ mainDependencies {
 
    api("org.georegression:georegression:0.22")
    api("net.java.dev.jna:jna:4.1.0")
-   api("org.boofcv:boofcv-ip:0.36")
-   api("org.boofcv:boofcv-swing:0.36")
-   api("org.boofcv:boofcv-io:0.36")
-   api("org.boofcv:boofcv-recognition:0.36")
-   api("org.boofcv:boofcv-calibration:0.36")
```
calvertdw, line 33: :+1:

calvertdw:
I would expect some changes to dependent repos. Are there? Otherwise, this looks good to me.

ihmc-humanoid-robotics/src/main/java/us/ihmc/humanoidRobotics/communication/packets/HumanoidMessageTools.java
ds58: ihmc-perception/CameraIntrinsics aren't available here and I don't want to add that dependency. These are not used in anything relevant.

ds58:
> I would expect some changes to dependent repos. Are there? Otherwise, this looks good to me.

I checked and they're arent any other repos which need changes

calvertdw (approved):

atlas/src/main/java/us/ihmc/atlas/multisenseBlobExperiments/MultiSenseBlobDetectionTestModule.java
```diff
@@ -66,17 +65,17 @@ public void run()
             while (latestBufferedImage == null)
                ThreadTools.sleep(10);
 
-            ImagePanel imagePanel = ShowImages.showWindow(latestBufferedImage, "Circle Detector");
-            JFrame frame = (JFrame) SwingUtilities.getWindowAncestor(imagePanel);
+//            ImagePanel imagePanel = ShowImages.showWindow(latestBufferedImage, "Circle Detector");
+//            JFrame frame = (JFrame) SwingUtilities.getWindowAncestor(imagePanel);
```
rjgriffin42, line 69: What's going on here? I would prefer enclosing these in a boolean of "visualize" or whatever, rather than commenting them out. Or is this to remove the JFrame dependency?

ihmc-humanoid-robotics/src/main/java/us/ihmc/humanoidRobotics/communication/packets/HumanoidMessageTools.java
rjgriffin42: Should it be removed then?

rjgriffin42 (commented):
I am personally not a fan of commenting out large swathes of code to remove a dependency. That gets pretty hard to revert if you want to. If you don't want to revert, then the functionality should be deleted. 

I would be fine if the plan is to migrate this to the Operator interface package, and then uncomment the code, but would need help with doing that.

atlas/src/main/java/us/ihmc/atlas/multisenseBlobExperiments/MultiSenseBlobDetectionTestModule.java
ds58: ImagePanel and ShowImages come from boofcv.

ihmc-humanoid-robotics/src/main/java/us/ihmc/humanoidRobotics/communication/packets/HumanoidMessageTools.java
ds58: If I remove it, it kind of explodes and I would have to end up deleting a lot more things

ds58:
> I am personally not a fan of commenting out large swathes of code to remove a dependency. That gets pretty hard to revert if you want to. If you don't want to revert, then the functionality should be deleted.
> 
> I would be fine if the plan is to migrate this to the Operator interface package, and then uncomment the code, but would need help with doing that.

See my comment:
> If I remove it, it kind of explodes and I would have to end up deleting a lot more things

This was the simplest way right now to remove this dependency.

TomaszTB (approved):
> If I remove it, it kind of explodes and I would have to end up deleting a lot more things

I looked into it a bit and there are a few low hanging fruit, but indeed it explodes very quickly. That also means that this PR will break a lot of (old, unused) code in the repo (if it isn't broken already). 

Should we plan to remove all that stuff at some point? I don't really like the idea of increasing the amount of broken, unused code long term. 

I think this change is good for now though.

PR #428, opened 2024-10-24
Add CameraModel + getDataPointer Methods to RawImage

TomaszTB:


ihmc-perception/src/main/java/us/ihmc/perception/RawImage.java
```diff
@@ -224,11 +267,26 @@ public GpuMat getGpuImageMat()
       return gpuImageMat;
    }
 
+   public BytePointer getDataPointer()
+   {
+      return getCpuImageMat().data();
+   }
+
+   public BytePointer getCUDADataPointer()
```
ds58, line 315: Maybe add a comment that because this calls getGpuImageMat, it allocates CUDA memory if it didn't exist

ds58 (approved):

PR #430, opened 2024-10-24
Remove onnx

ds58:
This onnx dependency is ~250MB, which is very large. The only place it was used is in `FootstepPredictor` and `HeightMapAutoencoder`. HeightMapAutoencoder had a main and was not used anywhere else. FootstepPredictor was used in this demo UI `NadiaRDXTerrainPlanningUI`. To be honest, I tried running that UI but everything was broken and I didn't want to spend time figuring it out or fixing it. 

I dont really see us using these modules as they exist currently. Ideally, we should use pytorch/opencv anyway since we heavily use that everywhere else in the code base. I don't know the exact reason onnx was used here.

### DIFF
--- a/ihmc-perception/src/main/java/us/ihmc/perception/neural/FootstepPredictor.java
+++ b/ihmc-perception/src/main/java/us/ihmc/perception/neural/FootstepPredictor.java
@@ -33,7 +33,10 @@ public class FootstepPredictor
    private static final int IMAGE_INPUT_HEIGHT = 201;
    private static final int IMAGE_INPUT_WIDTH = 201;
 
-   // TODO: We should use pytorch/opencv instead of ONNX. ONNX is a huge dependency we don't use anywhere else and we already have opencv and heavily use that.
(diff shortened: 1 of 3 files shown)


calvertdw:
To be clear, we do ingest ONNX files, we just feed them into the OpenCV implementation instead of the Microsoft one used here.

ihmc-perception/src/main/java/us/ihmc/perception/neural/FootstepPredictor.java
```diff
@@ -33,131 +33,133 @@ public class FootstepPredictor
    private static final int IMAGE_INPUT_HEIGHT = 201;
    private static final int IMAGE_INPUT_WIDTH = 201;
 
+   // TODO: We should use pytorch/opencv instead of ONNX. ONNX is a huge dependency we don't use anywhere else and we already have opencv and heavily use that.
```
calvertdw: I think you should be clear that ONNX models can still be used, just fed into OpenCV instead of the Microsoft runtime.

calvertdw (approved):

PR #434, opened 2024-10-25
Fix AvatarPostProcessingTests

PotatoPeeler3000:
The Planar Regions weren't getting converted into the request, now we convert the planar regions into a height map and pass that into the request so the planner can use it

### DIFF
--- a/ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/networkProcessor/footstepPostProcessing/AvatarPostProcessingTests.java
+++ b/ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/networkProcessor/footstepPostProcessing/AvatarPostProcessingTests.java
@@ -16,6 +16,7 @@ import org.junit.jupiter.api.Test;
 import controller_msgs.msg.dds.ArmTrajectoryMessage;
 import controller_msgs.msg.dds.FootstepDataListMessage;
 import controller_msgs.msg.dds.FootstepDataMessage;
+import perception_msgs.msg.dds.HeightMapMessage;
 import toolbox_msgs.msg.dds.FootstepPlanningRequestPacket;
 import controller_msgs.msg.dds.OneDoFJointTrajectoryMessage;
 import ihmc_common_msgs.msg.dds.TrajectoryPoint1DMessage;
@@ -54,6 +55,7 @@ import us.ihmc.footstepPlanning.FootstepPlanningModule;
 import us.ihmc.footstepPlanning.PlannedFootstep;
 import us.ihmc.footstepPlanning.graphSearch.parameters.DefaultFootstepPlannerParametersBasics;
 import us.ihmc.footstepPlanning.swing.SwingPlannerType;
+import us.ihmc.footstepPlanning.tools.PlanarRegionToHeightMapConverter;
 import us.ihmc.footstepPlanning.tools.PlannerTools;
 import us.ihmc.graphicsDescription.appearance.YoAppearance;
 import us.ihmc.graphicsDescription.yoGraphics.YoGraphicPosition;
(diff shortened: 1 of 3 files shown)


ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/networkProcessor/footstepPostProcessing/AvatarPostProcessingTests.java
```diff
@@ -398,6 +400,9 @@ private static FootstepPlanningRequestPacket getRequest(FullHumanoidRobotModel f
       request.getStartLeftFootPose().set(leftFoot);
       request.getStartRightFootPose().set(rightFoot);
 
+      HeightMapMessage heightMap = PlanarRegionToHeightMapConverter.convertFromPlanarRegionsToHeightMap(planarRegionsList);
```
PotatoPeeler3000, line 403: This adds the planar regions to the footstep planner request

ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/AStarFootstepPlanner.java
```diff
@@ -72,7 +72,7 @@ public class AStarFootstepPlanner
    private final SwingPlanningModule swingPlanningModule;
 
    /** Called each iteration. Should be very lightweight, mainly used for variable copying for the logger */
-   private List<Consumer<AStarIterationData<FootstepGraphNode>>> iterationCallbacks = new ArrayList<>();
+   private final List<Consumer<AStarIterationData<FootstepGraphNode>>> iterationCallbacks = new ArrayList<>();
```
PotatoPeeler3000, line 75: Recommended

ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/AStarFootstepPlanner.java
```diff
@@ -302,13 +302,10 @@ private void reportStatus(FootstepPlannerRequest request, FootstepPlannerOutput
...
       }
 
-      if (!request.getAssumeFlatGround())
-      {
-         swingPlanningModule.computeSwingWaypoints(request.getHeightMapData(),
-                                                   outputToPack.getFootstepPlan(),
-                                                   request.getStartFootPoses(),
-                                                   request.getSwingPlannerType());
-      }
+      swingPlanningModule.computeSwingWaypoints(request.getHeightMapData(),
```
PotatoPeeler3000, line 305: We don't need to check if we are walking on flat ground, inside `computerSwingWaypoints` if the height map is null it returns, so its fine to always try and computer the swing waypoints

ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/SwingPlanningModule.java
```diff
@@ -67,12 +67,13 @@ public void computeSwingWaypoints(HeightMapData heightMapData,
                                      SideDependentList<? extends Pose3DReadOnly> startFootPoses,
                                      SwingPlannerType swingPlannerType)
    {
-      swingTrajectories.clear();
       if (heightMapData == null || heightMapData.isEmpty())
       {
          return;
       }
 
+      swingTrajectories.clear();
```
PotatoPeeler3000: Clear after if we have a height map

ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/AStarFootstepPlanner.java
```diff
@@ -302,13 +302,10 @@ private void reportStatus(FootstepPlannerRequest request, FootstepPlannerOutput
...
+                                                outputToPack.getFootstepPlan(),
+                                                request.getStartFootPoses(),
+                                                request.getSwingPlannerType());
```
rjgriffin42, line 308: why did this get pulled out of its `if` case?

ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/SwingPlanningModule.java
rjgriffin42: why move this here? In this case, you'll hold onto the old swing trajectories if there's no height map data passed in, which is probably undesirabale behavior. Or is there another reason?

PotatoPeeler3000: Talked in person, could create some weird edge cases, so will move the clear back

ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/AStarFootstepPlanner.java
PotatoPeeler3000: Because if the height map is null it returns from that method, and its not a large computation so we can always check to see if we have a height map, and compute swing waypoints if needed

rjgriffin42 (approved):

PR #442, opened 2024-10-31
Fix out of order RDXROS2ColoredPointCloudVisualizer destroy()

ds58:


TomaszTB (approved):
Huh I never experienced this issue, but the change makes sense!

PR #446, opened 2024-11-05
Add CUDAKernel To Simplify Setting Parameters

TomaszTB:
Now, instead of doing something like: 
```
cudaProgram.loadKernel("add"); // load the kernel

// Create parameter pointers
IntPointer a = new IntPointer(1L).put(3);
IntPointer b = new IntPointer(1L).put(7);
IntPointer sum = new IntPointer();
PointerPointer<IntPointer> sumPointer = new PointerPointer(1L).put(sum
// allocate sum to device...

cudaProgram.run(/* other params */, a, b, sumPointer); // run the loaded kernel
```

You can do something like: 
```
CUDAKernel kernel = cudaProgram.loadKernel("add"); // load the kernel

// Create pointer parameters
IntPointer sum = new IntPointer();
// allocate sum to device...

kernel.withInt(3).withInt(3).withPointer(sum).run(/* other params */);
```

### DIFF
--- a/ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
+++ b/ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
@@ -1,0 +1,117 @@
+package us.ihmc.perception.cuda;
+
+import org.bytedeco.cuda.cudart.CUfunc_st;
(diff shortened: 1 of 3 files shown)


ds58 (approved):
I like the simplification of setting the parameters. Looks good to me.

PR #448, opened 2024-11-06
Reduce use of BytedecoImage in ArUco code

TomaszTB:
As part of the Perception Autonomy Cleanup effort, I'd like to reduce the usage of `BytedecoImage` where it's not needed. In a lot of cases, a simple `Mat` suffices. 

I've tested the changes using the BehaviorUI, and ZED ArUco Demo. Also ran the ArUco Marker Detection Demo, but I can't tell if it's working correctly (though there's no difference compared to develop).

### DIFF
--- a/ihmc-high-level-behaviors/src/test/java/us/ihmc/rdx/simulation/scs2/RDXArUcoMarkerDetectionDemo.java
+++ b/ihmc-high-level-behaviors/src/test/java/us/ihmc/rdx/simulation/scs2/RDXArUcoMarkerDetectionDemo.java
@@ -60,7 +60,7 @@ public class RDXArUcoMarkerDetectionDemo
             baseUI.getPrimaryScene().addRenderableProvider(cameraSensor::getRenderables);
 
             arUcoMarkerDetector = new OpenCVArUcoMarkerDetector();
-            arUcoMarkerDetector.setSourceImageForDetection(cameraSensor.getLowLevelSimulator().getRGBA8888ColorImage());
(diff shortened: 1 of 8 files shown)


ihmc-high-level-behaviors/src/test/java/us/ihmc/rdx/simulation/scs2/RDXArUcoMarkerDetectionDemo.java
```diff
@@ -60,7 +60,7 @@ public void create()
             baseUI.getPrimaryScene().addRenderableProvider(cameraSensor::getRenderables);
 
             arUcoMarkerDetector = new OpenCVArUcoMarkerDetector();
-            arUcoMarkerDetector.setSourceImageForDetection(cameraSensor.getLowLevelSimulator().getRGBA8888ColorImage());
+            arUcoMarkerDetector.setSourceImageForDetection(cameraSensor.getLowLevelSimulator().getRGBA8888ColorImage().getBytedecoOpenCVMat());
```
calvertdw, line 63: Do you think at this point we could rename `getBytedecoOpenCVMat()` to just `getMat()`?

calvertdw (approved):

ihmc-high-level-behaviors/src/test/java/us/ihmc/rdx/simulation/scs2/RDXArUcoMarkerDetectionDemo.java
TomaszTB: Probably. I think that's out of scope of this PR though

PR #451, opened 2024-11-07
Fix ConvexPolygonToolsTest#testAllMethodsForPolygonWithTwoPoints

calvertdw:


ihmc-robotics-toolkit/src/test/java/us/ihmc/robotics/geometry/ConvexPolygonToolsTest.java
```diff
@@ -378,6 +377,28 @@ private void plotPolygon(FrameConvexPolygon2D polygonWithLimitedVertices, int de
       testFrame.dispose();
    }
 
+   private void plotPolygons(ConvexPolygon2D polygonA, ConvexPolygon2D polygonB)
```
calvertdw, line 380: This was used to provide the following visualization as an insight to fixing this test.
image

PotatoPeeler3000:
Changing the asserts just to get the test to pass is not what we should be doing. There is an actual bug between `Euclid` and the `ConvexPolygonTools` that should be fixed. Why would `Euclid` return 1 point and the `ConvexPolygonTools` return 2? They should be identical right?

However I'm not sure if they should be returning 1 point or 2 points,

ihmc-robotics-toolkit/src/test/java/us/ihmc/robotics/geometry/ConvexPolygonToolsTest.java
```diff
@@ -739,8 +760,14 @@ public void testAllMethodsForPolygonWithTwoPoints()
          }
          else if (expectedIntersectionWithSparePolygon.length == 1)
          {
-            assertTrue(actualIntersectionWithSparePolygon.getNumberOfVertices() == 1);
-            assertTrue(expectedIntersectionWithSparePolygon[0].epsilonEquals(actualIntersectionWithSparePolygon.getVertex(0), epsilon));
+            // EuclidGeometryPolygonTools#intersectionBetweenLineSegment2DAndConvexPolygon2D
+            // will return the one point on the edge, where
+            // ConvexPolygonTools#computeIntersectionOfPolygons will return both
+            // the point on the edge and the point inside
+            assertTrue(actualIntersectionWithSparePolygon.getNumberOfVertices() == 2);
```
Beomyeong, line 767: I agree with you. `actualIntersectionWithSparePolygon` is 2d so, even the intersection is one point, it still has two values
I don't think the test failures come from the function itself. I guess it comes from the test itselt and this seems reasonable for me.

Beomyeong (approved):

PR #455, opened 2024-11-12
Remove ManagedROS2Node

ds58:


calvertdw (approved):

calvertdw:
I think it might be important that Luigi and Stephen sign off on this. I don't know how safety critical that would be.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/modules/ToolboxModule.java
```diff
@@ -65,9 +63,7 @@ public abstract class ToolboxModule implements CloseableAndDisposable
    protected final String robotName;
    protected final FullHumanoidRobotModel fullRobotModel;
 
-   private final boolean manageROS2Node;
```
TomaszTB, line 68: This boolean seems unrelated to the ManagedROS2Node (despite its name). Should it really be removed?

ds58: It didn't seem important to me. But, I've put it back just to reduce changes if they are unnecessary.

TomaszTB (approved):
Seems good to me, but I'm not an expert on this code

PR #456, opened 2024-11-12
Remove usages of ROS2Callback

ds58:


calvertdw (approved):

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/pointCloud/RDXROS2PointCloudVisualizer.java
```diff
@@ -76,11 +76,11 @@ public RDXROS2PointCloudVisualizer(String title, ROS2Node ros2Node, ROS2Topic<?>
...
-         if (isActive && ros2Callback == null)
+         if (isActive && !subscribed)
          {
             subscribe();
          }
-         else if (!isActive && ros2Callback != null)
+         else if (!isActive && subscribed)
          {
             unsubscribe();
          }
```
TomaszTB, line 86: I think there might be a bug here because `subscribed` becomes `true` when `subscribe()` is called, but doesn't get set to `false` in `unsubscribe()`. Maybe I'm wrong, but could you double check that this works properly?

TomaszTB (commented):
Just noticed one little thing.

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/pointCloud/RDXROS2PointCloudVisualizer.java
ds58: Oh, yeah I need to check that. Good catch

ds58: It does actually get set to false, not sure if you're talking about something else?

ds58: I was looking at the wrong class. You were right. I've pushed a fix.

TomaszTB (approved):

PR #457, opened 2024-11-12
[Feature] Created and documented a CUDA example

PotatoPeeler3000:


TomaszTB (approved):
I wrote parts of it so of course I approve :rofl:

ihmc-perception/src/test/java/us/ihmc/perception/cuda/ExampleCUDAKernel.java
```diff
@@ -0,0 +1,119 @@
...
+   // Even if you don't create a stream, CUDA will create a default one for you, so we make one ourselves to handle things better
+   public ExampleCUDAKernel()
+   {
+      // Note this name does NOT have to match the name of the program, however for readability its ideal to have them match.
+      String userFriendlyNameOfProgram = "userFriendlyNameOfProgram";
+      // Note this name DOES have to match the name of the CUDA kernel you want to run
+      String kernelName = "add_arrays";
+
+      // We create a stream to synchronize the method calls that happen on the GPU.
+      CUstream_st stream = new CUstream_st();
```
ds58, line 39: Is there a reason you don't use `CUDAStreamManager` here?

ihmc-perception/src/test/java/us/ihmc/perception/cuda/ExampleCUDAKernel.java
```diff
@@ -0,0 +1,119 @@
...
+         void add_arrays(int n, float *x, float *y)
+         {
+            int index = blockIdx.x * blockDim.x + threadIdx.x;
+            int stride = blockDim.x * gridDim.x;
+            for (int i = index; i < n; i += stride)
+               y[i] = x[i] + y[i];
+         }
+         """;
+
+   // Even if you don't create a stream, CUDA will create a default one for you, so we make one ourselves to handle things better
```
ds58, line 30: What's this comment for? It seems like it shouldnt be here

ihmc-perception/src/test/java/us/ihmc/perception/cuda/ExampleCUDAKernel.java
```diff
@@ -0,0 +1,119 @@
...
+
+import static org.bytedeco.cuda.global.cudart.*;
+
+/**
+ * This is a simple example of a kernel that adds two arrays together. The ways in which kernels can be run vary such that you can create all the JavaCPP Pointers
+ * inside a try-with-resources. The user can create more threads on the GPU. This example attempted to keep things really simple and avoided most of that overhead.
+ */
+public class ExampleCUDAKernel
+{
+
```
ds58: Can you remove the extra line break?

ds58 (changes requested):

ihmc-perception/src/test/java/us/ihmc/perception/cuda/ExampleCUDAKernel.java
PotatoPeeler3000: So we can show the low level logic of creating and destroying CUDA streams

ds58: I think if you do that, you shouldn't use CUDAProgram or CUDAKernel classes. The typical usage is to use `CUDAStreamManager`

PotatoPeeler3000: The CUDAStreamManager hides some of the stuff that needs to happen with the stream. In this example we wanted to expose that to the user.

ihmc-perception/src/test/java/us/ihmc/perception/cuda/ExampleCUDAKernel.java
PotatoPeeler3000: True, it should live in the javadoc for whatever method is getting called, I can't remember if we put it there but will take a look

ihmc-perception/src/test/java/us/ihmc/perception/cuda/ExampleCUDAKernel.java
PotatoPeeler3000: Dam formatter lol, let me get that guy

ihmc-perception/src/test/java/us/ihmc/perception/cuda/ExampleCUDAKernel.java
PotatoPeeler3000: Actually looked at this and because we don't use the StreamManager, gonna leave the comment, helpful to the user

ihmc-perception/src/test/java/us/ihmc/perception/cuda/ExampleCUDAKernel.java
PotatoPeeler3000: Mission acchomplished

ihmc-perception/src/test/java/us/ihmc/perception/cuda/ExampleCUDAKernel.java
ds58: Can you explain? I think you should use `CUDAStreamManager` here because that's how a user-developer would use it elsewhere.

PotatoPeeler3000: The CUDAStreamManager hides the fact that you need to close the stream, and it doesn't allow me to add comments explaining what the stream does. We only have the one stream in this example so I didn't need the manager. In the next example we are using the manager

ds58: Okay, fair enough

ds58 (approved):

PR #480, opened 2024-11-18
[Feature] Small RDXSensorSimulator Optimization

TomaszTB:
After learning a bit more about OpenGL, I realized we weren't actually getting anything useful from rendering depth and color separately. 
This PR changes the RDXSensorSimulator such that it only renders once, for both depth and color.

### DIFF
--- a/ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/simulation/sensors/RDXSensorSimulator.java
+++ b/ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/simulation/sensors/RDXSensorSimulator.java
@@ -54,12 +54,11 @@ public class RDXSensorSimulator
 
    // Images
    private Mat colorImage = null;
-   private Instant colorGrabTime = Instant.EPOCH;
 
    private Mat depthImage = null;
    private Mat depthData = null;
-   private Instant depthGrabTime = Instant.EPOCH;
 
+   private Instant grabTime = Instant.EPOCH;
    private long sequenceNumber = 0L;
 
    public RDXSensorSimulator(int imageWidth, int imageHeight, float verticalFOV, float minRange, float maxRange)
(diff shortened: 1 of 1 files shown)


ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/simulation/sensors/RDXSensorSimulator.java
```diff
@@ -173,19 +175,9 @@ public void render(RigidBodyTransform cameraTransformToWorld)
...
       }
 
-      if (depthImage != null) // If depth is enabled
+      if (depthImage != null)
       {
-         // Render and "grab" the color image
-         simulationScene.preRenderDepth(camera);
-         GL41.glViewport(0, 0, imageWidth, imageHeight);
-         simulationScene.renderDepth(RDXSceneLevel.GROUND_TRUTH.SINGLETON_SET);
-         simulationScene.postRenderDepth();
```
TomaszTB, line 182: The color render generates the depth data we are using. We don't have to render again for depth.

calvertdw: The point here is that it only renders for the ground truth scene level, so excluding virtual and modelled renderables. Virtual -> gizmos, etc.   Modelled -> point cloud, height map, etc.

TomaszTB: I think the color render also only renders for the ground truth scene level. Perhaps I have to double check that, but I'd be surprised if gizmos are visible in our simulated sensors.

TomaszTB: I took another look at the code and can confirm that the color render also only renders ground truth.

calvertdw (approved):

PR #490, opened 2024-11-23
Remove about 200 MB of LFS files and excessive resolution in models.

calvertdw:


ihmc-perception/src/main/resources/us/ihmc/perception/ffmpeg/example.webm
calvertdw, line 1: There's another copy of this in high level behaviors that actually is used in a demo.

ds58 (approved):

PR #492, opened 2024-11-25
Add new ContinuousHikingAPI and ContinuousHikingCommandMessage

PotatoPeeler3000:
Updated the ContinuousWalkingAPI and message which are now deprecated. Created the ContinuousHikingAPI and ContinuousHikingCommandMessage.msg which will be what the process use's going forard. This contains the same stuff as the old message. Just a new name going forward

### DIFF
--- a/ihmc-interfaces/src/main/messages/ihmc_interfaces/behavior_msgs/msg/ContinuousHikingCommandMessage.msg
+++ b/ihmc-interfaces/src/main/messages/ihmc_interfaces/behavior_msgs/msg/ContinuousHikingCommandMessage.msg
@@ -1,0 +1,42 @@
+# This message sends the current request for the Continuous Hiking process.
+# This holds parameters on how the Continuous Hiking process will behave.
+
+# Flag to start Continuous Hiking
+bool enable_continuous_hiking
+
+# Flag that allows the user to set a number of steps that will be taken, and Continuous Hiking will stop after that. This way the robot doesn't walk forever
+uint32 steps_before_safety_stop
+
+# Flag to walk straight forward with Continuous Hiking. This is generally a keyboard input, and is different then trying to walk with the Joystick
+bool walk_forwards
+
+# Flag to square up to the goal when finishing walking. We may want the robot to end at an exact location.
+# Setting this to true tells the robot to step on the final goal positions.
+bool square_up_to_goal
+
+# Flag to enable/disable planning with A* planner. This is the default planner to use
+bool use_astar_footstep_planner
+
+# Flag to enable/disable planning with monte-carlo footstep planner
+bool use_monte_carlo_footstep_planner
+
+# Flag to enable/disable using previous plan as reference. Generally this should be set to true to reduce planning time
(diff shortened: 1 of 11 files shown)


PotatoPeeler3000:
Wanted to change the name to ContinuousHiking, but didn't want to remove the old one yet. This way people don't have to update the command until they are ready too.

ihmc-interfaces/src/main/messages/ihmc_interfaces/behavior_msgs/msg/ContinuousHikingCommandMessage.msg
```diff
@@ -0,0 +1,42 @@
...
+bool use_monte_carlo_plan_as_reference
+
+# Walk with the Joystick Controller
+bool use_joystick_controller
+
+# Forward joystick value
+float64 forward_value
+
+# Backward walking from Joystick Controller
+bool walk_backwards
```
calvertdw, line 36: Why wouldn't backwards just be a negative forward_value?

PotatoPeeler3000: Don't know, I don't currently use the float forward_value. But I do use the boolean for walking backwards.

ihmc-interfaces/src/main/messages/ihmc_interfaces/behavior_msgs/msg/ContinuousHikingCommandMessage.msg
```diff
@@ -0,0 +1,42 @@
...
+
+# Flag to square up to the goal when finishing walking. We may want the robot to end at an exact location.
+# Setting this to true tells the robot to step on the final goal positions.
+bool square_up_to_goal
+
+# Flag to enable/disable planning with A* planner. This is the default planner to use
+bool use_astar_footstep_planner
+
+# Flag to enable/disable planning with monte-carlo footstep planner
+bool use_monte_carlo_footstep_planner
```
ds58, line 21: Should the planner selection be a string or a short or something? Like, it wouldn't make sense to enable both astar and monte carlo.

PotatoPeeler3000: Maybe in the future that might make sense. I haven't put much thought into it. On one of my branches you can do this "hybrid" plan where if both of those are true. Then you do the Monte Carlo Planner and you use that plan as a reference to the A* Planner. That could probably be handled with a string or short,  I'm not trying to change that implementation to much at the moment. Much lower on the priority list then some other tasks.

PotatoPeeler3000: So for now I'd like to leave this, and if it causes problems going forward I can change it then.

ds58 (approved):

PR #496, opened 2024-11-26
Added a subscriber to the continuous hiking process (CHP) that will clear the waypoint list.

PotatoPeeler3000:


ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousHikingStateMachine/ReadyToPlanState.java
```diff
@@ -263,4 +265,15 @@ public void addWayPointPoseToList(PoseListMessage poseListMessage)
       walkToGoalWayPointPoses.add(latestWayPoint);
       debugger.publishStartAndGoalForVisualization(continuousPlanner.getStartStancePose(), latestWayPoint);
    }
+
+   /**
+    * This allows the {@link ReadyToPlanState#walkToGoalWayPointPoses} to be cleared.
+    * This empties the list so the user can place a fresh goal that Continuous Hiking will use.
+    */
+   public void clearWayPointList(Empty emptyMessage)
```
PotatoPeeler3000, line 273: This subscribes to a message that can be sent to the user. The JavaDoc should explain what this does. The reason for adding this now is that if the user mis-clicks there was nothing to clear that goal and reset. This caused problems when trying to use Continuous Hiking from Python.

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ContinuousHikingStateMachine/ReadyToPlanState.java
```diff
@@ -263,4 +265,15 @@ public void addWayPointPoseToList(PoseListMessage poseListMessage)
...
+   {
+      LogTools.info("Clearing waypoint list for WALK_TO_GOAL");
+      walkToGoalWayPointPoses.clear();
+      continuousHikingParameters.setEnableContinuousHiking(false);
```
PotatoPeeler3000, line 277: Also if we are going to clear all the goals, we probably don't want to be walking, cause we have nothing to walk towards. So make sure to stop everything.

LuigiPenco93 (approved):

rjgriffin42:
This is good. However, this should be squashed into a single commit, t here's only four lines of code that were changed. Fine if you do it by "squash and merge"

rjgriffin42 (approved):

PotatoPeeler3000:
> This is good. However, this should be squashed into a single commit, t here's only four lines of code that were changed. Fine if you do it by "squash and merge"

Agreed, I haven't been considered the commit history too much but can start to consider it more.

PR #500, opened 2024-11-30
Update nvidia docker container to have perception dependencies installed.

calvertdw:


ds58:
No one uses this so I am not sure we should keep it

calvertdw:
> No one uses this so I am not sure we should keep it

This enabled me get ZED SDK installed on Arch by being able to copy out the resulting `/usr/local/zed` folder. Otherwise, I can't run/test/develop perception.

ds58 (approved):
Probably need to bump the base image version in the other containers e.g. `docker/nvidia-ros2-ide` or delete these if we don't intend to maintain them.

calvertdw:
> Probably need to bump the base image version in the other containers e.g. `docker/nvidia-ros2-ide` or delete these if we don't intend to maintain them.

I kind of don't want to spend the time right now on that since it's not towards anything, but I think those are useful reference and worth keeping around.

PR #513, opened 2024-12-04
Add optional JPEG compression in SRT video relay

ds58:


ihmc-perception/src/main/java/us/ihmc/perception/streaming/ROS2SRTVideoStreamImageMessageRelayWorker.java
```diff
@@ -1,26 +1,38 @@
...
+import javax.annotation.Nullable;
 import java.time.Instant;
 
 public class ROS2SRTVideoStreamImageMessageRelayWorker
 {
+   /**
+    * Set to true to enable compression to JPEG (using nvJPEG) before publishing the ImageMessage.
+    * Introduces GPU overhead and more latency, but useful for recording rosbags.
+    */
+   private static final boolean COMPRESS_USING_NVJPEG = true;
```
TomaszTB: Lets make it false by default

calvertdw: It would be better if this was a JVM parameter or literal parameter in the constructor. Making options that require local code changes sucks, especially on the OCU, there's a bunch of garbage left afterwards.

ds58: Yep that was a mistake

TomaszTB: Yeah, having random changes on the OCU sucks. I always struggle setting the JVM parameters correctly though. I'd be fine with keeping it a static field since we would hardly ever change it. @ds58 What do you think about this?

ds58: I do not see the benefit of using environment variables from the way we run the applications always from IDE

calvertdw (approved):

TomaszTB (approved):
Yeah, this is cool

PR #519, opened 2024-12-04
Get a better hold on action success and failure events.

calvertdw:


ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/behaviorTree/BehaviorTreeRootNodeExecutor.java
```diff
@@ -94,7 +95,22 @@ else if (state.getActionChildren().get(i).calculateExecuteAfterActionIndex() < e
...
+         {
+            String name = currentlyExecutingAction.getDefinition().getName();
+            double elapsedExecutionTime = currentlyExecutingAction.getState().getElapsedExecutionTime();
+            if (failed)
+            {
+               LogTools.error("Action failed after %.2f s: %s".formatted(elapsedExecutionTime, name));
+            }
+            else
+            {
+               LogTools.info("Action completed successfully in %.2f s: %s".formatted(elapsedExecutionTime, name));
```
calvertdw, line 111: There was no one place before where the event of success and failure were captured. This is important to begin implementing fallback behavior.

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/sequence/TrajectoryDurationTracker.java
```diff
@@ -40,13 +47,28 @@ public boolean getTimeIsUp()
       return timeIsUp;
    }
 
-   public boolean getHitTimeLimit()
+   public boolean getHitTimeLimit(BehaviorTreeNodeMessageLogger logger)
    {
+      if (hitTimeLimit)
+      {
+         logger.error(1, "Task execution timed out. %.2f > %.2f Nominal: %.2f".formatted(executionTimer.getElapsedTime(), timeout, nominalExecutionDuration));
```
calvertdw, line 54: I put this error message in this call so the operator gets more info about the failure, without having to adjust the print for the 7 usages.

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/behaviorTree/BehaviorTreeRootNodeExecutor.java
```diff
@@ -154,8 +170,6 @@ private void executeNextAction()
       state.getLogger().info("Triggering action execution: %s".formatted(actionToExecute.getDefinition().getName()));
       actionToExecute.update();
       actionToExecute.triggerActionExecution();
-      actionToExecute.updateCurrentlyExecuting();
-      currentlyExecutingActions.add(actionToExecute);
```
calvertdw, line 158: Removing these two lines is the most significant change in the PR. Since updateCurrentlyExecuting can set executing to false, there was a possibility of inconsistent and undefined state. Now it's only called in one place, making catching success and failure events guaranteed.

TomaszTB (approved):

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/behaviorTree/BehaviorTreeRootNodeExecutor.java
LuigiPenco93: So this 2 lines were not supposed to be executed, because the this.update() already does it?

calvertdw: I think my thinking was to call these here to keep state consistent, but that ruins other state guarantees. It's better to call it on the next update so it's only done in one place.

LuigiPenco93 (approved):

PR #527, opened 2024-12-10
Feature/quickster footstep provider

stefanfasano:
This is the initial groundwork for the implementation of Quickster's footstep calculator into the regular walking controller via the CSG (in step generator thread).

Tests included in this PR:
- QFPTest; this tests switching between CSG modes while walking in place and standing in place

The next steps include:

- New message type to ensure things on the controller side know which CSG mode is in use
- Implementation into whole body MPC scheme
- Hardware testing on robot
- Max speed limits along with max acceleration per step


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/highLevelStates/walkingController/states/WalkingSingleSupportState.java
```diff
@@ -311,28 +332,67 @@ public void onEntry()
...
 
+      //
+      controllerToolbox.updateBipedSupportPolygons();
+
       balanceManager.setSwingFootTrajectory(swingSide, feetManager.getSwingTrajectory(swingSide));
+      balanceManager.adjustFootstepInCoPPlan(nextFootstep);
+      balanceManager.computeICPPlan();
+   }
+
+   private boolean haveWeEntered = false;
```
christiandebuys, line 369: I am confused by this boolean. Isn't whether or not you are executing the method inside `onEntry()` enough information to set `firstTick `in the method? Why are `haveWeEntered `and `firstTick `both in `handleNewFootstep()`?

stefanfasano: There's probably a better way to do this whole thing. They don't exactly convey the same information, since first tick could be false if it is the second+ tick or if it has not entered the state yet, and that distinction is important. It is possible for the method that relies on this boolean (`handleNewFootstep()`) to be called without entering the state (`WalkingSingleSupportState`) because I have this method tied to a listener in another class (`WalkingMessageHandler`), and the class that this method belongs to is instantiated twice, once for each leg. So it is possible for this to be called by the listener even if the class (and therefore single support of the proper leg) has not been called.

The thing is, this is probably redundant because I later ended up adding a check to make sure that we only proceed with the `handleNewFootstep()` method if we are on the proper side of the robot. So I can probably delete the logic that relies on `haveWeEntered`

Beomyeong (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/highLevelStates/walkingController/states/WalkingSingleSupportState.java
```diff
@@ -311,28 +332,67 @@ public void onEntry()
...
+         feetManager.adjustSwingTrajectory(swingSide, nextFootstep, swingTime);
+
+      //
       if (feetManager.adjustHeightIfNeeded(nextFootstep))
       {
          walkingMessageHandler.updateVisualizationAfterFootstepAdjustement(nextFootstep);
          feetManager.adjustSwingTrajectory(swingSide, nextFootstep, swingTime);
       }
 
+      //
```
james-p-foster, line 361: Why all the `//` in this file?

james-p-foster (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/highLevelStates/walkingController/states/WalkingSingleSupportState.java
stefanfasano: ah some are so I can more easily identify where I made changes, and some are for comments that I meant to write but never did. I'll fix both of those

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/controller/ICPController.java
```diff
@@ -132,13 +136,15 @@ public class ICPController implements ICPControllerInterface
    private final ICPControllerHelper helper = new ICPControllerHelper();
 
    public ICPController(WalkingControllerParameters walkingControllerParameters,
+                        WalkingMessageHandler walkingMessageHandler,
```
rjgriffin42: I am not a fan of this change. The walking message handler is meant to be a high level thing, and the ICPController is a low level thing. I would much rather add a method to the ICPController that allows you to "override" the `useCoPFeedback` boolean, which is what you're doing, then have that set by a listener that you add, so that this is occuring at a high level, rather than embedded into the ICP Controller. It's a little more convoluted to write, but much more obvious than this object-oriented approach when it comes to tracing what's going on.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/controller/ICPControllerParameters.java
```diff
@@ -96,6 +96,14 @@ public boolean useCMPFeedback()
       return true;
    }
 
+   /**
+    * Enabling this boolean allows feedback control of CoP (feedback alpha < 1).
+    */
+   public boolean useCoPFeedback()
```
rjgriffin42, line 102: so if this is false, but useCMPFeedback is true, it will still move the base of the pendulum around, but will do so by moving the CMP. Is that intentional?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/LinearMomentumRateControlModule.java
```diff
@@ -171,7 +171,7 @@ public LinearMomentumRateControlModule(HighLevelHumanoidControllerToolbox contro
            controllerToolbox.getYoGraphicsListRegistry());
    }
 
-   public LinearMomentumRateControlModule(CenterOfMassStateProvider centerOfMassStateProvider,
+   public LinearMomentumRateControlModule(HighLevelHumanoidControllerToolbox controllerToolbox,
```
rjgriffin42: The LinearMomentumRateControlModule was written in such a way that it could be removed from the control thread. By adding the controller toolbox here, that gets messed up. I think the right thing to do would be to add the method to ICPController, like I said, then add that override cop feedback command as a field to the momentum rate input, such that the field is getting set in the motion control stuff based on some listener that's attached to the walking message handler, and then sent to this control module and used. Then you don't have to add the control toolbox or walking message handler to this class.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/BalanceManager.java
```diff
@@ -1026,6 +1027,13 @@ public void computeNormalizedEllipticICPError(RobotSide transferToSide)
...
                                                                      : maxICPErrorBeforeSingleSupportOuterY.getValue();
+
+      if (controllerToolbox.getWalkingMessageHandler().getUsingQFP().getBooleanValue())
+      {
+         maxICPErrorBeforeSingleSupportX += 0.3;
+         maxICPErrorBeforeSingleSupportY += 0.3;
+      }
+
       normalizedICPError.set(MathTools.square(icpError2d.getX() / maxICPErrorBeforeSingleSupportX)
                              + MathTools.square(icpError2d.getY() / maxICPErrorBeforeSingleSupportY));
```
rjgriffin42: It would likely be cleaner to instead just set the normalizedICPError to zero in that if statement.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/scs2/SCS2AvatarSimulationFactory.java
```diff
@@ -460,7 +460,7 @@ private void setupStepGeneratorThread()
       else
       {
          JoystickBasedSteppingPluginFactory joystickPluginFactory = new JoystickBasedSteppingPluginFactory();
-         if (heightMapForFootstepZ.hasValue())
+         if (heightMapForFootstepZ.hasValue() && heightMapForFootstepZ.get() != null)
```
rjgriffin42, line 463: I don't think this captures how the height map will work on hardware. THe height map should be continuously streamed to the robot. Right now, the step generator will only take a single instance of the height map at construction. This is out of scope of this PR, but I'm pointing out that this won't work as intended on the robot.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/stepAdjustment/ErrorBasedStepAdjustmentController.java
```diff
@@ -124,10 +125,13 @@ public class ErrorBasedStepAdjustmentController implements StepAdjustmentControl
 
    private final BipedSupportPolygons bipedSupportPolygons;
 
+   private final WalkingMessageHandler walkingMessageHandler;
+
    private final FramePoint3D vertexInWorld = new FramePoint3D();
    private final FrameConvexPolygon2D allowableAreaForCoPInFoot = new FrameConvexPolygon2D();
 
    public ErrorBasedStepAdjustmentController(WalkingControllerParameters walkingControllerParameters,
+                                             WalkingMessageHandler walkingMessageHandler,
```
rjgriffin42: This should also be passed in functionally, and not done through the walking message handler. In reality, in the state where you're calling the step adjustment, you should just be able to write an if block there, rather than turning it off inside the step adjustment controller.

Also I don't believe this class is actually used.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/desiredFootStep/footstepGenerator/ContinuousStepGenerator.java
```diff
@@ -305,32 +336,59 @@ else if (statusToProcess == FootstepStatus.COMPLETED)
...
+      }
+
       for (int i = startingIndexToAdjust; i < parameters.getNumberOfFootstepsToPlan(); i++)
       {
-         double xDisplacement = MathTools.clamp(stepTime.getValue() * desiredVelocityX, maxStepLength);
-         double yDisplacement = stepTime.getValue() * desiredVelocityY + swingSide.negateIfRightSide(defaultStepWidth);
-         double headingDisplacement = stepTime.getValue() * turningVelocity;
 
-         if (swingSide == RobotSide.LEFT)
+         if (currentCSGMode.getEnumValue() == ContinuousStepGeneratorMode.QFP && quicksterFootstepProvider.hasValue())
```
rjgriffin42, line 351: Can you add a FIXME note to delete this once you've done the rollout stuff inside the QFP?

ihmc-interfaces/src/main/messages/ihmc_interfaces/controller_msgs/msg/FootstepDataMessage.msg
```diff
@@ -9,6 +9,9 @@ byte TRAJECTORY_TYPE_OBSTACLE_CLEARANCE = 1
 byte TRAJECTORY_TYPE_CUSTOM = 2
 byte TRAJECTORY_TYPE_WAYPOINTS = 3
 
+byte CSG_MODE_STANDARD = 0
+byte CSG_MODE_QFP = 1
```
rjgriffin42, line 13: Would it make more sense for all this stuff to go into the data list message, rather than the message of each individual footstep? It seems like this is either true for the entire list of steps, or its true for none of them.

ihmc-interfaces/src/main/messages/ihmc_interfaces/controller_msgs/msg/FootstepDataMessage.msg
```diff
@@ -79,4 +84,7 @@ float64 liftoff_duration -1.0
 StepConstraintsListMessage step_constraints
 
 # If the controller should check and fail if the step is not reachable
-bool should_check_for_reachability false
\ No newline at end of file
+bool should_check_for_reachability false
+
+# If the desired footstep is for walking in place with zero forward or lateral velocity
```
rjgriffin42: I think it would be good to make a note that this is only used if the csg_mode is QFP

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/math/trajectories/C1ContinuousTrajectorySmoother.java
```diff
@@ -57,8 +58,11 @@ public C1ContinuousTrajectorySmoother(String namePrefix, FixedFramePositionTraje
 
       YoRegistry registry = new YoRegistry(namePrefix + getClass().getSimpleName());
 
-      trackingStiffness = new DoubleParameter(namePrefix + "TrackingStiffness", registry, 200.0);
-      trackingZeta = new DoubleParameter(namePrefix + "TrackingZeta", registry, 0.8);
+      trackingStiffness = new YoDouble(namePrefix + "TrackingStiffness", registry);
+      trackingStiffness.set(defaultTrackingStiffness);
```
rjgriffin42: I would assume the answer is yes, but I want to make sure that you checked all the parameter files to make sure this didn't change anything?

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/math/trajectories/C1ContinuousTrajectorySmoother.java
```diff
@@ -40,14 +39,16 @@ public class C1ContinuousTrajectorySmoother implements FixedFramePositionTraject
...
+   private final YoDouble trackingStiffness;
+   private final YoDouble trackingZeta;
 
    private final MatrixExponentialCalculator matrixExponentialCalculator = new MatrixExponentialCalculator(2);
    private final DMatrixRMaj closedLoopStateMatrix = new DMatrixRMaj(2, 2);
    private final DMatrixRMaj scaledClosedLoopStateMatrix = new DMatrixRMaj(2, 2);
    private final DMatrixRMaj stateTransitionMatrix = new DMatrixRMaj(2, 2);
 
+   private final double defaultTrackingStiffness = 200.0;
+   private final double defaultTrackingZeta = 0.8;
```
rjgriffin42: can these be static?

ihmc-humanoid-robotics/src/main/java/us/ihmc/humanoidRobotics/communication/controllerAPI/command/FootstepDataCommand.java
```diff
@@ -28,6 +29,7 @@ public class FootstepDataCommand implements Command<FootstepDataCommand, Footste
    private long sequenceId;
    private RobotSide robotSide;
    private TrajectoryType trajectoryType = TrajectoryType.DEFAULT;
+   private ContinuousStepGeneratorMode csgMode = ContinuousStepGeneratorMode.STANDARD;
```
rjgriffin42, line 32: if they get moved in the message, they should be moved here, too.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/BalanceManager.java
stefanfasano: will do

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/scs2/SCS2AvatarSimulationFactory.java
stefanfasano: sounds good, I'm guessing this isn't specific to the change I made, but an existing issue with this particular implementation right?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/controller/ICPController.java
stefanfasano: will do

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/controller/ICPControllerParameters.java
stefanfasano: No, should I be turning that false as well?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/LinearMomentumRateControlModule.java
stefanfasano: Good call, will do

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/stepAdjustment/ErrorBasedStepAdjustmentController.java
stefanfasano: ok, so to clarify, basically do what you suggested for ICPController right?

ihmc-interfaces/src/main/messages/ihmc_interfaces/controller_msgs/msg/FootstepDataMessage.msg
stefanfasano: With the current implementation, the FootstepDataList is a mix of QFP and standard (first one QFP, the rest are standard). I know we want to make the change such that multiple/all steps in the plan are computed with QFP, but I still think there could be instances perhaps when we switch between the two footstep providers, where the list will become mixed again. I'd rather leave this here because I feel it provides a more granular level of control over the timing of the parameter adjustments

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/math/trajectories/C1ContinuousTrajectorySmoother.java
stefanfasano: good call, are you referring to the xml param files? Which one(s) do you reckon these parameters would live in?

ihmc-humanoid-robotics/src/main/java/us/ihmc/humanoidRobotics/communication/controllerAPI/command/FootstepDataCommand.java
stefanfasano: what do you mean?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/BalanceManager.java
stefanfasano: done

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/LinearMomentumRateControlModule.java
stefanfasano: done

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/controller/ICPController.java
stefanfasano: done

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/stepAdjustment/ErrorBasedStepAdjustmentController.java
stefanfasano: done

ihmc-interfaces/src/main/messages/ihmc_interfaces/controller_msgs/msg/FootstepDataMessage.msg
stefanfasano: done

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/desiredFootStep/footstepGenerator/ContinuousStepGenerator.java
stefanfasano: done

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/math/trajectories/C1ContinuousTrajectorySmoother.java
stefanfasano: done

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/math/trajectories/C1ContinuousTrajectorySmoother.java
stefanfasano: done

PR #529, opened 2024-12-11
Feature/external controller

rjgriffin42:


external-control/src/main/cpp/external-control.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+namespace ihmc
+{
+    ExternalControlImpl::ExternalControlImpl(const double default_stiffness, const double default_damping, const int number_of_joints)
+    {
+        constant_position_controller_ = ConstantPositionController(default_stiffness, default_damping, number_of_joints);
+        desired_state_data_ = Eigen::VectorXd(13 + 2 * number_of_joints);
+        desired_control_data_ = Eigen::VectorXd(number_of_joints);
+        p_gains_ = Eigen::VectorXd(number_of_joints);
+        d_gains_ = Eigen::VectorXd(number_of_joints);
```
Beomyeong: This way of defining size may cause memory allocation issues. 
resizing the vector could be in this way
desired_state_data_.resize(13 + 2* number_of_joints);
desired_control_data_.resize(number_of_joints);
p_gains.resize(number_of_joints);
d_gains_.resize(number_of_joints);

And .resize() does not set the variable to be zero. So, using desired_state_data_.setZero() would be safe.

external-control/src/main/cpp/controllers/constant-position-controller.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+namespace ihmc
+{
+    ConstantPositionController::ConstantPositionController(const double default_stiffness, const double default_damping, const int number_of_joints)
+    {
+        number_of_joints_ = number_of_joints;
+        desired_joint_velocities_ = Eigen::VectorXd(number_of_joints_);
+        desired_joint_torques_ = Eigen::VectorXd(number_of_joints_);
+        desired_joint_stiffnesses_ = Eigen::VectorXd(number_of_joints_);
+        desired_joint_damping_ = Eigen::VectorXd(number_of_joints_);
```
Beomyeong: use .resize(Number_of_size); instead using Eigen::VectorXd(#). This may cause the memory issue.

PR #532, opened 2024-12-13
Implement ROS 2 peer clock offset estimator

calvertdw:


ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
```diff
@@ -0,0 +1,143 @@
...
+         MessageTools.fromMessage(receivedMessage.getReplyTarget(), receivedReplyTarget);
+
+         if (receivedMessage.getIsRequest() && receivedRequestTarget.equals(ourGuid)) // Reply
+         {
+            PeerClockOffsetEstimatorPingMessage replyMessage = new PeerClockOffsetEstimatorPingMessage();
+            replyMessage.set(receivedMessage);
+            replyMessage.setIsRequest(false);
+            MessageTools.toMessage(Instant.now(), replyMessage.getReplySendTime());
+
+            ThreadTools.startAsDaemon(() ->
```
TomaszTB: Perhaps an `ExecutorService` would be better than starting a new thread each time?

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
```diff
@@ -0,0 +1,143 @@
...
+         synchronized (publisher)
+         {
+            publisher.publish(requestMessage);
+         }
+
+         ++nextPeerToPing;
+      }
+   }
+
+   public void destroy()
```
TomaszTB, line 141: Should this also call `remove()` on the subscriber and publisher? I never really understood if that's necessary or not

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * Sends out pings to other instances of this class on the network to estimate
+ * the clock offsets on those computers. This can be used to synchronize data
+ * modifications. We account for and assume a symmetric network delay.
+ */
+public class ROS2PeerClockOffsetEstimator
```
TomaszTB, line 29: What about making this class a singleton? I can't imagine why you'd want more than one instance per process.

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
```diff
@@ -0,0 +1,143 @@
...
+   private final PeerClockOffsetEstimatorPingMessage requestMessage = new PeerClockOffsetEstimatorPingMessage();
+   private final PeerClockOffsetEstimatorPingMessage receivedMessage = new PeerClockOffsetEstimatorPingMessage();
+   private final SampleInfo sampleInfo = new SampleInfo();
+   private final Guid receivedRequestTarget = new Guid();
+   private final Guid receivedReplyTarget = new Guid();
+
+   public ROS2PeerClockOffsetEstimator(ROS2Node ros2Node)
+   {
+      publisher = ros2Node.createPublisher(TOPIC);
+      ourGuid = publisher.getPublisher().getGuid();
```
TomaszTB, line 54: There's a build issue. The `getPublisher()` method of `ROS2Publisher` is package private, so it can't be called here :(

TomaszTB (changes requested):
Looking pretty good (except for build issue). I left a few questions/suggestions

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
calvertdw: If you look, under the hood, that's all an ExecutorService does

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
calvertdw: I can do that. When you destroy the parent node, it removes all the pubs/subs automatically, but it doesn't hurt to remove them early.

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
calvertdw: Yes it depends on a new release of ros2-library

ds58: We shouldn't merge until that release

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
TomaszTB: I was specifically thinking of this: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executors.html#newCachedThreadPool--
It sounds like threads are reused according to the JavaDoc

ihmc-communication/build.gradle.kts
```diff
@@ -24,5 +24,6 @@ mainDependencies {
 
 testDependencies {
    api("us.ihmc:ihmc-robotics-toolkit-test:source")
+   api("us.ihmc:ihmc-graphics-libgdx:source")
```
ds58, line 27: I would rather this not get added

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
ds58: nah I don't think it should be

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
ds58: I agree with @TomaszTB here

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
```diff
@@ -0,0 +1,143 @@
...
+               }
+            }
+         }
+      });
+
+      requestThread.setFrequencyLimit(5.0);
+      requestThread.startRepeating();
+   }
+
+   private void requestThread()
```
ds58: Can this be called requestUpdate or something? It doesn't implicitly have anything to do with a thread by itself

calvertdw: changed to runRequestTask

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
calvertdw: Okay, done!

ihmc-communication/src/main/java/us/ihmc/communication/ros2/sync/ROS2PeerClockOffsetEstimator.java
calvertdw: That would go for a lot of things, I'd rather not go there on all of them

ds58 (approved):

TomaszTB (approved):

PR #535, opened 2024-12-16
Add OpenCLManager back to the RapidHeightMapManager

PotatoPeeler3000:
This way we can support both the OpenCL and the CUDA height maps in simulation

### DIFF
--- a/ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
+++ b/ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
@@ -18,6 +18,7 @@ import us.ihmc.perception.camera.CameraIntrinsics;
 import us.ihmc.perception.gpuHeightMap.RapidHeightMapExtractor;
 import us.ihmc.perception.gpuHeightMap.RapidHeightMapExtractorCUDA;
 import us.ihmc.perception.heightMap.TerrainMapData;
+import us.ihmc.perception.opencl.OpenCLManager;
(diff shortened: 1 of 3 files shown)


ds58 (approved):

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
```diff
@@ -108,8 +110,8 @@ public void update(Mat latestDepthImage,
 
    public HeightMapData getLatestHeightMapData()
    {
-      HeightMapData temp = new HeightMapData((float) RapidHeightMapExtractorCUDA.getHeightMapParameters().getGlobalCellSizeInMeters(),
-                                             (float) RapidHeightMapExtractorCUDA.getHeightMapParameters().getGlobalWidthInMeters(),
+      HeightMapData temp = new HeightMapData((float) RapidHeightMapExtractor.getHeightMapParameters().getGlobalCellSizeInMeters(),
+                                             (float) RapidHeightMapExtractor.getHeightMapParameters().getGlobalWidthInMeters(),
```
TomaszTB: Doesn't this change break the whole thing right now? I suppose it's alright since it's still WIP

TomaszTB (approved):

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
PotatoPeeler3000: Hmm, that change isn't supposed to be there lol, maybe I'm a bit rusty

PR #545, opened 2024-12-17
Support OpenCL and CUDA within the RapidHeightMapManager

PotatoPeeler3000:
Because the CUDA height map has a couple bugs in it still, we should support the OpenCL height map in simulation.
So I've added a check that if we can't run CUDA, then don't.
The reasons we aren't able to run CUDA is because:
- The user doesn't have an NVidia GPU
- Doesn't have CUDA installed
- Doesn't have the right version of CUDA installed


ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
```diff
@@ -9,13 +9,19 @@
...
+import java.io.InputStreamReader;
+
 import static org.bytedeco.cuda.global.cudart.*;
 import static org.bytedeco.cuda.global.nvjpeg.NVJPEG_STATUS_SUCCESS;
 import static org.bytedeco.cuda.global.nvrtc.NVRTC_SUCCESS;
 import static org.bytedeco.cuda.global.nvrtc.nvrtcGetErrorString;
 
 public class CUDATools
 {
+   public static final String REQUIRED_CUDA_VERSION = "12.6";
```
TomaszTB: I'm using 12.7 and everything works (although my CUDA Toolkit version is 12.6). I think robot's backpack is using 12.4 or something older and everything works. We shouldn't check for the CUDA version in my opinion

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
```diff
@@ -59,6 +65,53 @@ public static boolean hasCUDADevice()
       return getCUDADeviceCount() > 0;
    }
 
+   public static boolean doesCUDAExistAndMatchVersion()
```
TomaszTB: I'm pretty certain the `hasCUDA()` method is good enough. The `cudart` library wouldn't be present if CUDA is not available on a device (I've tested this on the old Lenovo ThinkPads). 
Also, if you really want to use `nvcc` then you might want to use its absolute path. I don't have `nvcc` in my `PATH`, so I this might not work for me.

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
```diff
@@ -48,11 +54,19 @@ public RapidHeightMapManager(OpenCLManager openCLManager,
...
+         rapidHeightMapExtractor = new RapidHeightMapExtractorCUDA(leftFootSoleFrame, rightFootSoleFrame, deviceDepthImage, 1);
+      }
+      else
+      {
+         heightMapBytedecoImage = new BytedecoImage(depthImageIntrinsics.getWidth(), depthImageIntrinsics.getHeight(), opencv_core.CV_16UC1);
+         heightMapBytedecoImage.createOpenCLImage(openCLManager, OpenCL.CL_MEM_READ_WRITE);
+         rapidHeightMapExtractor = new RapidHeightMapExtractor(openCLManager, leftFootSoleFrame, rightFootSoleFrame, heightMapBytedecoImage, 1);
+      }
 
-      deviceDepthImage = new GpuMat(depthImageIntrinsics.getWidth(), depthImageIntrinsics.getHeight(), opencv_core.CV_16UC1);
```
TomaszTB, line 54: Oh crap, this was incorrect. The height should come first, then width. This might fix some bugs Nehar was having???

TomaszTB (commented):
Just left a few thoughts/suggestions. 
Personally, I think it'd be good to support both CUDA and OpenCL, at least for now. 
I think it would become problematic in the long run though. The CUDA and OpenCL code seems to be diverging a bit, and I imagine it will only diverge more over time. It will become difficult to support both. We should really decide what we want to do long term.

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
PotatoPeeler3000: Ok interesting, lets deal with this in a future PR, not sure what would happen if we change it

TomaszTB (approved):
You've got build issues in `MonteCarloFootstepPlanningTest`. But otherwise seems good.

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
```diff
@@ -48,11 +54,19 @@ public RapidHeightMapManager(OpenCLManager openCLManager,
                                 CameraIntrinsics depthImageIntrinsics,
                                 ROS2PublishSubscribeAPI ros2)
    {
-      rapidHeightMapExtractor = new RapidHeightMapExtractorCUDA(leftFootSoleFrame, rightFootSoleFrame);
-      rapidHeightMapExtractor.setDepthIntrinsics(depthImageIntrinsics);
+      if (hasCUDAAvailable)
+      {
+         deviceDepthImage = new GpuMat(depthImageIntrinsics.getWidth(), depthImageIntrinsics.getHeight(), opencv_core.CV_16UC1);
+         rapidHeightMapExtractor = new RapidHeightMapExtractorCUDA(leftFootSoleFrame, rightFootSoleFrame, deviceDepthImage, 1);
```
ds58, line 60: If the CUDA height map has bugs like you said, why are we using it in RapidHeightMapManager at the moment? It's not clear what this PR is trying to do. I would expect that the CUDA implementation only be in some demo/development classes until it's ready to be swapped out.

PotatoPeeler3000: This PR supports using CUDA, when its available. And if its not available, we use the OpenCL height map.

ds58: I don't think the RapidHeightMapManager should support both. I think it should support the known working implementation (OpenCL), and when the CUDA height map extractor is ready, we switch it out.

PR #564, opened 2024-12-26
[Feature] Support Publishing Sensor Images at Lower Frequency

TomaszTB:
Now you can do something like: 
```
ImageSensorPublishThread d455Publisher = new ImageSensorPublishThread(/* params */).setFrequencyLimit(10.0);
```

### DIFF
--- a/ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
+++ b/ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
@@ -1,6 +1,8 @@
 package us.ihmc.perception;
 
+import us.ihmc.commons.Conversions;
 import us.ihmc.commons.thread.RepeatingTaskThread;
+import us.ihmc.commons.thread.Throttler;
 import us.ihmc.communication.packets.Packet;
 import us.ihmc.ros2.ROS2Node;
 import us.ihmc.ros2.ROS2Topic;
@@ -13,6 +15,9 @@ public class ImageSensorPublishThread extends RepeatingTaskThread
 {
    private final Map<Integer, ROS2Topic<? extends Packet<?>>> imageKeyToTopicMap;
    private final RawImagePublisher publisher;
+   private final Throttler publishThrottler;
+   private double publishPeriodLimit = -1.0;
+
    private final ImageSensor imageSensor;
 
    public ImageSensorPublishThread(ROS2Node ros2Node, ImageSensor sensorToPublish, Map<Integer, ROS2Topic<? extends Packet<?>>> imageKeyToTopicMap)
@@ -22,6 +27,14 @@ public class ImageSensorPublishThread extends RepeatingTaskThread
       imageSensor = sensorToPublish;
       this.imageKeyToTopicMap = imageKeyToTopicMap;
       publisher = new RawImagePublisher(ros2Node);
+      publishThrottler = new Throttler();
+   }
+
+   @Override
+   public ImageSensorPublishThread setFrequencyLimit(double publishFrequencyLimit)
+   {
+      publishPeriodLimit = Conversions.hertzToSeconds(publishFrequencyLimit);
+      return this;
    }
 
    @Override
@@ -30,6 +43,10 @@ public class ImageSensorPublishThread extends RepeatingTaskThread
       try
       {  // Wait for images to be grabbed
          imageSensor.waitForGrab();
+
+         // Check if we should publish
+         if (publishPeriodLimit > 0.0 && !publishThrottler.run(publishPeriodLimit))
+            return;
 
          for (Entry<Integer, ROS2Topic<? extends Packet<?>>> imageEntry : imageKeyToTopicMap.entrySet())
          {
--- a/ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
+++ b/ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
@@ -1,22 +1,21 @@
 package us.ihmc.perception;
 
-import us.ihmc.commons.Conversions;
 import us.ihmc.commons.thread.RepeatingTaskThread;
-import us.ihmc.commons.thread.Throttler;
 import us.ihmc.communication.packets.Packet;
 import us.ihmc.ros2.ROS2Node;
 import us.ihmc.ros2.ROS2Topic;
 import us.ihmc.sensors.ImageSensor;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 
 public class ImageSensorPublishThread extends RepeatingTaskThread
 {
    private final Map<Integer, ROS2Topic<? extends Packet<?>>> imageKeyToTopicMap;
+   private final Map<Integer, Long> lastPublishedSequenceNumbers = new HashMap<>();
    private final RawImagePublisher publisher;
-   private final Throttler publishThrottler;
-   private double publishPeriodLimit = -1.0;
+   private boolean publishingIsThrottled = false;
 
    private final ImageSensor imageSensor;
 
@@ -27,13 +26,22 @@ public class ImageSensorPublishThread extends RepeatingTaskThread
       imageSensor = sensorToPublish;
       this.imageKeyToTopicMap = imageKeyToTopicMap;
       publisher = new RawImagePublisher(ros2Node);
-      publishThrottler = new Throttler();
+   }
+
+   public ImageSensorPublishThread(ROS2Node ros2Node,
+                                   ImageSensor sensorToPublish,
+                                   Map<Integer, ROS2Topic<? extends Packet<?>>> imageKeyToTopicMap,
+                                   double publishFrequencyLimit)
+   {
+      this(ros2Node, sensorToPublish, imageKeyToTopicMap);
+      setFrequencyLimit(publishFrequencyLimit);
    }
 
    @Override
    public ImageSensorPublishThread setFrequencyLimit(double publishFrequencyLimit)
    {
-      publishPeriodLimit = Conversions.hertzToSeconds(publishFrequencyLimit);
+      super.setFrequencyLimit(publishFrequencyLimit);
+      publishingIsThrottled = publishFrequencyLimit > 0.0;
       return this;
    }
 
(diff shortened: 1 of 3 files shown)


ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
```diff
@@ -24,19 +28,51 @@ public ImageSensorPublishThread(ROS2Node ros2Node, ImageSensor sensorToPublish,
       publisher = new RawImagePublisher(ros2Node);
    }
 
+   public ImageSensorPublishThread(ROS2Node ros2Node,
+                                   ImageSensor sensorToPublish,
+                                   Map<Integer, ROS2Topic<? extends Packet<?>>> imageKeyToTopicMap,
+                                   double publishFrequencyLimit)
```
TomaszTB: Constructor with the frequency limit!

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
```diff
@@ -24,19 +28,51 @@ public ImageSensorPublishThread(ROS2Node ros2Node, ImageSensor sensorToPublish,
...
+
    @Override
    protected void runTask()
    {
       try
-      {  // Wait for images to be grabbed
-         imageSensor.waitForGrab();
+      {  // If publisher frequency is not set, run at sensor's frequency
+         if (!publishingIsThrottled)
+            imageSensor.waitForGrab();
```
TomaszTB, line 45: Publisher will default to running at the sensor's frequency if the frequency limit is not set.

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
```diff
@@ -24,19 +28,51 @@ public ImageSensorPublishThread(ROS2Node ros2Node, ImageSensor sensorToPublish,
...
+            // Get the image to publish
             RawImage imageToPublish = imageSensor.getImage(imageKey);
+
+            // Skip if the image is null
+            if (imageToPublish == null)
+               continue;
+
+            // Skip if this image was already published
+            if (lastPublishedSequenceNumbers.containsKey(imageKey) && imageToPublish.getSequenceNumber() == lastPublishedSequenceNumbers.get(imageKey))
+               continue;
```
TomaszTB, line 61: This avoids republishing the same image more than once. It also prevents the streaming stuff from going crazy. Turns out publishing the same image twice messes things up.

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
PotatoPeeler3000: Sweet! This constructor looks good, great that you can set the publish requency

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
PotatoPeeler3000: Gosh I guess due to the nature of how the imageSensor.waitForGrab() is being used we don't really have a better way to do this currently

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
PotatoPeeler3000: It doesn't make sense that publishing the same image twice would cause problems? We get some image (doesn't matter what it is or what it contains), and that gets handed to the streaming stuff. Then the streaming stuff publish's that image. Does it have knowledge of information that's contained in the image? This seems like a good test case

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
```diff
@@ -24,19 +28,51 @@ public ImageSensorPublishThread(ROS2Node ros2Node, ImageSensor sensorToPublish,
...
+                                   ImageSensor sensorToPublish,
+                                   Map<Integer, ROS2Topic<? extends Packet<?>>> imageKeyToTopicMap,
+                                   double publishFrequencyLimit)
+   {
+      this(ros2Node, sensorToPublish, imageKeyToTopicMap);
+      setFrequencyLimit(publishFrequencyLimit);
+   }
+
+   @Override
+   public ImageSensorPublishThread setFrequencyLimit(double publishFrequencyLimit)
```
PotatoPeeler3000, line 32: Whats the purpose of this method? And its public so currently there are 2 ways to set the publish frequency.
- Call the first constructor, then call the .setFrequencyLimit() method.
- Call the second constructor.
This makes it difficult for the user to know which way things are meant to be done.

To fix this the second constructor  could have these lines of code directly inside of it and we don't have this method.
```
   public ImageSensorPublishThread(ROS2Node ros2Node,
                                   ImageSensor sensorToPublish,
                                   Map<Integer, ROS2Topic<? extends Packet<?>>> imageKeyToTopicMap,
                                   double publishFrequencyLimit)
   {
      this(ros2Node, sensorToPublish, imageKeyToTopicMap);
      super.setFrequencyLimit(publishFrequencyLimit);
      publishingIsThrottled = publishFrequencyLimit > 0.0;
   }

Then we also don't have to have .setFrequencyLimit() return the `ImageSensorPublishThread` so that may reduce confusion.

Comment (PotatoPeeler3000) [reply ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java:32]: Well that last sentence is not supposed to be in the code lol rip

File: ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java [hunk 5]
```diff
@@ -6,13 +6,17 @@
...
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 
 public class ImageSensorPublishThread extends RepeatingTaskThread
 {
    private final Map<Integer, ROS2Topic<? extends Packet<?>>> imageKeyToTopicMap;
+   private final Map<Integer, Long> lastPublishedSequenceNumbers = new HashMap<>();
    private final RawImagePublisher publisher;
+   private boolean publishingIsThrottled = false;
```
PotatoPeeler3000, line 18: What happens if we don't have this? This prevents the streaming bugs?

PotatoPeeler3000 (commented):
If I'm using the `ImageSensorPublishThread` I'd like to know, what constructor to call from the highest level, that has the least number of parameters. I couldn't find a good class that does what I'm talking about of this :(
But what do you think of this to increase readability and usability:
```
   public ImageSensorPublishThread(ROS2Node ros2Node, ImageSensor sensorToPublish, Map<Integer, ROS2Topic<? extends Packet<?>>> imageKeyToTopicMap)
   {
      this(ros2Node, sensorToPublish, imageKeyToTopicMap, DEFAULT_PUBLISH_RATE);
   }

   public ImageSensorPublishThread(ROS2Node ros2Node,
                                   ImageSensor sensorToPublish,
                                   Map<Integer, ROS2Topic<? extends Packet<?>>> imageKeyToTopicMap,
                                   double publishFrequencyLimit)
   {
      super(sensorToPublish.getSensorName() + "PublishThread");

      imageSensor = sensorToPublish;
      this.imageKeyToTopicMap = imageKeyToTopicMap;
      publisher = new RawImagePublisher(ros2Node);
      
      setFrequencyLimit(publishFrequencyLimit);
   }
   ```
   It reads top to bottom, and means if you don't want to set the frequency limit you don't have too and you get the default value.

File: ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java (hunk 3 above)
Comment (TomaszTB) [reply ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java:61]: The streaming stuff uses the image's acquisition time to give the image a timestamp. So I'm guessing it gets confused when it tries to send a frame with the same timestamp as the previous

File: ihmc-perception/src/main/java/us/ihmc/sensors/RealsenseImageSensor.java [hunk 6]
```diff
@@ -105,11 +105,16 @@ protected boolean grab()
...
       {
+         if (grabbedImages[COLOR_IMAGE_KEY] != null)
+            grabbedImages[COLOR_IMAGE_KEY].release();
          grabbedImages[COLOR_IMAGE_KEY] = RawImage.createWithBGRImage(bgrImage,
                                                                       realsense.getColorCameraIntrinsics(),
                                                                       new FramePose3D(colorPose),
                                                                       grabTime,
                                                                       grabSequenceNumber);
+
+         if (grabbedImages[DEPTH_IMAGE_KEY] != null)
```
PotatoPeeler3000, line 116: Because this `grabbedImage` object is not initialized in the constructor, or at the top of the class, all these null checks are needed. This isn't something we can fix in this PR, but something to consider going forward.

ihmc-perception/src/main/java/us/ihmc/sensors/ZEDImageSensor.java
```diff
@@ -257,7 +265,7 @@ private RawImage slMatToRawImage(Pointer slMatPointer, PixelFormat imagePixelFor
                                       sl_mat_get_ptr(slMatPointer, SL_MEM_GPU),
                                       sl_mat_get_step_bytes(slMatPointer, SL_MEM_GPU));
       return new RawImage(null,
-                          imageGpuMat,
+                          imageGpuMat.clone(),
```
PotatoPeeler3000, line 268: What does cloning do?

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
TomaszTB: We can't get rid of the `setFrequencyLimit()` method because it comes from the super class. If we don't want multiple ways of doing things, we gotta remove the parameter from the constructor

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
TomaszTB: This tells the thread whether to use the sensor's frequency, or the frequency set by the programmer

ihmc-perception/src/main/java/us/ihmc/sensors/RealsenseImageSensor.java
TomaszTB: I'd prefer having a null object than a dummy image tbh

ihmc-perception/src/main/java/us/ihmc/sensors/ZEDImageSensor.java
TomaszTB: It performs a deep copy of the GpuMat and returns the copy

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
ihmc-rosie: So does that mean something is wrong with the `ROS2SRTSensorStreamer` class? Or maybe at a lower level, I haven't looked at that class much yet

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
ihmc-rosie: Would it make more sense to get the sensor's frequency direction from the sensor? and then use that as the `ImageSensorPublishThread` frequency?

ihmc-perception/src/main/java/us/ihmc/sensors/RealsenseImageSensor.java
ihmc-rosie: That means when other people use the `RealsenseImageSensor` they have to put null checks everywhere. It also means you can't get the CameraIntrinsics before runtime, which in turn means you can't initialize the heightmap in the constructor. Its kind of a cascading affect the more the `RealsenseImageSensor` is used. I'm not sure we need to send a dummy image, but the null object everywhere isn't good

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
TomaszTB: I don't really like that idea because it can introduce unnecessary latency. Ideally we publish as soon as the sensor gets a new image. That's why the 'waitForGrab()' method is good. But if you want to publish at an arbitrary frequency, you can't use `waitForGrab()` so I just accept that you'll have some added latency.

ihmc-perception/src/main/java/us/ihmc/perception/ImageSensorPublishThread.java
TomaszTB: Yeah, to be honest I don't know where it's "wrong." On one hand I'm tempted to say you shouldn't stream the same image twice using the `ROS2SRTSensorStreamer` (because it's like telling the streamer "here's a new image to send" but giving it an old image). On the other hand you could say the `FFmpegEncoder` should know not to send frames that it's sent before. Perhaps even LibAV should handle that problem internally.

I'm leaning towards adding a check in the `FFmpegEncoder`. Should be pretty simple to do.

PotatoPeeler3000 (approved):

PR #566, opened 2024-12-26
Fixed up stand alone process to publish color and depth from a realsense

ihmc-rosie:


TomaszTB (approved):
You've got some unnecessary new lines in this file, but seems good otherwise.

PR #572, opened 2025-01-06
Created some unit tests for both the OpenCL and CUDA kernels

PotatoPeeler3000:
Added unit tests for the HeightMapUtils.cl and HeightMapUtils.cuh files. This will help with debugging the height map because it will verify that these methods are doing the correct things.

### DIFF
--- a/ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAProgram.java
+++ b/ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAProgram.java
@@ -106,7 +106,7 @@ public class CUDAProgram implements AutoCloseable
    {
       try
       {
-         String programName = programPath.getFile();
+         String programName = programPath.getFile().substring(programPath.getFile().lastIndexOf('/') + 1);
 
          InputStream programContentsStream = programPath.openStream();
          String programContents = new String(programContentsStream.readAllBytes());
@@ -122,7 +122,8 @@ public class CUDAProgram implements AutoCloseable
 
             for (int i = 0; i < headerPaths.length; i++)
             {
-               headerNames[i] = headerPaths[i].getFile();
+               String path = headerPaths[i].getFile();
+               headerNames[i] = path.substring(path.lastIndexOf('/') + 1);
 
                InputStream headerInputStream = headerPaths[i].openStream();
                headerContents[i] = new String(headerInputStream.readAllBytes());
--- a/ihmc-perception/src/main/resources/openCL/HeightMapUtils.cl
+++ b/ihmc-perception/src/main/resources/openCL/HeightMapUtils.cl
@@ -1,13 +1,3 @@
-int key_to_x_index(int key, int center_index)
-{
-   return key % (2 * center_index + 1);
-}
-
-int key_to_y_index(int key, int center_index)
-{
-   return key / (2 * center_index + 1);
-}
-
 int indices_to_key(int x_index, int y_index, int centerIndex)
 {
    return x_index + y_index * (2 * centerIndex + 1);
@@ -25,6 +15,11 @@ int coordinate_to_key(float2 coordinates, float2 center, float resolution, int c
    return indices_to_key(x_index, y_index, center_index);
 }
 
+int2 coordinate_to_indices(float2 coordinates, float2 center, float resolution, int center_index)
+{
+   return (int2) (coordinate_to_index(coordinates.x, center.x, resolution, center_index), coordinate_to_index(coordinates.y, center.y, resolution, center_index));
+}
+
 float index_to_coordinate(int index, float center, float resolution, int center_index)
 {
    return (index - center_index) * resolution + center;
@@ -35,13 +30,18 @@ float2 indices_to_coordinate(int2 index, float2 center, float resolution, int ce
    return (float2) (index_to_coordinate(index.x, center.x, resolution, center_index), index_to_coordinate(index.y, center.y, resolution, center_index));
 }
 
+int key_to_x_index(int key, int center_index)
+{
+   return key % (2 * center_index + 1);
+}
+
+int key_to_y_index(int key, int center_index)
+{
+   return key / (2 * center_index + 1);
+}
+
 float2 key_to_coordinate(int key, float2 center, float resolution, int center_index)
 {
    int2 index = (int2) (key_to_x_index(key, center_index), key_to_y_index(key, center_index));
    return indices_to_coordinate(index, center, resolution, center_index);
-}
-
-int2 coordinate_to_indices(float2 coordinates, float2 center, float resolution, int center_index)
-{
-   return (int2) (coordinate_to_index(coordinates.x, center.x, resolution, center_index), coordinate_to_index(coordinates.y, center.y, resolution, center_index));
 }
--- a/ihmc-perception/src/main/resources/openCL/HeightMapUtilsTest.cl
+++ b/ihmc-perception/src/main/resources/openCL/HeightMapUtilsTest.cl
@@ -1,0 +1,26 @@
+void kernel test_indices_to_coordinate(global int* index,
+                                       global float* center,
+                                       global float* resolution,
+                                       global int* centerIndex,
+                                       global float* xResultCopiedToHost,
+                                       global float* yResultCopiedToHost)
+{
+    int gid = get_global_id(0);
+
+    printf("index: %d\n", index[gid]);
+    printf("center: %f\n", center[gid]);
+    printf("resolution: %f\n", resolution[gid]);
+    printf("centerIndex: %d\n", centerIndex[gid]);
+
+    int2 indexForConversion = (int2) (index[gid], 1);
+    float2 centerForConversion = (float2) (center[gid], center[gid]);
+    float resolutionForConversion = resolution[gid];
+    int centerIndexForConversion = centerIndex[gid];
+
+    float2 resultCoordinate = indices_to_coordinate(indexForConversion,
+                                                    centerForConversion,
+                                                    resolutionForConversion,
+                                                    centerIndexForConversion);
+    xResultCopiedToHost[gid] = resultCoordinate.x;
+    yResultCopiedToHost[gid] = resultCoordinate.y;
+}
--- a/ihmc-perception/src/main/resources/openCL/HeightMapUtilsTest.cl
+++ b/ihmc-perception/src/main/resources/openCL/HeightMapUtilsTest.cl
@@ -24,3 +24,33 @@ void kernel test_indices_to_coordinate(global int* index,
     xResultCopiedToHost[gid] = resultCoordinate.x;
     yResultCopiedToHost[gid] = resultCoordinate.y;
 }
+
+void kernel test_coordinate_to_indices(global float* coordinate,
+                                       global float* center,
+                                       global float* resolution,
+                                       global int* center_index,
+                                       global float* xResultCopiedToHost,
+                                       global float* yResultCopiedToHost)
+{
+    int gid = get_global_id(0);
+
+    printf("index: %f\n", coordinate[gid]);
+    printf("center: %f\n", center[gid]);
+    printf("resolution: %f\n", resolution[gid]);
+    printf("centerIndex: %d\n", center_index[gid]);
+
+    float2 coordinateForConversion = (float2) (coordinate[gid], coordinate[gid]);
+    float2 centerForConversion = (float2) (center[gid], center[gid]);
+    float resolutionForConversion = resolution[gid];
+    int centerIndexForConversion = center_index[gid];
+
+    int2 resultCoordinate = coordinate_to_indices(coordinateForConversion,
+                                                    centerForConversion,
+                                                    resolutionForConversion,
+                                                    centerIndexForConversion);
+
+    xResultCopiedToHost[gid] = resultCoordinate.x;
+    yResultCopiedToHost[gid] = resultCoordinate.y;
+}
+
+                                       
--- a/ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtils.cuh
+++ b/ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtils.cuh
@@ -1,5 +1,3 @@
-extern "C"
-
 __device__ float index_to_coordinate(int index, float center, float resolution, int center_index)
 {
     return (index - center_index) * resolution + center;
--- a/ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtils.cuh
+++ b/ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtils.cuh
@@ -1,0 +1,11 @@
+extern "C"
+
+__device__ float index_to_coordinate(int index, float center, float resolution, int center_index)
+{
+    return (index - center_index) * resolution + center;
+}
+
+__device__ float2 indices_to_coordinate(int2 index, float2 center, float resolution, int center_index)
+{
+    return make_float2(index_to_coordinate(index.x, center.x, resolution, center_index), index_to_coordinate(index.y, center.y, resolution, center_index));
+}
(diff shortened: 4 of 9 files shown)


ihmc-perception/src/main/resources/openCL/HeightMapUtils.cl
PotatoPeeler3000, line 1: Organized this file

ihmc-perception/src/main/resources/openCL/HeightMapUtilsTest.cl
PotatoPeeler3000, line 1: Couple kernels that allow for testing some of the HeightMapUtils.cl methods. I return the result from the `indices_to_coordinate` and `coordinate_to_indices` to check the expected value in Java

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtils.cuh
PotatoPeeler3000, line 1: Started to create a HeightMapUtils.cuh for cuda, this way I can test that these methods return the same thing for CUDA and OpenCL

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtilsTest.cu
PotatoPeeler3000, line 1: This is the kernel file that will call the HeightMapUtils.cuh methods and return the result to Java

ihmc-perception/src/test/java/us/ihmc/perception/cuda/examples/ExampleCUDAKernel.java
```diff
@@ -22,6 +24,8 @@ public class ExampleCUDAKernel
          __global__
          void add_arrays(int n, float *x, float *y)
          {
+            // Printing Hello World here with a newline character at the end to show how it can be used in a string
```
PotatoPeeler3000, line 27: Found this useful so I left it around

ihmc-perception/src/test/java/us/ihmc/perception/gpuHeightMap/HeightMapUtilsTest.java
PotatoPeeler3000, line 1: Big boy, kind of a mess, kind of documented lol. Maybe in another pass can clean this up

ihmc-perception/src/test/java/us/ihmc/perception/opencl/OpenCLKernelsSimpleTest.java
PotatoPeeler3000, line 1: Simple OpenCL test, we don't have any examples of these so wanted to start on that

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtils.cuh
TomaszTB: I'm used to header files containing declaration, but not definition. This file contains function definitions, so maybe this should be a `.cu` file? Although there's nothing stopping us from using head files like this. @ds58 What do you think? I'm cool with it either way.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtils.cuh
```diff
@@ -0,0 +1,21 @@
+extern "C"
```
TomaszTB: I thought we needed the `extern "C"` before every function definition. Is that not the case? Does this work?

ihmc-perception/src/test/java/us/ihmc/perception/cuda/examples/ExampleCUDAKernel.java
```diff
@@ -11,8 +11,10 @@
 import static org.bytedeco.cuda.global.cudart.*;
 
 /**
- * This is a simple example of a kernel that adds two arrays together. The ways in which kernels can be run vary such that you can create all the JavaCPP Pointers
- * inside a try-with-resources. The user can create more threads on the GPU. This example attempted to keep things really simple and avoided most of that overhead.
+ * This is a simple example of a kernel that adds two arrays together. The ways in which kernels can be run vary such that you can create all the JavaCPP
+ * Pointers
+ * inside a try-with-resources. The user can create more threads on the GPU. This example attempted to keep things really simple and avoided most of that
+ * overhead.
```
TomaszTB, line 17: This formatting is weird :(

TomaszTB (commented):
Would it be possible to move the test cl and cu files into test/resources? I think it should be easy with the CUDA ones. Not sure about OpenCL ones though.

ihmc-perception/src/test/java/us/ihmc/perception/gpuHeightMap/HeightMapUtilsTest.java
```diff
@@ -0,0 +1,439 @@
...
+import us.ihmc.perception.cuda.CUDAProgram;
+import us.ihmc.perception.cuda.CUDAStreamManager;
+import us.ihmc.perception.opencl.OpenCLManager;
+
+import java.net.URL;
+
+import static org.bytedeco.cuda.global.cudart.*;
+import static org.bytedeco.cuda.global.cudart.cudaMemcpyAsync;
+
+@Tag("ihmc-perception")
```
ds58: You dont need this tag. This actually won't run the way it is. The only test category that runs under ihmc-perception is `fast`. See [here](https://github.com/ihmcrobotics/ihmc-open-robotics-software/blob/develop/.github/workflows/gradle-test-fast.yml#L156)

The entire ihmc-perception project runs on a runner that has a GPU, so you don't need any tags.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtils.cuh
ds58: it's usually situational. But, I personally prefer headers to not have implementation.

PotatoPeeler3000:
> Would it be possible to move the test cl and cu files into test/resources? I think it should be easy with the CUDA ones. Not sure about OpenCL ones though.

I looked into this. The reason they are in the same folder is for the #include. If you want to move the HeightMapUtilsTest.cu file to the test/resources you need to have something like `#include "../../../../../../../main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtils.cuh"` for including the header file. And its rather fragile if the path changes, and harder to read

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtils.cuh
PotatoPeeler3000: Yep it does work, I've actually removed the `extern "C"` line because you only need it for __global__ methods. Its not needed for the __device__ methods

ihmc-perception/src/test/java/us/ihmc/perception/cuda/examples/ExampleCUDAKernel.java
PotatoPeeler3000: Its what the formatter does lol, sorry but its no longer customized

ihmc-perception/src/test/java/us/ihmc/perception/gpuHeightMap/HeightMapUtilsTest.java
PotatoPeeler3000: Ohh on interesting, just pushed a new commit will see what happens cause I removed the tag

PotatoPeeler3000:
> I'm used to header files containing declaration, but not definition. This file contains function definitions, so maybe this should be a .cu file? Although there's nothing stopping us from using head files like this. @ds58 What do you think? I'm cool with it either way.

These are simple methods that are self contained, from what I've found this seems reasonable to have in a header file

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAProgram.java
```diff
@@ -106,7 +106,7 @@ public CUDAProgram(URL programPath, URL[] headerPaths, String... compilationOpti
    {
       try
       {
-         String programName = programPath.getFile();
+         String programName = programPath.getFile().substring(programPath.getFile().lastIndexOf('/') + 1);
```
PotatoPeeler3000, line 109: This substring is the actual file name. Before it was `/path/to/fileName.cu` and now its `fileName.cu`

PotatoPeeler3000 (commented):
Fixed CI!

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAProgram.java
TomaszTB: Oh wow, nice catch. Kinda amazed it's been working without this

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/HeightMapUtils.cuh
TomaszTB: Could we make this a `.cu` file then?

ihmc-perception/src/test/java/us/ihmc/perception/cuda/examples/ExampleCUDAKernel.java
```diff
@@ -22,6 +24,8 @@ public class ExampleCUDAKernel
          __global__
          void add_arrays(int n, float *x, float *y)
          {
+            // Printing Hello World here with a newline character at the end to show how it can be used in a string
+            printf("Hello World\\n");
```
TomaszTB, line 28: I think there's an extra back slash in there. Doesn't that escape the newline character and just print "\n"?

ihmc-perception/src/test/java/us/ihmc/perception/gpuHeightMap/HeightMapUtilsTest.java
```diff
@@ -0,0 +1,438 @@
...
+
+      kernel.close();
+      program.close();
+
+      gpuCenterPointer.close();
+      gpuResolutionPointer.close();
+      gpuXHostPointer.close();
+      gpuYHostPointer.close();
+
+      cudart.cudaStreamDestroy(stream);
```
TomaszTB: Should release this using the CUDAStreamManager since it was gotten from it

ihmc-perception/src/test/java/us/ihmc/perception/gpuHeightMap/HeightMapUtilsTest.java
```diff
@@ -0,0 +1,438 @@
...
+
+      kernel.close();
+      program.close();
+
+      gpuCoordinatePointer.close();
+      gpuCenterPointer.close();
+      gpuXHostPointer.close();
+      gpuYHostPointer.close();
+
+      cudart.cudaStreamDestroy(stream);
```
TomaszTB: Also release using CUDAStreamManager

ihmc-perception/src/test/java/us/ihmc/perception/gpuHeightMap/HeightMapUtilsTest.java
```diff
@@ -0,0 +1,438 @@
...
+
+      cudaMallocAsync(gpuCenterPointer, gpuCenterPointer.sizeof() * centerPointerSize, stream);
+      cudaMallocAsync(gpuResolutionPointer, gpuResolutionPointer.sizeof() * resolutionPointerSize, stream);
+      cudaMallocAsync(gpuXHostPointer, gpuXHostPointer.sizeof() * xHostPointerSize, stream);
+      cudaMallocAsync(gpuYHostPointer, gpuYHostPointer.sizeof() * yHostPointerSize, stream);
+
+      cudaMemcpyAsync(gpuCenterPointer, centerPointer, (long) centerPointer.sizeof() * centerPointerSize, cudaMemcpyDefault, stream);
+      cudaMemcpyAsync(gpuResolutionPointer, resolutionPointer, (long) resolutionPointer.sizeof() * resolutionPointerSize, cudaMemcpyDefault, stream);
+      cudaMemcpyAsync(gpuXHostPointer, xHostPointer, (long) xHostPointer.sizeof() * xHostPointerSize, cudaMemcpyDefault, stream);
+      cudaMemcpyAsync(gpuYHostPointer, yHostPointer, (long) yHostPointer.sizeof() * yHostPointerSize, cudaMemcpyDefault, stream);
```
TomaszTB: We added malloc and memcpy methods to CUDATools to remove the `pointer.sizeof() * elementCount` boilerplate. You can just pass in the element count and the method calls `sizeof()` internally. Just so you can type a bit less in the future!

TomaszTB (changes requested):

ihmc-perception/src/test/java/us/ihmc/perception/cuda/examples/ExampleCUDAKernel.java
PotatoPeeler3000: That's why I left this around because if you want to print the \n in a string, you need the additional `\` character. So without the additional `\` character. We got problems

ihmc-perception/src/test/java/us/ihmc/perception/gpuHeightMap/HeightMapUtilsTest.java
PotatoPeeler3000: Yep! Good catch

ihmc-perception/src/test/java/us/ihmc/perception/gpuHeightMap/HeightMapUtilsTest.java
PotatoPeeler3000: Yep yep yep

ihmc-perception/src/test/java/us/ihmc/perception/gpuHeightMap/HeightMapUtilsTest.java
PotatoPeeler3000: Ok, gonna use that. Its not great that we need to use CUDATools  and use CUDAStreamManager because its not clear what methods should be used, especially when closing the stream. It might be better to have the tools method just do the xHostPointer.sizeOf() * xHostPointerSize + 1 calculation, and you still call the cudaMemcpyAsync().

PotatoPeeler3000:
> Could we make this a .cu file then?

I don't think we should do that, this is more clear that it doesn't have any kernels in it. Its a header file

ihmc-perception/src/test/java/us/ihmc/perception/gpuHeightMap/HeightMapUtilsTest.java
TomaszTB: I get what you mean. It'd be pretty easy to add a method that does the `pointer.sizeof() * elements` calculation, but I'm afraid it wouldn't reduce the boilerplate by much. 

To be honest I'm not too concerned about the CUDATools memcpy and malloc methods since using the CUDA functions directly is still just as valid. I didn't mean to say that you should make the change here. The code was fine :sweat_smile: 

The CUDAStreamManager is more of a problem though, since it breaks if you destroy the stream using the CUDA functions. Unfortunately I can't think of a better way of doing the CUDAStreamManager thing.

P.S. why do you need the +1? I don't think that's a standard pattern

ihmc-perception/src/test/java/us/ihmc/perception/cuda/examples/ExampleCUDAKernel.java
TomaszTB: Ohh I see

TomaszTB:
> > Could we make this a .cu file then?
> 
> I don't think we should do that, this is more clear that it doesn't have any kernels in it. Its a header file

I see your intentions. Kinda think the "Utils" part makes that clear, but I'm fine with it staying a header file. Should we make it a convention for the utils/tools files to be header files?

TomaszTB (approved):

PotatoPeeler3000:
> I see your intentions. Kinda think the "Utils" part makes that clear, but I'm fine with it staying a header file. Should we make it a convention for the utils/tools files to be header files?

Ehhh when we get enough header files maybe, that probably won't be for a while so I'm not worried about a convention at the moment

PR #574, opened 2025-01-06
Added simple CUDA unit tests

PotatoPeeler3000:
Added unit tests for some CUDA kernels.
Changed the CUDATools to throw exceptions when things report and error. This is so we can test those cases


ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
PotatoPeeler3000, line 1: On all of the error checking, throw an exception if there is one.

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDAProgramTest.java
```diff
@@ -139,7 +138,7 @@ public void testKernelWithHeader()
    }
 
    @Test
-   public void testLoadingKernelFromFile() throws URISyntaxException
```
PotatoPeeler3000, line 142: Unused

ihmc-perception/src/test/java/us/ihmc/perception/cuda/examples/CUDAExampleKernel.java
```diff
@@ -16,7 +16,7 @@
  * inside a try-with-resources. The user can create more threads on the GPU. This example attempted to keep things really simple and avoided most of that
  * overhead.
  */
-public class ExampleCUDAKernel
+public class CUDAExampleKernel
```
PotatoPeeler3000, line 19: I kept searching for `CUDAExample...` and nothing would come up cause it was called `ExampleCUDA...` so I changed it.

ihmc-perception/src/test/java/us/ihmc/perception/cuda/examples/CUDAExampleKernel2D.java
```diff
@@ -16,12 +16,12 @@
 /**
  * This is an example of using a more complex CUDA kernel that takes in an OpenCV {@link Mat} object. We do this often for kernels, so this example walks you
  * through how it would work for a simple case.
- * This uses more of the standard approach then the {@link ExampleCUDAKernel} did because we use the {@link CUDAStreamManager} and we are creating objects
+ * This uses more of the standard approach then the {@link CUDAExampleKernel} did because we use the {@link CUDAStreamManager} and we are creating objects
  * in the try-with-resources statement.
  */
-public class ExampleCUDAKernel2D
+public class CUDAExampleKernel2D
```
PotatoPeeler3000, line 22: Same story here, changed it cause I kept searching for the wrong thing

ihmc-perception/src/test/java/us/ihmc/perception/opencl/OpenCLSimpleKernelsTest.java
PotatoPeeler3000, line 1: Added a helpful comment and renamed this class

ihmc-perception/src/test/resources/us/ihmc/perception/cuda/CUDASimpleKernels.cu
PotatoPeeler3000, line 1: A couple simple CUDA kernels to test different cases

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,111 @@
...
+import java.net.URL;
+
+import static org.bytedeco.cuda.global.cudart.*;
+import static org.bytedeco.cuda.global.cudart.cudaMemcpyDefault;
+import static org.junit.jupiter.api.Assertions.*;
+
+public class CUDASimpleKernelsTest
+{
+   @Test
+   public void testPassVariableToGPU()
```
PotatoPeeler3000: Make sure that we can pass a value to the gpu and get back a value we expect.

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,111 @@
...
+   }
+
+   /**
+    * Here the goal is to pass in the wrong number of variables and see that the GPU crashes accordingly.
+    * This test shows what happens when you make that mistake.
+    */
+   // TODO: this test fails when releasing the stream. That seems wrong because there are 6 parameters being passed into the kernel
+   // TODO: So it would seem like the kernel should not even run, but it does... (this could use some investigating)
+   @Test
+   public void testWrongNumberOfKernelVariables()
```
PotatoPeeler3000: Not sure how the kernel is running with the wrong set of parameters. This test is failing because the CUDAStreamManager.releaseStream(stream); throws an exception. Not sure why. I would have thought it should fail when it tries to run the kernel with the wrong amount of arguments.
Anyways, this test is failing with a TODO that is should be fixed

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,111 @@
...
+      CUDAStreamManager.releaseStream(stream);
+   }
+
+
+   /**
+    * A stream needs to be created before creating a CUDAProgram. This should throw an exception if the program is created first.
+    * This test checks that an exception is thrown and that `cudaErrorDeviceUninitialized` is in the message
+    */
+   @Test
+   public void testCreateStreamAfterManager()
```
PotatoPeeler3000: An example of what not to do

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,111 @@
...
+      assertTrue(thrown.getMessage().contains("cudaErrorDeviceUninitialized"));
+   }
+
+   /**
+    * When writing a CUDA kernel, each kernel needs to declare (extern "C") just above the kernel.
+    * If this doesn't happen, that kernel should fail as it won't be loaded correctly.
+    * Making sure that if fails here...
+    */
+   @Test
+   void testNoExternCInGPUKernel()
```
PotatoPeeler3000: An example of what not do to

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
```diff
@@ -134,7 +136,10 @@ public static void checkNVJPEGError(int errorCode)
             case 10 -> "NVJPEG_STATUS_INCOMPLETE_BITSTREAM";
             default -> "UNKNOWN";
          };
-         LogTools.error("NVJPEG Error ({}): {}", errorCode, errorName);
+
+         String errorMessage = String.format("NVJPEG Error (%d): %s", errorCode, errorName);
+         LogTools.error(errorMessage);
+         throw new RuntimeException(errorMessage);
```
PotatoPeeler3000: Was there a reason to not throw exceptions here? It makes it way easier to test if we throw an exception? And the kernel doesn't work with the error so there isn't a reason to continue the process

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDAProgramTest.java
PotatoPeeler3000: Just the `throws URISYntaxException`

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,209 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+
+import java.net.URL;
+
+import static org.bytedeco.cuda.global.cudart.*;
+import static org.bytedeco.cuda.global.cudart.cudaMemcpyDefault;
+import static org.junit.jupiter.api.Assertions.*;
+
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
```
PotatoPeeler3000: I don't like that I have to do this, if the test that requires this fails on CI, I might disable it

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
PotatoPeeler3000: This was helpful for me

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,209 @@
...
+   }
+
+   @Test
+   public void testNoSemicolonInKernel()
+   {
+      URL kernelPath = getClass().getResource("CUDAKernelCodeWrong.cu");
+
+      stream = CUDAStreamManager.getStream();
+
+      // Since we haven't declared the (extern "C"), we expect this kernel to fail loading
```
PotatoPeeler3000, line 182: Good example of that each kernel needs the extern "C" line

ihmc-perception/src/test/resources/us/ihmc/perception/cuda/CUDAKernelCodeWrong.cu
```diff
@@ -0,0 +1,7 @@
+// Simple kernel that only takes in an int for testing
+extern "C"
+__global__ void no_semicolon(int value)
+{
+    // When this is compiled this should fail because its not proper C code (missing ;)
+    printf("Only an int being passed to this kernel")
```
PotatoPeeler3000, line 6: Bad C code, ehhh easy to make, thought this wasn't bad

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
TomaszTB: Only reason I could think of is to not crash the whole autonomy process if one thing goes wrong. I don't know it that's a great reason, though. 

Also, what's the benefit of both throwing an exception and printing the message? Maybe remove the logging?

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,209 @@
...
+      }
+
+      if (program != null)
+      {
+         program.close();
+      }
+
+      if (stream != null)
+      {
+         CUDAStreamManager.releaseStream(stream);
```
TomaszTB, line 39: Idk if it's crucial, but maybe set the stream to null here too?

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,209 @@
...
+      URL kernelPath = getClass().getResource("CUDASimpleKernels.cu");
+
+      stream = CUDAStreamManager.getStream();
+      program = new CUDAProgram(kernelPath);
+      kernel = program.loadKernel("pass_in_variable");
+
+      FloatPointer gpuResult = new FloatPointer();
+
+      // The convention is to use CUDATools for allocating memory at the moment
+      CUDATools.mallocAsync(gpuResult, gpuResult.sizeof(), stream);
```
TomaszTB, line 55: You should pass in the number of floats you want to allocate, not the sizeof.

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,209 @@
...
+//      cudaMallocAsync(gpuResult, gpuResult.sizeof(), stream);
+
+      kernel.withInt(5).withPointer(gpuResult);
+      kernel.run(stream, new dim3(), new dim3(), 0);
+
+      FloatPointer cpuResult = new FloatPointer(1F);
+
+      // The convention is to use CUDATools for allocating memory at the moment
+      CUDATools.memcpyAsync(cpuResult, gpuResult, cpuResult.sizeof(), stream);
```
TomaszTB, line 64: Same thing here. Number of elements you want to copy instead of the sizeof each element

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,209 @@
...
+      cudaFreeAsync(gpuResult, stream);
+
+      gpuResult.close();
+   }
+
+   /**
+    * Test that when you pass in the wrong type into the GPU kernel that it fails.
+    * WHen the kernel expects and int and you pass a in float, it should fail.
+    */
+   // TODO this test should fail but seems to pass even when the wrong paramter is passed into the kernel
```
TomaszTB: I wonder if it does type casting? Or if `sizeof(float) == sizeof(int)`, it might successfully pass in the bits but it's a junk `int` value in the kernel

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
TomaszTB: It's probably due to the asynchronous calls. When we call `kernel.run(/*stuff*/)`, it tells CUDA to run the kernel whenever it's ready. CUDA doesn't realize that the number of parameters doesn't match until it tries to run it, sometime after the call to `kernel.run(/*stuff*/)`. 

CUDA reports these errors in subsequent CUDA function calls (once it knows there was an error), so that's why you're seeing errors from the CUDAStreamManager. 

You should check the error value returned by the call to `cudaStreamSynchronize()` after the kernel. I would guess that'll also report the error experienced by the kernel.

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,209 @@
...
+      kernel = program.loadKernel("pass_in_variable");
+
+      // Pass in way to many variables and see what happens
+      kernel.withInt(5).withFloat(4.0f).withLong((long) 2.0);
+      kernel.withInt(5).withFloat(4.0f).withLong((long) 2.0);
+
+      // TODO how is this not failing when the kernel is trying to pass in a lot of variables that don't get defined in the kernel on the GPU
+      kernel.run(stream, new dim3(), new dim3(), 0);
+
+      cudaStreamSynchronize(stream);
```
TomaszTB: This might report the error (probably check for the runtime exception here?)
```suggestion
      CUDATools.checkCUDAError(cudaStreamSynchronize(stream));
```

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,209 @@
...
+   }
+
+   /**
+    * This test is meant to test the case where you don't synchronize the stream after running the kernel.
+    * Need to do some more testing to see why this still works, and if there is a better way to test this.
+    * That is why this test is disabled at the moment, it doesn't appear to be that useful
+    */
+   @Disabled
+   @Test
+   public void testNotSynchronizingStream()
```
TomaszTB: This is testing for a race condition, and that's kinda hard.. you'd have to get lucky for it to pass every time.

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+      kernel.withInt(5);
+
+      kernel.run(stream, new dim3(), new dim3(), 0);
+
+      cudaStreamSynchronize(stream);
+   }
+
+   /**
+    * A stream needs to be created before creating a CUDAProgram. This should throw an exception if the program is created first.
```
TomaszTB: Ohhh that's an interesting detail you found! I'm inclined to call that a bug. 

So there are two CUDA APIs available: CUDA Runtime API and CUDA Driver API. Driver API is for low level stuff, and it's less friendly. Runtime API is slightly higher level, and should be preferred over the Driver API.

Unfortunately we can't run kernels using the Runtime API because that seems to require compiling using NVCC. So we use Driver API in the CUDAProgram and CUDAKernel. 

The Runtime API will initialize CUDA correctly upon the first call to one of its functions. Seems like the Driver API doesn't do that? So when you create a stream, it calls a Runtime API function and initializes CUDA for the whole process. But when you create a CUDAProgram first, it tries to call Driver API functions before CUDA is initialized. 

I think the solution is to call some Runtime API function as the first thing in the CUDAProgram's constructor. Then we wouldn't have to worry about CUDA not being initialized. 

Let me know if you'd like me to attempt a fix for this.

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
```diff
@@ -0,0 +1,209 @@
...
+      {
+         CUDAProgram program = new CUDAProgram(kernelPath);
+         program.close();
+      });
+
+      assertTrue(thrown.getMessage().contains("cudaErrorDeviceUninitialized"));
+   }
+
+   /**
+    * When writing a CUDA kernel, each kernel needs to declare (extern "C") just above the kernel.
```
TomaszTB, line 158: I found out the `extern "C"` is only needed for `__global__` functions. `__device__` functions seem to be fine without it. Just spreading a bit of knowledge!

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
TomaszTB: I think that comment isn't meant to be there

TomaszTB (changes requested):
Looks good for the most part, but should we be merging so many disabled tests that we still have to figure out into develop?

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
ds58: I dont think we should crash the whole process if something goes wrong

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
TomaszTB: I played around a bit, and can confirm that the bits are passed in successfully but the value is garbage. Seems like we can't expect an exception to be thrown in this case :(

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
PotatoPeeler3000: The whole process doesn't function properly if the kernel doesn't work. By continuing to run the program its not clear to the user that something is wrong. By crashing the program the user knows that whatever kernel was being called won't work

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
PotatoPeeler3000: Ok notes this, will check this out

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
PotatoPeeler3000: Ya it was, is it unclear that its talking about the extern "C" in the kernel?

PotatoPeeler3000:
> Looks good for the most part, but should we be merging so many disabled tests that we still have to figure out into develop?

It gives us something to figure out. Better to know what is wrong in the tests then have know idea.

PotatoPeeler3000:
> Only reason I could think of is to not crash the whole autonomy process if one thing goes wrong. I don't know it that's a great reason, though. Also, what's the benefit of both throwing an exception and printing the message? Maybe remove the logging?

Ya I could remove the logging, that isn't super important anymore

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
PotatoPeeler3000: Na, it didn't change anything. I tried that too to fix one of the tests, but the problem still existed

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
PotatoPeeler3000: Ok will have to look into this

PotatoPeeler3000: We could however return the bad value and ensure it doesn't equal the value that was passed into the kernel

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
PotatoPeeler3000: Ok I'll check that out

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
PotatoPeeler3000: Hmmmmmmm, ok I don't understand that

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
PotatoPeeler3000: Don't know yet, we can leave it for now cause the test shows the correct order of how things should be created.

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
PotatoPeeler3000: Yep, that is correct!

ihmc-perception/src/test/java/us/ihmc/perception/cuda/examples/CUDAExampleKernel2D.java
ds58: Shouldnt it be CUDAKernel2DExample?

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
ds58: That's not necessarily true; depends on the responsibility of the process and everything that it's running

ds58 (changes requested):
This is a general thing, but I think we should only start with working tests. You have a lot of tests that aren't complete / don't work yet. I think we should wait on those.

PotatoPeeler3000:
> This is a general thing, but I think we should only start with working tests. You have a lot of tests that aren't complete / don't work yet. I think we should wait on those.

The general thing would have been to write the tests to begin with. However, we are way past that. When someone has time the failing tests can be fixed, at least we know there is a problem that way. I can't wait as I'm trying to rely on the CUDA code but I have zero idea why things don't work sometimes. The tests give me a better idea even if they are failing.

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
TomaszTB: That's the test above. This test is for the missing semicolon

ihmc-perception/src/test/java/us/ihmc/perception/cuda/CUDASimpleKernelsTest.java
TomaszTB: I looked more into this because I didn't fully understand it either. 

Turns out passing in too many variables results in illegal memory access. The error is indeed reported by `cudaStreamSynchronize()`, but it also happens to be an error that corrupts the CUDA state. So once it occurs, all CUDA calls will fail until the process terminates. 

I'm not sure how we could run this test successfully. It might require initializing CUDA and running the test in another process that can terminate after the test. 

https://stackoverflow.com/questions/56329377/reset-cuda-context-after-exception/56330491#56330491

ihmc-perception/src/test/java/us/ihmc/perception/cuda/examples/CUDAExampleKernel.java
ds58: Can it be CUDAKernelExample?

ds58 (dismissed):
I think the 2 disabled CUDA tests should still be removed. Tomasz brought up excellent points about why the expectations for those tests are too high. If you can figure out a way to not corrupt the CUDA state for testWrongNumberOfKernelVariables that would be amazing, but probably a lot of work to run it in a separate process and get the result

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
```diff
@@ -108,7 +108,19 @@ public static void checkCUDAError(int errorCode)
          {
             String errorMessage = String.format("CUDA Error (%s): %s", errorName.getString(), errorString.getString());
             LogTools.error(errorMessage);
-            throw new RuntimeException(errorMessage);
+         }
+      }
+   }
+
+   public static void throwExceptionIfNotSuccessful(int errorCode) throws Exception
```
TomaszTB: Can we name this `throwOnCUDAError` or something of that sort? I want it to be specific to CUDA, since the message would be incorrect for NVJPEG and NVRTC. Also the current name is very long haha

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAProgram.java
```diff
@@ -218,6 +218,7 @@ private void initialize(String programName, String programCode, String[] headerN
       // Load the module using the PTX
       error = cuModuleLoadData(module, ptx);
       checkCUDAError(error);
+      throwExceptionIfNotSuccessful(error);
```
TomaszTB: Should we throw if any of the NVRTC calls before this one fail? That would require a `throwOnNVRTCError` method though

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAProgram.java
```diff
@@ -218,6 +218,7 @@ private void initialize(String programName, String programCode, String[] headerN
       // Load the module using the PTX
       error = cuModuleLoadData(module, ptx);
       checkCUDAError(error);
```
TomaszTB: Probably no need to log the error if we're gonna throw an exception anyway

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
```diff
@@ -27,10 +28,11 @@ public class CUDAKernel implements AutoCloseable
 
    private int error;
 
-   public CUDAKernel(String name, CUmod_st kernelModule)
+   public CUDAKernel(String name, CUmod_st kernelModule) throws Exception
    {
       error = cuModuleGetFunction(kernelFunction, kernelModule, name);
       checkCUDAError(error);
```
TomaszTB: Probably no need to log the error if we're gonna throw an exception anyway

TomaszTB (changes requested):
Could you update with develop and add the exception handling to the `CUDADepthColorizer`? I think it'll need it too

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
PotatoPeeler3000: Ya went with `throwCUDAError`

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAProgram.java
PotatoPeeler3000: Ehhh gonna leave it for now

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAProgram.java
PotatoPeeler3000: True

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
PotatoPeeler3000: True

PotatoPeeler3000:
> Could you update with develop and add the exception handling to the `CUDADepthColorizer`? I think it'll need it too

Yep good call, just did that and pushed

PotatoPeeler3000:
> Can it be CUDAKernelExample?

Gonna leave it for now if that's alright

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
```diff
@@ -40,6 +42,8 @@ public class RapidHeightMapManager
    private GpuMat deviceDepthImage;
    private final Mat hostDepthImage = new Mat();
    private BytedecoImage heightMapBytedecoImage;
+   private MutableBytePointer bytedecoByteBufferPointer;
```
TomaszTB: We don't like this thing. It scares me. We want to get rid of it...

Anyway, seems like you have some residue from testing stuff here. Could it be cleaned up?

TomaszTB (approved):
I pushed a bit of cleanup. Just a bit I didn't touch with the `RapidHeightMapManager`. Not sure if you want to keep it for whatever reason (I'd prefer if you didn't). Seems good to me!

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
PotatoPeeler3000: Ok yep totally right, removed it!

PR #598, opened 2025-01-13
SVO logging

ds58:


ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
```diff
@@ -180,7 +185,8 @@ protected int openCamera()
    @Override
    public boolean isSensorRunning()
    {
-      return sl_is_opened(cameraID) && !lastGrabFailed;
+      boolean recentlyGrabbed = lastGrabTime != null && lastGrabTime.isAfter(Instant.now().minusSeconds(1));
+      return sl_is_opened(cameraID) && !lastGrabFailed && recentlyGrabbed;
```
calvertdw, line 190: :+1:

ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
```diff
@@ -109,6 +112,8 @@ protected boolean startSensor()
             sl_enable_positional_tracking(cameraID, positionalTrackingParameters, "");
          }
 
+         sl_enable_streaming(cameraID, SL_STREAMING_CODEC_H264, 8000, (short) streamingPort, -1, 0, 16084, CAMERA_FPS);
```
calvertdw, line 116: Does this mean it's always streaming, even when doing local tests & development?

ds58: Streaming is always enabled, not necessarily streaming to anything

calvertdw (approved):

ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
TomaszTB, line 116: You could use `StreamingTools.getOpenPort()` here. Though it's kinda nice knowing that ZED uses port 30000 and up. Up to you to decide which method you prefer.

TomaszTB (approved):

TomaszTB:
I imagine we'll add the sensor pose into the ZED Stream in the future? Or should that happen in this PR?

ds58:
> I imagine we'll add the sensor pose into the ZED Stream in the future? Or should that happen in this PR?

Future change

ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
ds58: I want to keep it like this, since ZED SDK typically always uses port 30000 or around there

PR #611, opened 2025-01-23
Direct CoM Streaming

stephenmcc:
Currently when sending CoM setpoints to the KST, they get heavily filtered on the controller side via integrating up error in the pelvis position. This sets up a direct pipeline to stream CoM setpoints to the controller via the precomputed ICP trajectory. It's disabled by default. Not sure how useful it will be because the latency is likely preferable, but figured it's nice to have.

### DIFF
--- a/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/BalanceManager.java
+++ b/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/BalanceManager.java
@@ -78,6 +78,7 @@ public class BalanceManager implements SCS2YoGraphicHolder
    private static final boolean USE_ERROR_BASED_STEP_ADJUSTMENT = true;
    private static final ReferenceFrame worldFrame = ReferenceFrame.getWorldFrame();
    private static final boolean viewCoPHistory = false;
+   private static final FrameVector2D zeroVector = new FrameVector2D();
 
    private final YoRegistry registry = new YoRegistry(getClass().getSimpleName());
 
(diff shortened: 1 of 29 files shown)


ihmc-interfaces/src/main/messages/ihmc_interfaces/toolbox_msgs/msg/KinematicsStreamingToolboxConfigurationMessage.msg
```diff
@@ -44,6 +44,10 @@ bool enable_chest_taskspace true
 # Default value is true.
 bool enable_pelvis_taskspace true
 
+# Whether center of mass should be controlled.
+# Default value is true.
```
LuigiPenco93: #isn't default false

ihmc-interfaces/src/main/messages/ros1/toolbox_msgs/msg/KinematicsStreamingToolboxConfigurationMessage.msg
```diff
@@ -52,6 +52,10 @@ bool enable_chest_taskspace
 # Field default value True
 bool enable_pelvis_taskspace
 
+# Whether center of mass should be controlled.
+# Default value is true.
```
LuigiPenco93: default false

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/BalanceManager.java
```diff
@@ -255,6 +263,7 @@ public BalanceManager(HighLevelHumanoidControllerToolbox controllerToolbox,
 
       WalkingMessageHandler walkingMessageHandler = controllerToolbox.getWalkingMessageHandler();
       momentumTrajectoryHandler = walkingMessageHandler == null ? null : walkingMessageHandler.getMomentumTrajectoryHandler();
+      blendDuration.set(2.5);
```
LuigiPenco93: it'd be good to have this as a controller parameter, unless it should always be 2.5

LuigiPenco93 (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/BalanceManager.java
stephenmcc: I switched it to use an already created parameter in PrecomputedICPPlanner

PR #618, opened 2025-01-26
KST Multi-Contact Integration

stephenmcc:
The controller currently publishes the hand load status. This has the KST lock the position of the hand so the loaded hand doesn't move. Subsequent PR will add integration with front-end VR

### DIFF
--- a/ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsToolboxModule/HumanoidKinematicsToolboxController.java
+++ b/ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsToolboxModule/HumanoidKinematicsToolboxController.java
@@ -10,14 +10,19 @@ import us.ihmc.avatar.initialSetup.RobotInitialSetup;
 import us.ihmc.commonWalkingControlModules.controllerCore.WholeBodyControllerCore;
 import us.ihmc.commonWalkingControlModules.controllerCore.command.feedbackController.CenterOfMassFeedbackControlCommand;
 import us.ihmc.commonWalkingControlModules.controllerCore.command.feedbackController.FeedbackControlCommandBuffer;
+import us.ihmc.commonWalkingControlModules.controllerCore.command.feedbackController.PointFeedbackControlCommand;
 import us.ihmc.commonWalkingControlModules.controllerCore.command.feedbackController.SpatialFeedbackControlCommand;
 import us.ihmc.commonWalkingControlModules.controllerCore.command.inverseKinematics.InverseKinematicsCommandBuffer;
 import us.ihmc.commonWalkingControlModules.controllerCore.command.inverseKinematics.JointLimitReductionCommand;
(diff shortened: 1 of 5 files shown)


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsToolboxModule/HumanoidKinematicsToolboxController.java
```diff
@@ -499,16 +513,40 @@ public void updateInternal()
...
+         if (multiContactRegionCalculator.hasSolvedWholeRegion())
+         {
+            // Update the region by querying the stability in another direction
+            multiContactRegionCalculator.performUpdateForNextVertex();
+         }
+         else
+         {
+            // Query new direction until initial region has been constructed
+            multiContactRegionCalculator.performUpdateForNextVertex();
+         }
```
rjgriffin42: it looks like both blocks are doing the same thing. Is that intentional?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsToolboxModule/HumanoidKinematicsToolboxController.java
```diff
@@ -582,68 +620,40 @@ protected void updateCoMPositionToHold()
...
-         feedbackControlCommand.setInverseKinematics(initialFootPoses.get(robotSide), KinematicsToolboxHelper.zeroVector6D);
-      }
-   }
-
-   /**
-    * Creates and sets up the feedback control commands for holding the active contact points in place.
-    *
-    * @param bufferToPack
-    */
-   private void addHoldSupportRigidBodyCommands(FeedbackControlCommandBuffer bufferToPack)
```
rjgriffin42, line 618: From the way this was implemented before, it seems like you could have an arbitrary rigid body in contact, whereas now you can only have the hands or feet. Is that the functionality you want?

rjgriffin42 (approved):

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsToolboxModule/HumanoidKinematicsToolboxController.java
stephenmcc: Oops no it isn't. Will fix that

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsToolboxModule/HumanoidKinematicsToolboxController.java
stephenmcc: Yeah, I've settled on just feet/hand contacts for now. The Valkyrie paper was all about fully generic contacts, this method is from that. But my current approach is just support planar feet and point hand contacts and revisit if need be. For controller/RDX integration it's a lot cleaner and simpler.

PR #619, opened 2025-01-27
RDX Multi-Contact Integration

stephenmcc:
Adding front-end support for new KST feature of having the arms be load-bearing. Adds an enum matching NadiaArmConfiguration, in which a hand action either does a gripper action, loads the arm or nothing (if no arm is present). This also swaps to the correct ROS message for the sake gripper. Integration with perception hasn't been done yet, this currently has a fixed contact normal.


calvertdw (approved):
Love to see it! As long as it works and is run through a good test on the robot and probably make sure the simulation still works, I'm good.

PR #624, opened 2025-01-30
Stability margin gradient calculator and IK objective

stephenmcc:
Adds an objective to the IK for improving the stability margin. This complements the existing hard constraint for stability, where the robot can't put the CoM within 4cm of the feasible region. This adds an objective cost between 4cm -> 12cm where the margin is penalized, the weight is increased when close to the 4cm threshold.

Disabled by default, there's a field in HumanoidKinematicsToolboxConfigurationMessage to enable it.


stephenmcc:
IKObjectiveDemo

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/staticEquilibrium/StabilityMarginRegionCalculator.java
```diff
@@ -164,10 +170,33 @@ public CenterOfMassStabilityMarginRegionCalculator(String namePrefix, double rob
       }
    }
 
-   public void clear()
+   public static StabilityMarginRegionCalculator createForCoMStabilityMargin(String prefix, double robotMass, YoRegistry registry, YoGraphicsListRegistry graphicsListRegistry)
```
PotatoPeeler3000, line 173: What was the choice of this being a static method rather then a constructor call to create this object? We return the new `StabilityMarginRegionCalculator`? What are the benefits of this? Is it so that we can create the `CenterOfMassStabilityMarginOptimizationModule()`?

ihmc-common-walking-control-modules/src/test/java/us/ihmc/commonWalkingControlModules/staticEquilibrium/CenterOfPressureOptimizationModuleTest.java
```diff
@@ -0,0 +1,54 @@
...
+import us.ihmc.euclid.referenceFrame.FramePose3D;
+import us.ihmc.euclid.referenceFrame.ReferenceFrame;
+import us.ihmc.euclid.tuple3D.Point3D;
+import us.ihmc.euclid.tuple3D.Vector3D;
+import us.ihmc.euclid.tuple4D.Quaternion;
+import us.ihmc.graphicsDescription.yoGraphics.YoGraphicsListRegistry;
+import us.ihmc.robotics.referenceFrames.PoseReferenceFrame;
+import us.ihmc.yoVariables.registry.YoRegistry;
+
+public class CenterOfPressureOptimizationModuleTest
```
PotatoPeeler3000, line 20: Might be good to add some Java Doc at the top explaing what this does

PotatoPeeler3000 (approved):
Just from looking at this from a code perspective it seems reasonable. Some body with more knowledge on the topic should probably approve this too ha.

Beomyeong (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/staticEquilibrium/StabilityMarginRegionCalculator.java
stephenmcc: The top level class `StabilityMarginRegionCalculator` is always the same, but it can be configured to use one of two solvers under-the-hood, `{CenterOfMass/CenterOfPressure}StabilityMarginOptimizationModule`. I didn't want to require the user to construct the solver and pass it in (which is still an option), so this is kind of a verbose helper method to get around that.

PR #626, opened 2025-01-30
Subscribe to status message about controller knowledge of drift offset in z and adjust height map accordingly

PotatoPeeler3000:
Explaination: The controller is publishing a `PlanOffsetMessage` in double support when walking where it says the foot landed at some pose and that's likely different then the expected/planned pose. So publish the different between what was planned and what really happened. We assume that to be some drift that we can subscribe to and account for.

- Added height map kernel to offset the global height map by the amount of drift coming from the controller.
___________________________________________________________________________________________________________________
Also moved the Snapping kernels to its own class so that the logic is seperate and the code is a bit more readable.
- Moved snapping kernels to its own class
- Pass down the controller footstep tracker from a higher level to have access to it in the perception pipeline.


ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
```diff
@@ -281,13 +287,38 @@ public void renderImGuiWidgets()
...
       }
-      ImGui.sameLine();
       stepsBeforeSafetyStop.render(0.0, 50.0);
       ImGui.checkbox("Use A* Footstep Planner", useAStarFootstepPlanner);
       ImGui.checkbox("Use Monte-Carlo Footstep Planner", useMonteCarloFootstepPlanner);
       ImGui.checkbox("Use Monte-Carlo Reference", useMonteCarloReference);
       ImGui.unindent();
       ImGui.separator();
+
+      if (ImGui.button("Fake Controller Drift"))
```
PotatoPeeler3000, line 303: Leaving this button in the UI so that in simulation I have a manual check to see if the controller drift offset kernel works.
This fakes sending the correct messages in order to trigger the controller drift kernel for the height map

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
```diff
@@ -460,6 +491,11 @@ public ContinuousPlannerSchedulingTask getContinuousPlannerSchedulingTask()
       return continuousPlannerSchedulingTask;
    }
 
+   public ControllerFootstepQueueMonitor getControllerFootstepQueueMonitor()
```
PotatoPeeler3000, line 500: Pass this up to then pass back down into the height map manager

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ControllerFootstepQueueMonitor.java
```diff
@@ -30,6 +39,8 @@ public ControllerFootstepQueueMonitor(ROS2Helper ros2Helper,
       this.continuousHikingLogger = continuousHikingLogger;
       ros2Helper.subscribeViaCallback(HumanoidControllerAPI.getTopic(FootstepQueueStatusMessage.class, simpleRobotName), this::footstepQueueStatusReceived);
       ros2Helper.subscribeViaCallback(HumanoidControllerAPI.getTopic(FootstepStatusMessage.class, simpleRobotName), this::footstepStatusReceived);
+      ros2Helper.subscribeViaCallback(getTopic(PlanOffsetStatus.class, simpleRobotName), this::acceptPlanOffsetStatus);
```
PotatoPeeler3000, line 42: Add a couple call backs to subscribe to the messages I need to get the drift offset data

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapDriftOffset.java
```diff
@@ -0,0 +1,57 @@
...
+import controller_msgs.msg.dds.PlanOffsetStatus;
+import us.ihmc.behaviors.activeMapping.ControllerFootstepQueueMonitor;
+import us.ihmc.euclid.tuple3D.Vector3D;
+
+/**
+ * This class is meant to keep track of the drift in Z that is being calculated in the controller on the robot.
+ * This class takes in a {@link ControllerFootstepQueueMonitor} in order to subscribe to the {@link PlanOffsetStatus}.
+ * It then keeps track of the current drift since walking has started.
+ */
+public class RapidHeightMapDriftOffset
```
PotatoPeeler3000, line 12: Class to handle the drift offset being sent from the controller

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapDriftOffset.java
```diff
@@ -0,0 +1,57 @@
...
+   {
+      this.controllerFootstepQueueMonitor = controllerFootstepQueueMonitor;
+   }
+
+   public void reset()
+   {
+      mostRecentPlanOffsetProcessed.setToZero();
+   }
+
+   public float getUpdateDriftOffset()
```
PotatoPeeler3000, line 32: This gets called in an update loop and returns either the float value for how much drift has occured, or NaN if no drift has occured.

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
```diff
@@ -37,17 +38,20 @@ public class RapidHeightMapManager
...
+   private RapidHeightMapDriftOffset rapidHeightMapDriftOffset;
+
+   private GpuMat deviceDepthImage;
+   private BytedecoImage heightMapBytedecoImage;
+
    public RapidHeightMapManager(ROS2PublishSubscribeAPI ros2,
                                 DRCRobotModel robotModel,
                                 ReferenceFrame leftFootSoleFrame,
                                 ReferenceFrame rightFootSoleFrame,
+                                ControllerFootstepQueueMonitor controllerFootstepQueueMonitor,
```
PotatoPeeler3000, line 54: Pass this guy down, created this at a higher level so the perception pipeline has access to it.

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
```diff
@@ -57,7 +61,8 @@ public RapidHeightMapManager(ROS2PublishSubscribeAPI ros2,
       if (runWithCUDA)
       {
          deviceDepthImage = new GpuMat(depthImageIntrinsics.getHeight(), depthImageIntrinsics.getWidth(), opencv_core.CV_16UC1);
-         rapidHeightMapExtractor = new RapidHeightMapExtractorCUDA(leftFootSoleFrame, rightFootSoleFrame, depthImageIntrinsics, deviceDepthImage, 1);
+         rapidHeightMapExtractor = new RapidHeightMapExtractorCUDA(leftFootSoleFrame, rightFootSoleFrame, deviceDepthImage, depthImageIntrinsics, 1);
+         rapidHeightMapDriftOffset = new RapidHeightMapDriftOffset(controllerFootstepQueueMonitor);
```
PotatoPeeler3000, line 65: If we are running CUDA we create this guy

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
```diff
@@ -123,6 +130,15 @@ public void update(Mat latestDepthImage, Instant imageAcquisitionTime, Reference
       cameraPose.setToZero(cameraFrame);
       cameraPose.changeFrame(ReferenceFrame.getWorldFrame());
 
+      if (runWithCUDA)
```
PotatoPeeler3000, line 133: Again, this guy is only created with CUDA so need to check here

ihmc-high-level-behaviors/src/test/java/us/ihmc/rdx/perception/RDXRapidHeightMapExtractorCUDADemo.java
```diff
@@ -94,7 +94,7 @@ public void create()
 
             heightMapImage = new GpuMat(intrinsics.getWidth(), intrinsics.getHeight(), opencv_core.CV_16UC1);
 
-            extractor = new RapidHeightMapExtractorCUDA(leftFootSoleFrame, rightFootSoleFrame, intrinsics, heightMapImage, 1);
+            extractor = new RapidHeightMapExtractorCUDA(leftFootSoleFrame, rightFootSoleFrame, heightMapImage, intrinsics, 1);
```
PotatoPeeler3000, line 97: Changed parameter order

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
```diff
@@ -33,43 +32,32 @@
...
    private final int mode; // 0 -> Ouster, 1 -> Realsense
-   public int sequenceNumber = 0;
-
    private final GpuMat inputDepthImage;
    private final GpuMat localHeightMapImage;
    private final GpuMat globalHeightMapImage;
    private final GpuMat terrainCostImage;
    private final GpuMat contactMapImage;
    private final GpuMat sensorCroppedHeightMapImage;
-   private final GpuMat steppabilityImage;
```
PotatoPeeler3000, line 53: The snapping stuff got moved to its own class

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
```diff
@@ -133,13 +124,7 @@ public RapidHeightMapExtractorCUDA(ReferenceFrame leftFootSoleFrame,
          contactMapImage = new GpuMat(globalCellsPerAxis, globalCellsPerAxis, opencv_core.CV_8UC1);
          sensorCroppedHeightMapImage = new GpuMat(heightMapParameters.getCropWindowSize(), heightMapParameters.getCropWindowSize(), opencv_core.CV_16UC1);
 
-         // Initialize matrices and images
```
PotatoPeeler3000, line 136: Yep different class now

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
```diff
@@ -211,11 +195,13 @@ public void reset()
          height = Math.min(footSoleFrames.get(RobotSide.LEFT).getTransformToWorldFrame().getTranslationZ(),
                            footSoleFrames.get(RobotSide.RIGHT).getTransformToWorldFrame().getTranslationZ()) - thicknessOfTheFoot;
       }
-      int offset = (int) ((height + heightMapParameters.getHeightOffset()) * heightMapParameters.getHeightScaleFactor());
+      resetOffset = (int) ((height + heightMapParameters.getHeightOffset()) * heightMapParameters.getHeightScaleFactor());
```
PotatoPeeler3000, line 203: More accurate

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
```diff
@@ -324,19 +303,7 @@ public void update(RigidBodyTransform sensorToWorldTransform, RigidBodyTransform
...
-      snappingKernel.withPointer(snapNormalXImage.data()).withLong(snapNormalXImage.step());
-      snappingKernel.withPointer(snapNormalYImage.data()).withLong(snapNormalYImage.step());
-      snappingKernel.withPointer(snapNormalZImage.data()).withLong(snapNormalZImage.step());
-      snappingKernel.withPointer(snappedAreaFractionImage.data()).withLong(snappedAreaFractionImage.step());
-      snappingKernel.withPointer(snappingParametersDevicePointer).withInt(heightMapParameters.getCropWindowSize());
-
-      snappingKernel.run(stream, snappingKernelGridDim, blockSize, 0);
-      error = cudaStreamSynchronize(stream);
-      CUDATools.checkCUDAError(error);
+      snappedFootstepsExtractor.update(globalHeightMapImage, sensorOrigin, globalCenterIndex, cropCenterIndex);
```
PotatoPeeler3000, line 311: All this logic for the snapping happens in the snapping extractor now. Better organized

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
```diff
@@ -346,55 +313,56 @@ public void update(RigidBodyTransform sensorToWorldTransform, RigidBodyTransform
       error = cudaStreamSynchronize(stream);
       CUDATools.checkCUDAError(error);
       terrainMapData.setHeightMap(finalCroppedHeightMap);
+   }
+
+   public void updateHeightOffset(float z)
```
PotatoPeeler3000, line 324: This handles the drift offset calculation. It adjusts the global height map accordingly

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorInterface.java
```diff
@@ -15,6 +15,10 @@ public interface RapidHeightMapExtractorInterface
 
    void update(RigidBodyTransform sensorToWorldTransform, RigidBodyTransform sensorToGroundTransform, RigidBodyTransform groundToWorldTransform);
 
+   default void updateHeightOffset(float z)
```
PotatoPeeler3000, line 18: Needed to add this to the interface

PotatoPeeler3000: Kinda gross

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/SnappingHeightMapExtractor.java
```diff
@@ -0,0 +1,199 @@
...
+import us.ihmc.perception.steppableRegions.SteppableRegionCalculatorParameters;
+
+import java.net.URL;
+
+import static org.bytedeco.cuda.global.cudart.cudaFree;
+import static org.bytedeco.cuda.global.cudart.cudaStreamSynchronize;
+import static us.ihmc.perception.gpuHeightMap.RapidHeightMapExtractorCUDA.BLOCK_SIZE_XY;
+import static us.ihmc.perception.gpuHeightMap.RapidHeightMapExtractorCUDA.heightMapParameters;
+
+public class SnappingHeightMapExtractor
```
PotatoPeeler3000, line 26: The snapping stuff now lives in its own class

PotatoPeeler3000: It works the same as the OpenCL kernels so the logic is all the same. That being said I think there are bugs in the implementation that I should look at. Could also add JavaDoc to this class.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -484,350 +460,22 @@ extern "C" __global__ void croppingKernel(unsigned short *inputMap, size_t pitch
...
 extern "C"
-__global__ void computeSnappedValuesKernel(unsigned short *globalMap, size_t pitchGlobal,
-                                           unsigned short *steppabilityMap, size_t pitchSteppability,
-                                           unsigned short *snapHeightMap, size_t pitchSnapHeight,
-                                           unsigned short *snapNormalXMap, size_t pitchSnapNormalX,
-                                           unsigned short *snapNormalYMap, size_t pitchSnapNormalY,
-                                           unsigned short *snapNormalZMap, size_t pitchSnapNormalZ,
-                                           unsigned short *snappedAreaFractionMap, size_t pitchSnappedAreaFraction,
-                                           float *params, int croppedMapXY)
+__global__ void planOffsetKernel(unsigned short * matrixToModify, size_t pitchMatrixToModify,
```
PotatoPeeler3000, line 464: Kernel for the drift offset.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -484,350 +460,22 @@ extern "C" __global__ void croppingKernel(unsigned short *inputMap, size_t pitch
...
-            }
-        }
-    }
+    unsigned short *skipRow = (unsigned short *)((char *)matrixValuesToSkip + indexY * pitchMatrixValuesToSkip);
+    if (skipRow[indexX] != resetOffset)
+        return;
 
-        unsigned short *steppableConnectionsElement = (unsigned short *)((char *)steppableConnectionsMap + key.x * pitchSteppableConnectionsMap) + key.y;
-        *steppableConnectionsElement = static_cast<unsigned short>(boundaryConnectionsEncodedAsOnes);
+    unsigned short *matrixRow = (unsigned short *)((char *)matrixToModify + indexY * pitchMatrixToModify);
```
PotatoPeeler3000, line 479: The diff is bad on github, I'd check this out locally

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
```diff
@@ -91,8 +98,7 @@ public class RDXContinuousHikingPanel extends RDXPanel implements RenderableProv
    private ROS2StoredPropertySetGroup clientStoredPropertySets;
    private boolean runSubscriberOnly = false;
    private boolean publishAndSubscribe;
-   private final ControllerFootstepQueueMonitor controllerFootstepQueueMonitor;
-   private final ContinuousHikingLogger continuousHikingLogger;
+   private double value = -0.1;
```
rjgriffin42: This should be renamed.

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
```diff
@@ -281,13 +287,38 @@ public void renderImGuiWidgets()
...
+
+         // Simulate that the controller has drifted by some z value
+         PlanOffsetStatus planOffsetStatus = new PlanOffsetStatus();
+         Vector3D planOffset = new Vector3D(0, 0, value);
+         planOffsetStatus.getOffsetVector().set(planOffset);
+         LogTools.info("Plan Offset Status: " + planOffsetStatus.getOffsetVector());
+         ros2Helper.publish(getTopic(PlanOffsetStatus.class, "Nadia"), planOffsetStatus);
+
+         // The amount of drift that we want to simulation and adjust for if we do this over and over
+         if (value < 5)
```
rjgriffin42: 5.0 to avoid casting.

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ControllerFootstepQueueMonitor.java
```diff
@@ -56,9 +67,28 @@ private void footstepStatusReceived(FootstepStatusMessage footstepStatusMessage)
...
 
+      footstepStarted = footstepStatusMessage.getFootstepStatus() == FootstepStatusMessage.FOOTSTEP_STATUS_STARTED;
+
       this.footstepStatusMessage.set(footstepStatusMessage);
    }
 
+   private void acceptWalkingStatusMessage(WalkingStatusMessage message)
+   {
+      // Declared locally since this represents the absolute state which other threads can access
+      walkingStarted.set(false);
```
rjgriffin42: would this be better named as "isWalking"?

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ControllerFootstepQueueMonitor.java
```diff
@@ -68,4 +98,19 @@ public AtomicReference<FootstepStatusMessage> getFootstepStatusMessage()
...
+   }
+
+   public boolean isFootstepStarted()
+   {
+      return footstepStarted;
+   }
+
+   public boolean isWalkingStarted()
+   {
+      return walkingStarted.getAndSet(false);
```
rjgriffin42: you should rename this to `pollIsWalkingStarted()` (or, `pollIsWalking`). Changing the value like this after you've called is a poll, since if you call it twice in a row it will return different values.

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
```diff
@@ -114,17 +103,19 @@ public RapidHeightMapExtractorCUDA(ReferenceFrame leftFootSoleFrame,
       URL mathUtilsHeaderPath = getClass().getResource("MathUtils.cuh");
       URL kernelPath = getClass().getResource("RapidHeightMapExtractor.cu");
 
-      recomputeDerivedParameters();
       terrainMapData = new TerrainMapData(heightMapParameters.getCropWindowSize(), heightMapParameters.getCropWindowSize());
 
+      recomputeDerivedParameters();
```
rjgriffin42, line 113: why did this get moved?

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
```diff
@@ -346,55 +313,56 @@ public void update(RigidBodyTransform sensorToWorldTransform, RigidBodyTransform
...
-      terrainMapData.setSnapNormalYImage(cpuSnapNormalYMap);
-      terrainMapData.setSnapNormalZImage(cpuSnapNormalZMap);
-      terrainMapData.setSnappedAreaFractionImage(cpuSnappedAreaFractionMap);
+      emptyRegisterKernl.run(stream, registerKernelGridDim, blockSize, 0);
+      error = cudaStreamSynchronize(stream);
+      CUDATools.checkCUDAError(error);
 
-      terrainMapData.setSteppabilityImage(cpuSteppabilityMap);
-   }
+      // Run the snapping kernel
```
rjgriffin42: this comment seems inaccurate

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorInterface.java
rjgriffin42: I gotta ask, why didn't you add it to the interface, then?

Nick's reply- "I needed to add it to the interface so that when the extractor gets called in the manager. It doesn't matter which implementation is used it will still work"

rjgriffin42 (dismissed):
A few small change requests.

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
TomaszTB: Wouldn't it be better to default the fake drift value to 0.0?

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
```diff
@@ -281,13 +287,38 @@ public void renderImGuiWidgets()
...
+         {
+            value -= 0.1;
+         }
+         else
+         {
+            value += 0.1;
+         }
```
TomaszTB: What's the purpose of the condition here? Seems like if we start from 0 and subtract a bit each time, `value < 5` will always be true

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapDriftOffset.java
TomaszTB: I'd expect it to return 0.0 if no drift occurred

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapDriftOffset.java
```diff
@@ -0,0 +1,61 @@
...
+ * This class takes in a {@link ControllerFootstepQueueMonitor} in order to subscribe to the {@link PlanOffsetStatus}.
+ * It then keeps track of the current drift since walking has started.
+ */
+public class RapidHeightMapDriftOffset
+{
+   private final ControllerFootstepQueueMonitor controllerFootstepQueueMonitor;
+   private final Vector3D mostRecentPlanOffsetProcessed = new Vector3D();
+   private final Vector3D incrementalOffset = new Vector3D();
+
+   public RapidHeightMapDriftOffset(ControllerFootstepQueueMonitor controllerFootstepQueueMonitor)
```
TomaszTB, line 18: I'd say this class should instantiate its own `ControllerFootstepQueueMonitor` since it polls values from it. 

The `ControllerFootstepQueueMonitor` instance cannot be shared between objects since they'll get different results depending on which object polls values from it first. Seems dangerous to me that one `ControllerFootstepQueueMonitor` instance could be passed into multiple objects with this design.

(The alternative would be to remove the polling mechanisms from `ControllerFootstepQueueMonitor`)

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapDriftOffset.java
```diff
@@ -0,0 +1,61 @@
...
+
+      Vector3D latestPlanOffset = latestPlanOffsetMessage.getOffsetVector();
+
+      if (controllerFootstepQueueMonitor.isFootstepStarted() && latestPlanOffset != null)
+      {
+         incrementalOffset.set(latestPlanOffset);
+         // We subtract the drift that has already been accounted for to only adjust by the new drift
+         incrementalOffset.sub(mostRecentPlanOffsetProcessed);
+         // Our internal tracker needs to add the drift because that is the new value being accounted for, so add it to the running total
+         mostRecentPlanOffsetProcessed.add(incrementalOffset);
```
TomaszTB: Something about the logic here seems fishy to me. It seems like this line can be replaced with this: 
```suggestion
         mostRecentPlanOffsetProcessed.set(latestPlanOffset);
```
And that makes me question setting the `mostRecentPlanOffsetProcessed` to zero above. Am I misunderstanding something?

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
```diff
@@ -33,43 +32,32 @@
...
-   private final CUDAKernel snappingKernel;
-
+   private final CUDAKernel planOffsetKernel;
+   private final CUDAKernel emptyRegisterKernl;
    private final float[] worldToGroundTransformArray = new float[16];
    private final float[] groundToWorldTransformArray = new float[16];
    private final float[] groundToSensorTransformArray = new float[16];
    private final float[] sensorToGroundTransformArray = new float[16];
-
    private final FloatPointer groundToSensorTransformHostPointer;
```
TomaszTB: Can we get the newlines separating blocks of fields back? My brain struggles looking at huge walls of fields like this. The newlines help group related fields.

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
TomaszTB, line 324: Come to think of it, these methods should be annotated with `@Override` (doesn't need to be part of this PR though)

TomaszTB (changes requested):
Got some initial thoughts. I wasn't able to get through it all today, so I'll leave more comments tomorrow. Thought these might be good to share right away though

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapDriftOffset.java
PotatoPeeler3000: That's fair, returning NaN allows the user to check where its used if the kernel should be run or not. I supposed you could return 0.0 and then check that if (!drift == 0.0) then run kernel.
I found that NaN seemed easier to understand as its literally not a number ha

PotatoPeeler3000: So I do this:
```
         float driftOffsetInZ = rapidHeightMapDriftOffset.getUpdateDriftOffset();
         if (!Float.isNaN(driftOffsetInZ))
         {
            rapidHeightMapExtractor.updateHeightOffset(driftOffsetInZ);
         }
```

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapDriftOffset.java
PotatoPeeler3000: Ahhh ok ya its because I didn't explain what's happening on the controller. And its a lot of text to explain but I'll try.

PotatoPeeler3000: So when the robot starts walking. The drift tracker from the controller is reset to 0. Meaning anything down the chain needs to account for that. So when the robot starts walking. I reset to zero drift.
What the controller is sending is the total amount of drift. So locally we track what was the last drift we accounted for. And subtract that from the new total. To get the new drift that we should adjust for.
So first, set the incrementalDrift to be the latest total amount of drift.
Second, subtract the drift that was used last time. Will be zero if its the first time.
Third, not that we have our incremental change to account for. Add that to the current total drift tracker (this is the `mostRecentPlanOffsetProcessed`).
And the checks from the `ControllerFootstepQueueMonitor` are to deal with how and when the drift is sent from the controller.

PotatoPeeler3000: PS. I'm doing these comments from IntelliJ so sorry if the formatting sucks lol

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorInterface.java
PotatoPeeler3000: I needed to add it to the interface so that when the extractor gets called in the manager. It doesn't matter which implementation is used it will still work

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
PotatoPeeler3000: I want to fake drift. So this button will fake -0.1 (10 centimeters) of drift. This way in the UI I can see the height map change its value by that much. Now the robot in simulation isn't drifting. So the heightmap in the UI will move down by 10 centimeters. At least it should if the drift kernel is working. This is a quick check that I made while testing and thought it would be a good thing to leave.

PotatoPeeler3000: Agreed, `simulatedDriftInMeters`

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
PotatoPeeler3000: Done

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ControllerFootstepQueueMonitor.java
PotatoPeeler3000: Yep, assuming the WalkingStatusMessage is only published when the state changes. Then ya I agree. Its not always easy for me to tell at what rate or if its a state change that publishes the status messages.

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ControllerFootstepQueueMonitor.java
PotatoPeeler3000: Good to know, changed to `pollIsWalking`

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
PotatoPeeler3000: Organization, no real reason

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
PotatoPeeler3000: Yep copy and paste error

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
PotatoPeeler3000: Ya your right about that. In this case its not that useful. It's like this cause its a test case that isn't perfect lol. When I was testing I'd change the `value` (now called `simulatedDriftInMeters`) to be like 5.1 and see how that effects the height map.
However, this can be -5.0 to then start occilating back and forth at that value. Actually -1.0 is better.

PotatoPeeler3000: And made it if (number > -1.0) then subtract, otherwise add

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapDriftOffset.java
PotatoPeeler3000: As long as the user is careful with the `ControllerFootstepQueueMonitor` is should be ok. The fields in there are mostly Atomic. Meaning they can be read from and written to in a single operation. This ensures thread safety as its not possible to have race conditions. Look up Atomic's in Java for more details on this.

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
PotatoPeeler3000: I would but every time I run the formatter it removes them again. And you know I'm a bit fan of letting the formatter decide things lol

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
PotatoPeeler3000: Don't worry, fixed!

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
TomaszTB: I don't know what formatter you're using, but mine doesn't do that. It removes two consecutive newlines, but even then it leaves one newline. I'd still really like the newlines back

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/SnappingHeightMapExtractor.java
```diff
@@ -0,0 +1,199 @@
...
+   {
+      int error;
+
+      // Populate parameters buffer for the snapping kernel
+      float[] snappingParametersArray = populateSnappingParametersArray(sensorOrigin, globalCenterIndex, cropCenterIndex);
+      snappingParametersHostPointer.put(snappingParametersArray);
+
+      // Handle memory allocation and copy values to the GPU
+      CUDATools.mallocAsync(snappingParametersDevicePointer, snappingParametersArray.length, stream);
+      error = cudaStreamSynchronize(stream);
```
TomaszTB: You don't need the synchronize here

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/SnappingHeightMapExtractor.java
```diff
@@ -0,0 +1,199 @@
...
+      steppabilityImage.close();
+      snapHeightImage.close();
+      snapNormalXImage.close();
+      snapNormalYImage.close();
+      snapNormalZImage.close();
+      snappedAreaFractionImage.close();
+
+      snappingParametersHostPointer.close();
+      snappingParametersDevicePointer.close();
+      cudaFree(snappingParametersDevicePointer);
```
TomaszTB: Might be worth checking the returned error value

TomaszTB (approved):
Couple minor things I'd like changed, but I still approve as is.

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/SnappingHeightMapExtractor.java
PotatoPeeler3000: Good call!

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/SnappingHeightMapExtractor.java
PotatoPeeler3000: Alright

PR #634, opened 2025-02-03
[Feature] CUDA Point Cloud Extractor

TomaszTB:
CUDA implementation of the OpenCL point cloud extractor!


ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
```diff
@@ -0,0 +1,165 @@
...
+   private final FloatPointer gpuPointCloudPointer = new FloatPointer();
+   private final IntPointer pointCloudSize = new IntPointer();
+
+   public CUDAPointCloudExtractor()
+   {
+      // Get URLs to the CUDA files
+      URL pointCloudExtractionURL = getClass().getResource("PointCloudExtraction.cu");
+      URL utilsURL = getClass().getResource("Utils.cu");
+      URL perceptionUtilsURL = getClass().getResource("PerceptionUtils.cu");
+      URL mathUtilsURL = getClass().getResource("/us/ihmc/perception/gpuHeightMap/MathUtils.cuh");
```
TomaszTB: This line is bad. We should decide where we want to keep the utils files before merging this PR

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/MathUtils.cuh
```diff
@@ -1,3 +1,6 @@
+#ifndef MATH_UTILS
+#define MATH_UTILS
```
TomaszTB, line 2: I think we should start adding these checks to utils files. I was considering including this file in `PerceptionUtils.cu` and this would've been good to have. 

(I didn't end up including it, but I kept these checks since they're good in general)

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/MathUtils.cuh
```diff
@@ -10,6 +13,13 @@ __device__ float dot2D(const float2 a, const float2 b)
     return a.x * b.x + a.y * b.y;
 }
 
+__device__ float3 transformPoint3D(float3 point, const float* transform)
```
TomaszTB, line 16: Same as the function below, but takes in a transformation matrix instead of rotation matrix rows and translation. Also added a test for it in `MathUtilsTest`. 

Should I name this "transformPoint3D**32**" since it operates on floats? I thought it was a bit redundant, but happy to change it.

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
ds58: Looks like MathUtils.cuh should move in with the rest of them, as it is the outlier.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/MathUtils.cuh
ds58: I agree

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/MathUtils.cuh
ds58: Yeah I think it's redundant. That would only be necessary if we have a transformPoint3D which acts on doubles.

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
```diff
@@ -0,0 +1,165 @@
...
+
+      // Copy the point cloud data from GPU to CPU, then to a Java float[]
+      FloatPointer cpuPointCloudPointer = new FloatPointer(numberOfFloats);
+      float[] pointsArray = new float[(int) numberOfFloats];
+      CUDATools.memcpyAsync(cpuPointCloudPointer, gpuPointCloudPointer, numberOfFloats, stream);
+      CUDATools.checkCUDAError(cudaStreamSynchronize(stream));
+      cpuPointCloudPointer.get(pointsArray);
+
+      // Create a list of points from the float[]
+      List<Point3D32> result = new ArrayList<>(numberOfPoints);
```
ds58: Here maybe try a parallel stream? Seems like it could scale well

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
PotatoPeeler3000: True, I'd move the `MathUtils.cuh` into the same folder as the `Utils.cu`

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
```diff
@@ -0,0 +1,165 @@
...
+      catch (Exception e)
+      {
+         throw new RuntimeException(e);
+      }
+
+      // Get a stream
+      stream = CUDAStreamManager.getStream();
+
+      // Allocate fixed size page-locked memory (on host)
+      CUDATools.checkCUDAError(cudaMallocHost(transformPointer, 16L * transformPointer.sizeof())); // 16 floats for transform matrix
```
PotatoPeeler3000: I think doing this on one line is more confusing then its worth. Now we don't know what the `cudaMallocHost` call returns. And its harder to read

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
```diff
@@ -0,0 +1,165 @@
...
+    *       Thus, the number of points in the output will not necessarily equal the number of pixels within the depth image.
+    */
+   public List<Point3D32> extractPointCloud(RawImage depthImage)
+   {
+      if (depthImage.get() == null)
+         return new ArrayList<>();
+
+      // Ensure enough GPU memory is allocated
+      long maxPointCloudFloats = 3L * depthImage.getWidth() * depthImage.getHeight();
+      if (gpuPointCloudPointer.isNull() || gpuPointCloudPointer.limit() < maxPointCloudFloats)
```
PotatoPeeler3000, line 83: This check seems confusing

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
```diff
@@ -0,0 +1,165 @@
...
+            .withFloat(depthImage.getPrincipalPointX())
+            .withFloat(depthImage.getPrincipalPointY())
+            .withFloat(depthImage.getDepthDiscretization())
+            .withPointer(transformPointer)
+            .withPointer(gpuPointCloudPointer)
+            .withPointer(pointCloudSize)
+            .run(stream, gridSize, blockSize, 0);
+
+      // Synchronize the stream to ensure we can read the point cloud size
+      CUDATools.checkCUDAError(cudaStreamSynchronize(stream));
```
PotatoPeeler3000: Saving this one line of code is not worth it. Its now harder to read and we don't know what `CUDATools.checkCUDAError()` takes in as a parameter.

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
```diff
@@ -0,0 +1,165 @@
...
+import us.ihmc.log.LogTools;
+import us.ihmc.perception.RawImage;
+
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.bytedeco.cuda.global.cudart.*;
+
+public class CUDAPointCloudExtractor implements AutoCloseable
```
PotatoPeeler3000, line 19: Is there a reason to implement AutoCloseable? I think we might be doing it wrong

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
```diff
@@ -0,0 +1,165 @@
...
+      cpuPointCloudPointer.close();
+      depthImage.release();
+
+      return result;
+   }
+
+   @Override
+   public void close()
+   {
+      try
```
PotatoPeeler3000: Why would the context already be deleted? Is that something that needs to be checked for in all close methods?

ihmc-perception/src/main/resources/us/ihmc/perception/cuda/PointCloudExtraction.cu
```diff
@@ -0,0 +1,49 @@
...
+                                  float cy,
+                                  float depthDiscretization,
+                                  float* depthToWorldTransform,
+                                  float3* pointCloud,
+                                  int* pointCloudSize)
+{
+    int startX = Utils::getThreadCoordX();
+    int strideX = Utils::getStrideX();
+
+    int startY = Utils::getThreadCoordY();
```
PotatoPeeler3000, line 24: I'd argue this is also more work then its worth. For the user its nice to know the math going into things.

ihmc-perception/src/test/java/us/ihmc/perception/gpuHeightMap/MathUtilsTest.java
```diff
@@ -175,6 +177,52 @@ private float[] runTransformPoint3d32_2OnCUDA() throws Exception
       return resultFromGKernel;
    }
 
+   @Test
+   public void testTransformPoint2CUDA() throws Exception
```
PotatoPeeler3000, line 182: Could you add a comment on what this test does?

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
ds58: Made sense to me

TomaszTB: It's checking whether there's enough GPU memory allocated. 
We need to (re)allocate memory if: 
a) pointer is null (no memory allocated), or
b) limit is less than what we need (too little memory allocated)

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
TomaszTB: `AutoCloseable` lets us instantiate the class in a try-with-resources block. IntelliJ will also show a warning if it detects that an `AutoCloseable` class isn't closed when it goes out of scope. I like using it. 

I don't think there's much to do wrong with it. It's basically the same as having a `destroy` method, but with some extra utility. I'm confident we're using it the way it's meant to be used.

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAPointCloudExtractor.java
TomaszTB: Maybe this was a misconception I had. I tested it and it doesn't throw any exceptions like that.

ihmc-perception/src/main/resources/us/ihmc/perception/cuda/PointCloudExtraction.cu
TomaszTB: Hmm I was going for some abstraction so the user doesn't have to worry about the math (also because I forget the math all the time haha). 

Would it be alright if we keep it for now, and decide later whether we like these functions or not? I'd have to remove calls to them in a few places if we want to remove them. It'd be its own PR.

ds58: I think the abstraction here is appropriate and keeps the kernel cleaner

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
```diff
@@ -111,7 +111,7 @@ public RapidHeightMapExtractorCUDA(ReferenceFrame leftFootSoleFrame,
 
       // Load header and main file
       URL heightMapUtilsHeaderPath = getClass().getResource("HeightMapUtils.cuh");
-      URL mathUtilsHeaderPath = getClass().getResource("MathUtils.cuh");
+      URL mathUtilsHeaderPath = getClass().getResource("/us/ihmc/perception/cuda/MathUtils.cuh");
```
TomaszTB, line 108: Now I don't like this line :(

ds58 (approved):

PotatoPeeler3000 (approved):

PR #641, opened 2025-02-04
Prevent setting to null values when there are no changes to undo, whi…

calvertdw:
…ch caused crash.


ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/behaviorTree/BehaviorTreeNodeDefinition.java
```diff
@@ -121,8 +121,11 @@ public void setOnDiskFields()
 
    public void undoAllNontopologicalChanges()
    {
-      name.setValue(onDiskName);
-      notes.setValue(onDiskNotes);
+      if (isUndoAvailable())
+      {
+         name.setValue(onDiskName);
```
calvertdw, line 126: The primary crash was name getting set to null, which native crashed an ImGui function.

LuigiPenco93 (approved):

PR #646, opened 2025-02-05
Feature/flying points filter

ihmc-rosie:


ihmc-perception/src/main/java/us/ihmc/perception/filters/CUDAFlyingPointsFilter.java
```diff
@@ -0,0 +1,68 @@
...
+   }
+
+   public void destroy()
+   {
+      flyingPointFilterCUDAProgram.close();
+      flyingPointFilterKernel.close();
+      CUDAStreamManager.releaseStream(stream);
+   }
+
+   public Mat applyFilter(Mat inputImage)
```
ds58: Seems like it would be better for this to take a GpuMat and return a GpuMat. You may not want this to internally upload and download the image, in case you need to do additional processing.

In this case with the height map, I would expect the image to stay on the GPU from the height map extractor, and then pass it into this filter. Then, at a later point, you download it.

ds58: Should you have the option to pass in a CUDA stream into `applyFilter`? Seems useful

LaughingCommit: The function is called from the manager with a `Mat` while another stream ran. Filter applications with other streams caused crashes, so creating/destroying the stream in-process and passing a `Mat` is more reliable.

ds58: >Filter applications with other streams caused crashes

Why does it cause a crash? I think that's something that needs to be investigated

If this filter has to perform more memory copies than necessary within the pipeline, it's gonna be slow

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
```diff
@@ -0,0 +1,109 @@
...
+    {
+        *outputPixel = static_cast<unsigned short>(depthValue); // Keep original depth value
+    }
+}
+
+
+
+
+
+
```
ds58: Can you remove the extra newlines at the end of this file?

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
ds58, line 1: This should be moved to `ihmc-perception/src/main/resources/us/ihmc/perception/cuda`

ds58 (dismissed):

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
ds58: Also, I would rename to just `FlyingPointsFilter.cu`, appending `Kernel` is redundant (and not exactly correct, this file would represent a program not necessarily a kernel) and we don't do it for the other ones

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapUpdateThread.java
```diff
@@ -81,6 +81,10 @@ protected void runTask()
       catch (InterruptedException ignored)
       {
       }
+      catch (Exception e)
+      {
+         e.getStackTrace();
```
ds58: `LogTools.error(e)`

ihmc-perception/src/main/java/us/ihmc/perception/filters/CUDAFlyingPointsFilter.java
TomaszTB: I started writing `async` methods (like the CUDA async functions) in these classes which takes in a stream, and the non-async method just calls the async method with an internally stored stream... e.g.:
```
public CUstream_st stream = CUDAStreamManager.getStream();

public void methodAsync(CUstream_st stream)
{
   /* do stuff using passed in stream */
}

public void method()
{
   methodAsync(stream);
   cudaStreamSynchronize(stream);
}
```

TomaszTB: I agree with Dex here. I don't think this class should use `Mat`s at all. Everything that it does should be done with `GpuMat`s

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
```diff
@@ -0,0 +1,109 @@
+extern "C"
```
TomaszTB: I don't think you need this `extern "C"` here

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Computes the median value from a given window of pixel values.
+ * @param window The size of the kernel to check its neighbors.
+ * @param size Number of elements in the window.
+ * @return The median value.
+ */
```
TomaszTB: Does Javadoc work in `.cu` files???

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
```diff
@@ -46,23 +47,26 @@ public class RapidHeightMapManager
...
                                 ReferenceFrame leftFootSoleFrame,
                                 ReferenceFrame rightFootSoleFrame,
                                 ControllerFootstepQueueMonitor controllerFootstepQueueMonitor,
                                 CameraIntrinsics depthImageIntrinsics,
-                                boolean runWithCUDA)
+                                boolean runWithCUDA) throws Exception
    {
       this.ros2 = ros2;
       this.runWithCUDA = runWithCUDA;
+      flyingPointsFilter = new CUDAFlyingPointsFilter();
```
PotatoPeeler3000: This requires people to have CUDA on there machine for this to work, you wann throw this in the `runWithCUDA` section.

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
```diff
@@ -143,6 +147,11 @@ public void update(Mat latestDepthImage, Instant imageAcquisitionTime, Reference
 
       Mat croppedHeightMapImage = rapidHeightMapExtractor.getTerrainMapData().getHeightMap();
 
+      if (RapidHeightMapExtractorCUDA.getHeightMapParameters().getFlyingPointsFilter())
```
PotatoPeeler3000: You'll wanna add a part of this if statment that check if `runWithCUDA` is true. That way if someone uses OpenCL it doesn't cause any problems

ihmc-perception/src/test/java/us/ihmc/perception/filters/CUDAFlyingPointsFilterTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.bytedeco.opencv.opencv_core.GpuMat;
+import org.bytedeco.opencv.opencv_core.Mat;
+import org.junit.jupiter.api.Test;
+
+public class CUDAFlyingPointsFilterTest
+{
+   private CUDAFlyingPointsFilter flyingPointsFilter;
+   private GpuMat inputGpuMat = new GpuMat();
+   private GpuMat outputGpuMat;
+   private Mat outputMat = new Mat();
```
TomaszTB: I'd say these should be local to the test method, since they should be closed after use

ihmc-perception/src/test/java/us/ihmc/perception/filters/CUDAFlyingPointsFilterTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+   /**
+    * Utility method to print the matrix values for debugging.
+    *
+    * @param cpuResult The matrix to be printed.
+    */
+   private static void printResult(Mat cpuResult)
```
TomaszTB: Might wanna use the `PerceptionDebugTools#printMat()` method

ihmc-perception/src/main/java/us/ihmc/perception/filters/CUDAFlyingPointsFilter.java
```diff
@@ -0,0 +1,61 @@
...
+   {
+      GpuMat deviceInputImage = new GpuMat(inputImage.rows(), inputImage.cols(), opencv_core.CV_16UC1);
```
PotatoPeeler3000: You wanna add some formatting to this section? When I say that I mean to add a blank line between sections that make sense to you. For example all the "initialization" stuff is kind of a group. Then we add all the parameters to the filter. Then you run the filter, then you download it, then you close stuff. Throw some blank lines between these sections so its more readable.

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
PotatoPeeler3000: Move the kernel file to where the rest of the height map stuff lives.
image

ihmc-perception/src/test/java/us/ihmc/perception/filters/CUDAFlyingPointsFilterTest.java
```diff
@@ -0,0 +1,71 @@
...
+         }
+         System.out.println();
+      }
+   }
+
+   /**
+    * Test case for validating the CUDA flying points filter with a simple 3x3 matrix.
+    */
+   @Test
+   public void testSimpleMatrix() throws Exception
```
TomaszTB, line 16: Ideally this test would `assert` that the result is correct. Right now it only tests whether the code can run without throwing exceptions

ihmc-perception/src/test/java/us/ihmc/perception/filters/CUDAFlyingPointsFilterTest.java
```diff
@@ -0,0 +1,71 @@
...
+      printResult(inputMat);
+
+      // Uploading input matrix to GPU memory
+      inputGpuMat.upload(inputMat);
+
+      // Applying the CUDA filter
+      outputMat = flyingPointsFilter.applyFilter(inputMat);
+
+      // Downloading processed matrix back to CPU memory
+//      outputGpuMat.download(outputMat);
```
PotatoPeeler3000: Can this be removed? Also many of these comments are not necessary. You can tell it was make with chatgpt cause the ai likes to add a lot of comments lol. I'd only keep the ones that wouldn't be obvious when reading the code.

ihmc-perception/src/test/java/us/ihmc/perception/filters/CUDAFlyingPointsFilterTest.java
```diff
@@ -0,0 +1,71 @@
...
+         {
+            // Extracting and printing matrix values
+            System.out.print(cpuResult.row(i).col(j).data().getShort() + " ");
+         }
+         System.out.println();
+      }
+   }
+
+   /**
+    * Test case for validating the CUDA flying points filter with a simple 3x3 matrix.
```
PotatoPeeler3000: Test case for what? What's it testing? Is there supposed to be a flying point in the 3x3 matrix? Is that what its testing?
I'd explain this more to make it clear that we expect a flying point in the 3x3 matrix and we expect the result to not have that flying point.
And if the window size is 5, does this test even pass?

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
```diff
@@ -46,23 +47,26 @@ public class RapidHeightMapManager
...
       this.ros2 = ros2;
       this.runWithCUDA = runWithCUDA;
+      flyingPointsFilter = new CUDAFlyingPointsFilter();
 
       if (runWithCUDA)
       {
          deviceDepthImage = new GpuMat(depthImageIntrinsics.getHeight(), depthImageIntrinsics.getWidth(), opencv_core.CV_16UC1);
          rapidHeightMapExtractor = new RapidHeightMapExtractorCUDA(leftFootSoleFrame, rightFootSoleFrame, deviceDepthImage, depthImageIntrinsics, 1);
          rapidHeightMapDriftOffset = new RapidHeightMapDriftOffset(controllerFootstepQueueMonitor);
+
```
PotatoPeeler3000: Can remove this new line/

TomaszTB (dismissed):
It needs some improvements before we merge it. Let me know if you have trouble with something (the CUDA stream stuff?)

PotatoPeeler3000 (commented):
Go through the comments on the code, you can kinda guess that parts were written with chatgpt cause of all the comments. I'd only keep the comments that are necessary to help understand the code. Anything that can be understood by reading the code. I'd remove the comment for that section.

TomaszTB:
> Move the kernel file to where the rest of the height map stuff lives.

I wouldn't put it with height map stuff since it can be applied to any depth image. It's not height map specific

ihmc-perception/src/main/java/us/ihmc/perception/filters/CUDAFlyingPointsFilter.java
```diff
@@ -0,0 +1,63 @@
...
+
+   public GpuMat applyFilter(GpuMat deviceInputImage)
+   {
+      GpuMat deviceOutputImage = new GpuMat(deviceInputImage.rows(), deviceInputImage.cols(), opencv_core.CV_16UC1);
+
+      int gridSizeX = (deviceInputImage.cols() + BLOCK_SIZE_XY - 1) / BLOCK_SIZE_XY;
+      int gridSizeY = (deviceInputImage.rows() + BLOCK_SIZE_XY - 1) / BLOCK_SIZE_XY;
+
+      dim3 blockSize = new dim3(BLOCK_SIZE_XY, BLOCK_SIZE_XY, 1);
+      dim3 gridSize = new dim3(gridSizeX, gridSizeY, 1);
```
rjgriffin42, line 52: does creating these on the fly cause latency? Or is it actually better for the GPU?

ihmc-perception/src/main/java/us/ihmc/perception/filters/CUDAFlyingPointsFilter.java
```diff
@@ -0,0 +1,63 @@
...
+import us.ihmc.perception.cuda.CUDATools;
+
+public class CUDAFlyingPointsFilter
+{
+   private static final int BLOCK_SIZE_XY = 16;
+   private final CUDAKernel flyingPointFilterKernel;
+   private final CUDAProgram flyingPointFilterCUDAProgram;
+   private final CUstream_st stream;
+
+   public CUDAFlyingPointsFilter() throws Exception
```
rjgriffin42, line 23: Is there a more specific type of exception that this throws, so that when you catch it at a higher level, it's more useful?

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
```diff
@@ -138,11 +139,24 @@ public void update(Mat latestDepthImage, Instant imageAcquisitionTime, Reference
...
       }
-      
+
       rapidHeightMapExtractor.update(sensorToWorld, sensorToGround, groundToWorld);
 
       Mat croppedHeightMapImage = rapidHeightMapExtractor.getTerrainMapData().getHeightMap();
 
+      if (runWithCUDA && RapidHeightMapExtractorCUDA.getHeightMapParameters().getFlyingPointsFilter())
+      {
+         CUDAFlyingPointsFilter flyingPointsFilter = new CUDAFlyingPointsFilter();
```
rjgriffin42: doesn't it seem that you'd be better off creating this at construction, rather than every time you call update?

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
```diff
@@ -0,0 +1,100 @@
+
+// Computes the median value from a given window of pixel values.
+// window = The size of the kernel to check its neighbors.
+
+__device__ float computeMedian(float* window, int size)
+{
+    // Sort the window using Bubble Sort (can be replaced with a faster sort)
```
rjgriffin42: This isn't super descriptive. How is it sorting the data? ascending order?

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
```diff
@@ -0,0 +1,100 @@
...
+                float temp = window[j];
+                window[j] = window[j + 1];
+                window[j + 1] = temp;
+            }
+        }
+    }
+    // Return median based on even or odd size
+    if (size % 2 == 0)
+    {
+        return (window[size / 2 - 1] + window[size / 2]) / 2.0f;
```
rjgriffin42: replace with `* 0.5f`, as it's faster than `/ 2.0f`

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
```diff
@@ -0,0 +1,100 @@
...
+{
+    float sum = 0.0f;
+    for (int i = 0; i < size; ++i)
+    {
+        sum += window[i];
+    }
+    return sum / size;
+}
+
+__device__ float computeStdDev(float* window, int size, float mean)
```
rjgriffin42: This is specifically the _population_ standard deviation, rather than the _sample_ stand deviation

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
```diff
@@ -0,0 +1,100 @@
...
+{
+    int xIndex = blockIdx.x * blockDim.x + threadIdx.x;
+    int yIndex = blockIdx.y * blockDim.y + threadIdx.y;
+
+    if (xIndex >= cols || yIndex >= rows) return;
+
+    unsigned short *inputPixel = (unsigned short *)((char *)in + yIndex * pitchIn) + xIndex;
+    unsigned short depthValue = *inputPixel;
+
+    const int windowSize = 5;
```
rjgriffin42: where did 5 come from?

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
```diff
@@ -0,0 +1,100 @@
...
+    const int halfWindow = windowSize / 2;
+    float window[windowSize * windowSize];
+    int count = 0;
+
+    // Collect neighboring pixel values into the window
+    for (int dx = -halfWindow; dx <= halfWindow; ++dx)
+    {
+        for (int dy = -halfWindow; dy <= halfWindow; ++dy)
+        {
+            int nx = xIndex + dx;
```
rjgriffin42: This can be pushed into the outer-loop, and you can push the if condition there, too. So
```
int nx = xIndex + dx;

if (nx < 0 || nx >= cols)
   continue;
 ```
That avoids a bunch of uneccessary computation.

File: ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu [hunk 23]
```diff
@@ -0,0 +1,100 @@
...
+                                   unsigned short *out, size_t pitchOut,
+                                   int rows, int cols)
+{
+    int xIndex = blockIdx.x * blockDim.x + threadIdx.x;
+    int yIndex = blockIdx.y * blockDim.y + threadIdx.y;
+
+    if (xIndex >= cols || yIndex >= rows) return;
+
+    unsigned short *inputPixel = (unsigned short *)((char *)in + yIndex * pitchIn) + xIndex;
+    unsigned short depthValue = *inputPixel;
```
rjgriffin42: What is the point of this being two separate lines?

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    float median = computeMedian(window, count);
+    float mean = computeMean(window, count);
+    float stdDev = computeStdDev(window, count, mean);
+
+    unsigned short *outputPixel = (unsigned short *)((char *)out + yIndex * pitchOut) + xIndex;
+
+    // If the depth value deviates too much from the median, mark it as a flying point
+    if (fabsf(depthValue - median) >= 0.02f * median || fabsf(depthValue - median) > 1.5f * stdDev)
```
rjgriffin42: I'd pull the `fabsf(depthValue - median)` into its own local variable, since you're using it twice.

or you can say `fabsf(depthValue - median) > min(0.02f * median, 1.5f  * stdDev)` and do a single comparator.

PotatoPeeler3000 (approved):

ihmc-perception/src/main/java/us/ihmc/perception/filters/CUDAFlyingPointsFilter.java
TomaszTB: The `blockSize` could be created during construction, but `gridSize` needs to be created using the input image size, so it's easier to create it when you get the input image. 

The `dim3` object is just 3 integers, so I think it'd be a tiny optimization to create them upon construction

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapManager.java
TomaszTB: Also needs to be destroy after being created

ihmc-perception/src/main/java/us/ihmc/perception/filters/CUDAFlyingPointsFilter.java
TomaszTB: That exception comes from the `CUDAProgram`, which gets it from `CUDATools`, and that just throws an `Exception` right now. I can make a separate PR for throwing a more specific exception

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
LaughingCommit: The two lines separate the calculation of the pixel's memory address from accessing its value. The first line computes the address of the pixel, and the second line dereferences that address to retrieve the depth value.

ihmc-perception/src/main/resources/CUDA/FlyingPointFilterKernel.cu
LaughingCommit: The population standard deviation is more appropriate here because we're considering the entire dataset (all pixels in the window), not just a sample. In this case, we want to calculate the variation in depth values across all neighboring pixels, so using the population formula (dividing by `size` instead of `size - 1`) gives a more accurate representation of the overall spread of values.

PotatoPeeler3000 (approved):

ds58 (approved):

PR #654, opened 2025-02-10
Fix RDX catching unchecked exceptions

calvertdw:


ihmc-high-level-behaviors/src/test/java/us/ihmc/rdx/RDXUIDemo.java
```diff
@@ -22,6 +23,9 @@
 
 public class RDXUIDemo
 {
+   /** Currently relying on this hack to be removed when fixed. */
+   private static final Mat matToLoadBytedecoNatives = new Mat();
```
ds58: `matToLoadBytedecoNatives` -> `matToLoadJavaCPPNatives`

PR #667, opened 2025-02-13
Needed to be checking the absolute value of the distance, fixed that issue

PotatoPeeler3000:


ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/graphSearch/stepChecking/HeightMapCliffAvoider.java
```diff
@@ -130,7 +130,7 @@ public boolean isStepValid(DiscreteFootstep candidateStep, DiscreteFootstep stan
          Point2DReadOnly polygonVertex = scaledFootPolygon.getVertex(pointIdx);
          double zBestFitPlane = bestFitPlane.getZOnPlane(polygonVertex.getX(), polygonVertex.getY());
          double zHeightMap = heightMapData.getHeightAt(polygonVertex.getX(), polygonVertex.getY());
-         double distanceFromBestFitPlane = zHeightMap - zBestFitPlane;
+         double distanceFromBestFitPlane = Math.abs(zHeightMap - zBestFitPlane);
```
PotatoPeeler3000, line 133: Small change but large bugfix, we need to check the absolute value distance from the plane. That way if its above or below by 3cm. Either way that distance is larger then the 0.5cm that is `parameters.getCliffHeightThreshold()` so it rejects the footstep. 
Previously it would accept the footstep if the distance was -3cm. Because -0.03 < 0.05, so it passes as a valid step. But in reality the distance is greater then 0.05.

PotatoPeeler3000 (commented):
This seems to pretty successfully fix the issue of stepping on cliffs. Making the foot placement substantially better

PotatoPeeler3000:
image

This green footstep used to be valid because at the front of the foot (the right side). The height map height is less then the best fit plane. So when taking those vertex's and asking if the distance is < 0.05 (the parameter value). It would return true cause the value used to be -0.03.

Now I can't get the foot to step there, so this seems to pretty effective fix the stepping on cliff issue

ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/graphSearch/stepChecking/HeightMapCliffAvoider.java
rjgriffin42: I just want to check, are you meaning to say it should be checking against 0.005? Because the absolute value of -0.03 is also less than 0.05. If that's not the case, I don't understand how any of this works.

PotatoPeeler3000: Lol ya I meant 0.005

ihmc-rosie (approved):

PR #680, opened 2025-02-19
[Bugfix] Fix CUDAKernel Usage of Stream for Timing + Minor Optimizations

TomaszTB:


ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
```diff
@@ -51,7 +51,12 @@ public CUDAKernel(String name, CUmod_st kernelModule) throws Exception
    public void enableKernelTimings(boolean enableKernelTimings)
    {
       this.enableKernelTimings = enableKernelTimings;
-      kernelTimings = new CUDAKernelTimings();
+      if (enableKernelTimings && kernelTimings == null)
+      {
+         kernelTimings = new CUDAKernelTimings();
+         cudaEventCreate(start);
+         cudaEventCreate(end);
+      }
```
TomaszTB, line 59: Only create objects once, if kernel timing is enabled

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
```diff
@@ -74,11 +79,7 @@ public void run(CUstream_st stream, dim3 gridSize, dim3 blockSize, int sharedMem
          parametersPointer.put(i, parameters.get(i));
 
       if (enableKernelTimings)
-      {
-         cudaEventCreate(start);
-         cudaEventCreate(end);
-         cudaEventRecord(start);
-      }
+         cudaEventRecord(start, stream);
```
TomaszTB, line 82: Gotta pass in the stream to the `cudaEventRecord` function. Otherwise it records events on the default stream (so we weren't really recording the kernel's execution time, but the execution times of things that ran on the default stream during the kernel's execution)

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
```diff
@@ -159,6 +160,12 @@ public void close()
    {
       clearParameters();
       kernelFunction.close();
+
+      if (kernelTimings != null)
+      {
+         cudaEventDestroy(start);
+         cudaEventDestroy(end);
+      }
```
TomaszTB, line 168: Should destroy the events

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
```diff
@@ -169,10 +176,10 @@ private class CUDAKernelTimings
    {
       private static final int MAX_ENTRIES = 250;
       private final LinkedList<Float> executionTimes = new LinkedList<>();
+      private final float[] milliseconds = new float[1];
```
TomaszTB, line 179: Made this array a field (super minor optimization)

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
PotatoPeeler3000: This seems to make things to complicated. Why do we need to check if `kernelTimings` is null?
And we are calling `enableKernelTimings()` to probably start recording stuff. So it seems likely that `enableKernelTimings` will be true?

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
PotatoPeeler3000: I ran the kernels and got similar results for the timings as before? That makes me think we are doing something wrong somewhere with the stream.

PotatoPeeler3000 (approved):

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
TomaszTB: I just used the null check instead of having a boolean field like `kernelTimingsInitialized`. It's just to see whether the objects related to the timing stuff have been created. 

On develop all the places that call `enableKernelTimings` pass in `false` since we don't want the timing stuff running when using the robot. Seems to me that passing in `false` will be the default, and we'll only pass in `true` when manually testing stuff.

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAKernel.java
TomaszTB: Well, we are sharing streams across different places so the results are still not gonna be fully accurate. We could have a discussion about getting rid of the `CUDAStreamManager` to ensure accurate timings from this stuff. 

I'm guessing the results look similar due to the `cudaEventRecord` functions being called at identical intervals. But the documentation definitely suggests that we should pass in the stream on which the kernel will run (https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html#group__CUDART__EVENT_1gf4fcb74343aa689f4159791967868446).

ds58 (approved):

PR #690, opened 2025-02-21
Feature/behavior-python-module

LuigiPenco93:
Python module to communicate with ai2r node and do some reasoning.
Supporting:
- triggering of specific behavior among those loaded
- info about succeded or failed behavior
- control of parameters of go to behavior
- control of parameters of hand pose action


ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/ai2r/AI2RNodeExecutor.java
```diff
@@ -96,29 +138,41 @@ public void update()
       {
          statusMessage.getRobotMidFeetUnderPelvisPoseInWorld().set(syncedRobot.getFramePoseReadOnly(HumanoidReferenceFrames::getMidFeetUnderPelvisFrame));
 
+         // ------------ Scene --------------
```
calvertdw: I'd suggest making these comments normal or using methods instead to organizer, as we do elsewhere in the code

calvertdw (approved):
Awesome!

PR #693, opened 2025-02-21
Scene graph demo improvements

ds58:
- Fixes a somewhat common crashing bug related to the ZED SVO recorder panel
- Adds a utility for ensuring you have a GPU of some minimum spec (In the RDXSceneGraphDemo, this was problematic because ZED Neural mode takes so much GPU it is super laggy on lower end spec or laptops). This utility could also be useful other places.

### DIFF
--- a/ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
+++ b/ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
@@ -65,19 +65,24 @@ public class CUDATools
       return getCUDADeviceCount() > 0;
    }
 
-   public static String getFirstDeviceName()
+   /**
+    * Get the name of a CUDA device
+    * @param device the device number
+    * @return the name of the device
+    */
+   public static String getDeviceName(int device)
    {
-      try (IntPointer device = new IntPointer(1);
+      try (IntPointer devicePointer = new IntPointer(1);
            cudaDeviceProp deviceProperties = new cudaDeviceProp())
       {
-         cudaGetDevice(device);
-         cudaGetDeviceProperties(deviceProperties, device.get());
+         devicePointer.put(device);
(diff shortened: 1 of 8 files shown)


ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
```diff
@@ -63,6 +65,61 @@ public static boolean hasCUDADevice()
       return getCUDADeviceCount() > 0;
    }
 
+   public static String getFirstDeviceName()
```
TomaszTB: Why "first" device name?

ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
```diff
@@ -155,6 +157,11 @@ protected boolean startSensor()
       return true;
    }
 
+   public void disableNeuralMode(boolean disableNeuralMode)
```
TomaszTB: Maybe the depth mode should be a constructor parameter? That way we can pass in whatever depth mode we want. It would allow us to use different modes on a case-by-case basis

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDATools.java
ds58: You're right. I changed it to `getDeviceName(int device)`

ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
ds58: That's a good idea. Much better.

TomaszTB (approved):

calvertdw (approved):

PR #696, opened 2025-02-27
Removed one of the height map graphics and only allow for the RDXHeightMapRender in the UI

PotatoPeeler3000:


ihmc-high-level-behaviors/src/test/java/us/ihmc/rdx/perception/RDXSteppableRegionCalculatorDemo.java
```diff
@@ -88,7 +88,7 @@ public void create()
 
             ros2Node.createSubscription2(PerceptionAPI.HEIGHT_MAP_OUTPUT, message ->
             {
-               heightMapVisualizer.acceptHeightMapMessage(message);
+               heightMapVisualizer.setupForImageMessage(ros2Helper);
```
TomaszTB: I think this call should happen outside of the subscription, no? Otherwise you'll create a new subscription each time this subscription receives a message

PotatoPeeler3000: Correct lol I moved it

TomaszTB (approved):
Seems fine to me, although I'm not very familiar with this code

PR #698, opened 2025-02-27
Manually Freeing CUDA Memory

TomaszTB:
You were creating GpuMats without closing them. Fortunately the JavaCPP deallocator thread closed them properly, but I have trust issues with that thread because it has failed on me in the past. This PR does the memory management without relying on the JavaCPP deallocator thread. 

Before: see how the GPU memory usage fluctuates a bunch
Before

After: GPU memory usage is stable
After

This is running the NadiaContinuousHikingKinSim, btw


PotatoPeeler3000 (approved):
Thank you!

PR #719, opened 2025-03-12
Feature/fst footstep constraint

stefanfasano:
Addition of elliptical kinematic reachability constraint for manual and VR placed footsteps


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/desiredFootStep/StepPositionLimiter.java
stefanfasano, line 1: This class already existed and has been tested in ihmc-internal. I am just moving it (and its corresponding unit test) to open-robotics. Let me know if there is another project/package within open-robotics that would be a better home for this class

LuigiPenco93 (approved):

Beomyeong (approved):

PR #741, opened 2025-03-28
[Feature] Improved RDXSensorSimulator

TomaszTB:
Adds noise and makes the render call faster by delaying the post-processing of data until `get*Image` is called. 

Noise increases as the distance increases. 

NoiseSimulation


calvertdw (approved):
Sick :ok_hand:

ds58 (approved):

PR #745, opened 2025-04-02
Need to reset the drift offset when a new footstep plan is sent to the robot, otherwise we will over compensate.

PotatoPeeler3000:
image
The drift offset tracker on the controller gets reset to zero when the `currentNumberOfSteps` becomes zero. But that can happen when the robot is still walking cause the tracker decrements when the footstep is started, not when it ends.
By not resetting the tracker till the robot is not in walking state, we don't have to worry about keeping track of when the `currentNumberOfSteps` is empty. We only have 2 things to track:
- is the robot walking
- did we send a footstep data list message with OVERRIDE


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/messageHandlers/WalkingMessageHandler.java
```diff
@@ -203,9 +203,8 @@ public void handleFootstepDataListCommand(FootstepDataListCommand command)
                clearFlamingoCommands();
                break;
             case QUEUE:
-               // TODO review the use of this. 
-               boolean checkForInconsistencies = !upcomingFootsteps.isEmpty() || currentNumberOfFootsteps.getIntegerValue() > 0;
-               if (checkForInconsistencies)
+               // Check for inconsistencies between the previous command options and the current command options
+               if (isWalking.getBooleanValue())
```
PotatoPeeler3000, line 207: The `upcomingFootsteps` becomes empty when the last step starts executing. So if you send a `FootstepDataListMessage` when the last step is in swing, and the message has the `ExecutionMode.QUEUE`. When that message is received on the controller, the `upcomingFootsteps` is empty, so we jump to the else statement and reset the `planOffsetVector` (setToZero()).

To me it doesn't really make sense to reset the `planOffsetVector` which the robot is not in stance. The else statement could have the last step still executing.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/messageHandlers/WalkingMessageHandler.java
```diff
@@ -243,7 +242,7 @@ else if (upcomingFootsteps.isEmpty())
                }
                else
                {
-                  // We don't have any steps in the queue, so it's effectively the same thing as overriding.
+                  // We aren't walking, we are in stance, so it's effectively the same thing as overriding.
```
PotatoPeeler3000, line 245: Now we are guaranteed to not be `isWalking` if this code is executed. So we should reset the plan offset stuff.

ihmc-humanoid-robotics/src/main/java/us/ihmc/humanoidRobotics/communication/ControllerFootstepQueueMonitor.java
```diff
@@ -38,6 +41,18 @@ public ControllerFootstepQueueMonitor(ROS2Node ros2Node, String simpleRobotName)
...
+      ros2Node.createSubscription2(getTopic(FootstepDataListMessage.class, simpleRobotName), this::interceptFootstepDataListMessage);
+   }
+
+   private void interceptFootstepDataListMessage(FootstepDataListMessage footstepDataListMessage)
+   {
+      if (footstepDataListMessage.getOffsetFootstepsHeightWithExecutionError())
+      {
+         if (footstepDataListMessage.getQueueingProperties().getExecutionMode() == QueueableMessage.EXECUTION_MODE_OVERRIDE)
+         {
+            receivedNewFootstepPlanWithOverride.set(true);
```
PotatoPeeler3000, line 53: We want to know if we have send a new footstep plan with `OVERRIDE` cause we need to reset our tracker for the height map drift. This is because the controller resets its internal vector if it receives an `OVERRIDE` FootstepDataListMessage.

PR #750, opened 2025-04-04
Fix bug disabling concurrency.

calvertdw:


ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/behaviorTree/BehaviorTreeRootNodeExecutor.java
```diff
@@ -291,7 +291,7 @@ private boolean shouldExecuteNextLeaf()
       }
       else
       {
-         return !currentlyExecutingLeaves.isEmpty();
+         return currentlyExecutingLeaves.isEmpty();
```
calvertdw, line 294: I broke this on Dec 10th
image

calvertdw: The logic is supposed to be "if no actions/leaves are executing".

calvertdw: currentlyExecutingLeaves being empty is already equivalent to !anyActionExecuting, so the '!' was an error

LuigiPenco93 (approved):

PR #759, opened 2025-04-10
Update parameters and tune for the GPU height map

PotatoPeeler3000:


ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
```diff
@@ -31,7 +31,7 @@
 
 public class RapidHeightMapExtractorCUDA
 {
-   private static final boolean PRINT_TIMING_FOR_KERNELS = false;
+   private static final boolean PRINT_TIMING_FOR_KERNELS = true;
```
TomaszTB, line 34: This might drown out more important warnings or errors, no? Are you sure we want this for open house?

TomaszTB (approved):
No idea what these parameters do, but I'll trust the height map expert

Beomyeong (approved):

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
ihmc-rosie: ^^^Good catch, don't need that to be true, changing

PR #761, opened 2025-04-10
Add ROS 2 and UI for cooldown and is calibrating features.

calvertdw:


ihmc-interfaces/src/main/messages/ros1/controller_msgs/msg/SakeHandDesiredCommandMessage.msg
```diff
@@ -15,6 +15,15 @@ bool request_calibration
 # Request to reset the gripper error state after overheating
 bool request_reset_errors
 
+# Set to true to enable automatic cooldown
+bool enable_automatic_cooldown
+
+# Set to false to disable automatic cooldown
+bool disable_automatic_cooldown
```
TomaszTB: I think it'd be less confusing if it was one boolean

TomaszTB (commented):
Looks pretty good. Having two separate booleans to enable/disable the auto cooldown is confusing though. I think it can be done with just one

ihmc-interfaces/src/main/messages/ros1/controller_msgs/msg/SakeHandDesiredCommandMessage.msg
calvertdw: The reason for this is that you don't know the initial state of the other side when you are commanding, so a boolean is not enough, this makes it so it is a request for either value, but two falses means we aren't sure of the state, but aren't trying to change it either.

calvertdw: Imagine you open a 2nd UI and the status is not received yet, but the user tries to command something else, that's where having this be false might change the state of the robot side unintentionally.

TomaszTB: That makes sense. I think the comment on `disable_automatic_cooldown` is wrong though. Cool down would be disabled if it was set to `true`, right?

TomaszTB (approved):

PR #766, opened 2025-04-14
Initialize the logger if the Continuous Hiking subscriber is enabled, set things to work from the Autonomy Process for the Simulation UI

PotatoPeeler3000:


ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/perception/RDXContinuousHikingPanel.java
```diff
@@ -220,6 +218,8 @@ public void startContinuousPlannerSchedulingTask(ActiveMappingParameterToolBox a
       runSubscriberOnly = true;
       this.clientStoredPropertySets = clientStoredPropertySets;
 
+      // We only want to start the logger if we create the subscriber for Continuous Hiking
+      ContinuousHikingLogger continuousHikingLogger = new ContinuousHikingLogger();
```
PotatoPeeler3000, line 222: This is what it should have been originally, we log the subscriber side of things, so if that isn't started, we don't need to setup a logger

Beomyeong (approved):

PR #772, opened 2025-04-18
Bugfix/revert kst zero com

LuigiPenco93:
had to revert this commit related to multi-contatct on develop, since I can't figure out the bug that introduced: setting the com of KST always to zero in world.
Also added back the check box for the ghost robot visualization

### DIFF
--- a/ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/vr/RDXVRKinematicsStreamingMode.java
+++ b/ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/vr/RDXVRKinematicsStreamingMode.java
@@ -94,6 +94,7 @@ public class RDXVRKinematicsStreamingMode
    private final RetargetingParameters retargetingParameters;
    private final DRCRobotModel robotModel;
    private RDXMultiBodyGraphic ghostRobotGraphic;
+   private final ImBoolean ghostPreviewEnabled = new ImBoolean(true);
    private FullHumanoidRobotModel ghostFullRobotModel;
    private OneDoFJointBasics[] ghostOneDoFJointsExcludingHands;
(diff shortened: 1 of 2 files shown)


ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/vr/RDXVRKinematicsStreamingMode.java
```diff
@@ -799,6 +800,10 @@ public void renderImGuiWidgets()
       {
          setEnabled(false);
       }
+      if (ImGui.checkbox(labels.get("Ghost preview"), ghostPreviewEnabled))
+      {
+         visualizeIKPreviewGraphic(ghostPreviewEnabled.get());
+      }
```
calvertdw, line 726: I think it would be better to just hide the hands. Otherwise, how do you know the IK is not freaking out when starting to stream?

calvertdw: Or have it only disappear when streaming is active

LuigiPenco93: added left B button for hiding, unhiding graphics

calvertdw: Sweet. Can you add that and the reset IK to the list of shortcuts in the menu? And any others that may be missing. This would have saved me lots of time a couple weeks ago

calvertdw (approved):
I approve just the RDX part. I'd get Stephen's review for the other part

PR #779, opened 2025-04-24
Cleanup/switch to gpu mats

PotatoPeeler3000:


ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
```diff
@@ -287,6 +280,12 @@ public void update(GpuMat latestDepthImageGPU,
       croppingKernel.run(stream, croppingKernelGridDim, blockSize, 0);
       error = cudaStreamSynchronize(stream);
       CUDATools.checkCUDAError(error);
+
+      // All that memory we allocated on the GPU, need to free that up now
+      cudaFree(groundToSensorTransformDevicePointer);
+      cudaFree(sensorToGroundTransformDevicePointer);
+      cudaFree(worldToGroundTransformDevicePointer);
+      cudaFree(parametersDevicePointer);
```
TomaszTB: Bonus points if you make these `cudaFreeAsync` so the CPU doesn't have to wait for the memory to be freed before continuing

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapManager.java
TomaszTB, line 1: Now that you only use the `update()` method that takes in a `RawImage`, how about changing the `update()` method that takes in a `Mat` to use a `GpuMat`? It'll save an upload to the GPU.

TomaszTB (approved):
Awesome changes. Love it.

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractorCUDA.java
PotatoPeeler3000: Yep good call, and I free'd more memory in the other update method (there shouldn't be two methods for update but that's currently what we got)

PR #782, opened 2025-04-28
[Bugfix] testROS2SRTStreamerAndSubscriber

TomaszTB:


ihmc-perception/src/test/java/us/ihmc/perception/streaming/SRTStreamerSubscriberTest.java
```diff
@@ -311,13 +311,13 @@ public void testROS2SRTStreamerAndSubscriber() throws InterruptedException
...
 
-      Thread sendThread = ThreadTools.startAThread(() ->
+      Thread sendThread = ThreadTools.startAsDaemon(() ->
       {
          // Send 2 seconds of video
          for (int i = 0; i < 2 * FPS; ++i)
          {
             throttler.waitAndRun();
-            streamer.sendFrame(rawImage);
+            streamer.sendFrame(RawImage.createWithBGRImage(sampleImage, cameraIntrinsics, testTransform, Instant.now(), 0L));
```
TomaszTB, line 320: Using the same image over and over was causing the timestamp problem that I added the check for. I guess the test was passing even if the timestamps didn't increase, but since I added the early return it started failing.

ihmc-perception/src/test/java/us/ihmc/perception/streaming/SRTStreamerSubscriberTest.java
```diff
@@ -332,7 +332,7 @@ public void testROS2SRTStreamerAndSubscriber() throws InterruptedException
       // Ensure we can receive an image
       assertTrue(subscriberHasReceivedFrame.get());
 
-      sendThread.join();
+      sendThread.join(3000);
```
TomaszTB, line 335: Hopefully this prevents the test from hanging indefinitely in the future

ds58 (approved):

PR #794, opened 2025-05-01
Split the autonomy height map thread and the continuous hiking height map thread

PotatoPeeler3000:


ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapManager.java
```diff
@@ -96,23 +96,41 @@ public RapidHeightMapManager(ROS2Node ros2Node,
...
+      hostCroppedHeightMap.close();
       depthImageCopy.release();
    }
 
-   public void update(Mat latestDepthImage,
-                      Instant acquisitionTime,
-                      CameraIntrinsics depthIntrinsicsCopy,
-                      ReferenceFrame cameraFrame,
-                      ReferenceFrame cameraZUpFrame) throws Exception
+   private void updateInternal(Mat latestDepthImage,
```
PotatoPeeler3000, line 129: Create a seperate update method that doesn't publish

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapManager.java
```diff
@@ -253,5 +258,6 @@ public void destroy()
       rapidHeightMapExtractor.destroy();
       snappedFootstepsExtractor.destroy();
       flyingPointsFilter.destroy();
+      compressedCroppedHeightMapPointer.close();
```
PotatoPeeler3000, line 261: This wasn't being closed

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapThread.java
TomaszTB, line 1: It's like 95% identical code to the autonomy thread. Peak code duplication :100: 

Now if you want to make a change to the height map thread you gotta do it in two places, there are two separate (though nearly identical) code paths to test and maintain, and this class doesn't enable anything that wasn't possible with the original. I'm really not convinced this is a positive change

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapManager.java
TomaszTB, line 129: Wait, the input depth image is still a CPU `Mat`? Should use a `GpuMat` from the `RawImage`

TomaszTB (commented):
I like the other changes but I really don't see the separate thread as a positive change

TomaszTB:
I'll only approve if @ihmc-rosie also approves

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapThread.java
PotatoPeeler3000: This is all true, I'm not convinced this is a positive change either, I'd like to try it and see how it goes.

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapManager.java
PotatoPeeler3000: Yep, this happen's in one of the following PR's, I would change it here but I don't wanna deal with the merge conflicts

PR #803, opened 2025-05-08
[Bugfix] MYWA

TomaszTB:
image

Fixed YOLO scene node publishing, lowered detection manager confidence threshold


ds58 (approved):

calvertdw (approved):

ihmc-perception/src/main/java/us/ihmc/perception/detections/DetectionManagerSettings.java
```diff
@@ -10,7 +10,7 @@ public class DetectionManagerSettings extends StoredPropertySet
 
    public static final DoubleStoredPropertyKey poseFilterAlpha = keys.addDoubleKey("Pose Filter Alpha", 0.2);
    public static final DoubleStoredPropertyKey maxMatchDistanceSquared = keys.addDoubleKey("Max Match Distance Squared", 1.0);
-   public static final DoubleStoredPropertyKey acceptanceAverageConfidence = keys.addDoubleKey("Acceptance Average Confidence", 0.75);
+   public static final DoubleStoredPropertyKey acceptanceAverageConfidence = keys.addDoubleKey("Acceptance Average Confidence", 0.25);
```
TomaszTB, line 13: I lowered this a ridiculous amount because the confidence on the mustard bottle was low. Maybe we should try to use a different bottle and increase this number? That could prevent bad detections from getting into the scene graph

PR #823, opened 2025-05-22
This is what the height map should be

PotatoPeeler3000:
Improvements: 
-
In step 1:
- We can compute the actual variance of the data, as well as the number of points to fuse in
- We should consider a of variance for the data based on distance from the camera, as well as the “speed” of that point relative to the origin. This would account for translational and rotational speed.
- We can compute the local normal of the data, and use that to filter out “flying point” cells
- Stop using a “guess” as to the initial height when computing the depth map pixel

In step 2,:
- Instead of using a very large, fixed global map, we should use a smaller height map, and translate it every time we get a new local map update
- Whenever the global map is translated, we should add variance to each of the cells
- Switch the fusing of the local map into the global map from just overwriting the value to using a kalman filter or weighted average


ihmc-perception/src/main/java/us/ihmc/perception/filters/DepthImageBodyCollisionFilter.java
```diff
@@ -157,6 +157,7 @@ public void close()
    {
       // Need to free this memory on the GPU to shut down correctly
       cudaFree(deviceCollidableGeometryPointer);
+      cudaFree(collidableGeometryPointer);
```
PotatoPeeler3000, line 160: Missed that guy!

ihmc-perception/src/main/java/us/ihmc/perception/tools/PerceptionDebugTools.java
```diff
@@ -280,7 +280,7 @@ public static String matToString(String name, Mat image, int rowBegin, int colBe
             if (image.type() == opencv_core.CV_64FC1)
                matString.append("%.5f\t".formatted(image.ptr(i, j).getDouble()));
             if (image.type() == opencv_core.CV_32FC1)
-               matString.append("%.2f\t".formatted(image.ptr(i, j).getFloat()));
+               matString.append("%.6f\t".formatted(image.ptr(i, j).getFloat()));
```
PotatoPeeler3000, line 279: If the map is a bunch of floats, lets see more of those decimals

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/FilteredRapidHeightMapExtractor.cu
PotatoPeeler3000, line 1: No longer needed, captured with the new kalman filter in the height map

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
PotatoPeeler3000, line 1: The diff for these are terrible, I'd view them in your IDE, its much better

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -0,0 +1,549 @@
...
+   private int centerIndexTerrain;
+   private int cellsPerAxisGlobal;
+   private int cellsPerAxisTerrain;
+
+   private final RigidBodyTransform previousSensorToWorld = new RigidBodyTransform();
+   private int previousCellX;
+   private int previousCellY;
+   private float resetOffset;
+
+   private FilteredVerticalSurfacesExtractor filteredVerticalSurfacesExtractor;
```
TomaszTB:
```suggestion
   private final FilteredVerticalSurfacesExtractor filteredVerticalSurfacesExtractor;
```

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -0,0 +1,549 @@
...
+import us.ihmc.sensorProcessing.heightMap.HeightMapTools;
+
+import java.net.URL;
+
+import static org.bytedeco.cuda.global.cudart.*;
+
+public class RapidHeightMapExtractor
+{
+   private static final boolean PRINT_TIMING_FOR_KERNELS = false;
+   static final int BLOCK_SIZE_XY = 32;
```
TomaszTB:
```suggestion
   private static final int BLOCK_SIZE_XY = 32;
```

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -0,0 +1,549 @@
...
+         cudaFreeAsync(groundToSensorTransformDevicePointer, stream);
+         checkCUDAError();
+      }
+
+      // ---------- Run the translate kernel ---------
+      {
+         int currentCellX = (int) Math.round(sensorOrigin.getX32() / heightMapParameters.getCellSizeInMeters());
+         int currentCellY = (int) Math.round(sensorOrigin.getY32() / heightMapParameters.getCellSizeInMeters());
+
+         // This means we have moved more then 2cm. So each cell should shift to one of its neighboring cells
```
TomaszTB:
```suggestion
         // This means we have moved more than 2cm. So each cell should shift to one of its neighboring cells
```

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -0,0 +1,549 @@
...
+   private final float[] worldToGroundTransformArray = new float[16];
+   private final float[] groundToSensorTransformArray = new float[16];
+   private final float[] sensorToGroundTransformArray = new float[16];
+
+   private final FloatPointer groundToSensorTransformHostPointer;
+   private final FloatPointer groundToSensorTransformDevicePointer;
+   private final FloatPointer sensorToGroundTransformHostPointer;
+   private final FloatPointer sensorToGroundTransformDevicePointer;
+   private final FloatPointer zUpCameraToWorldAlignedGroundHostPointer;
+   private final FloatPointer ZUpCameraToWorldAlignedGroundDevicePointer;
```
TomaszTB: Don't apply this through GitHub, but start field name with lower case
```suggestion
   private final FloatPointer zUpCameraToWorldAlignedGroundDevicePointer;
```

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -0,0 +1,549 @@
...
+   }
+
+   public void updateHeightOffset(RigidBodyTransform groundToWorldTransform, float z, CameraIntrinsics cameraIntrinsics, double footHeight)
+   {
+      int error;
+
+      // Populate parameter buffers with the necessary values
+      float[] parametersArray = populateParameterArray(heightMapParameters, cameraIntrinsics, footHeight);
+      parametersHostPointer.put(parametersArray);
+      CUDATools.mallocAsync(parametersDevicePointer, parametersArray.length, stream);
```
TomaszTB, line 381: You call `malloc` here but I don't see a corresponding `free` call

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -0,0 +1,549 @@
...
+         // Invert translation-only transform
+         worldToGroundNoRotation.invert();
+
+         // This transformation only has rotation
+         RigidBodyTransform groundToWorldAlignedGround = new RigidBodyTransform(worldToGroundNoRotation);
+         groundToWorldAlignedGround.multiply(groundToWorldTransform);
+
+         groundToWorldAlignedGround.get(worldToGroundTransformArray);
+         zUpCameraToWorldAlignedGroundHostPointer.put(worldToGroundTransformArray);
+         CUDATools.mallocAsync(ZUpCameraToWorldAlignedGroundDevicePointer, worldToGroundTransformArray.length, stream);
```
TomaszTB: Same here (call `malloc` but no `free`)

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -0,0 +1,549 @@
...
+   private final GpuMat previousGlobalMeanMap;
+   private final GpuMat previousGlobalVarianceMap;
+   private final GpuMat terrainCroppedHeightMap;
+   private final GpuMat scaledHeightMap;
+   private final GpuMat emptyGlobalHeightMap;
+
+   private final CUDAKernel updateKernel;
+   private final CUDAKernel translateKernel;
+   private final CUDAKernel registerKernel;
+   private final CUDAKernel terrainCroppingKernel;
```
TomaszTB, line 57: I don't see this one getting closed :eyes:

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -124,79 +117,60 @@ __device__ float get_spatial_average(int xIndex, int yIndex, unsigned short *glo
...
-    // Thread indices for output image
-    int xIndex = blockIdx.x * blockDim.x + threadIdx.x; // Output dimension x (new cols)
-    int yIndex = blockIdx.y * blockDim.y + threadIdx.y; // Output dimension y (new rows)
+    // Transform cell center from previous Z-up to current Z-up
+    float3 cellCenterInGroundNoRotation = transformPoint3D32_2(
+        cellCenterInZUp,
+        make_float3(zUpCameraToWorldAlignedGround[0], zUpCameraToWorldAlignedGround[1], zUpCameraToWorldAlignedGround[2]),
+        make_float3(zUpCameraToWorldAlignedGround[4], zUpCameraToWorldAlignedGround[5], zUpCameraToWorldAlignedGround[6]),
+        make_float3(zUpCameraToWorldAlignedGround[8], zUpCameraToWorldAlignedGround[9], zUpCameraToWorldAlignedGround[10]),
+        make_float3(zUpCameraToWorldAlignedGround[3], zUpCameraToWorldAlignedGround[7], zUpCameraToWorldAlignedGround[11]));
```
TomaszTB: Can't you use `transformPoint3D` from `MathUtils.cuh`? Then you wouldn't have to do all the `make_float3` calls
(same thing on line 220 and 278)

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -124,79 +117,60 @@ __device__ float get_spatial_average(int xIndex, int yIndex, unsigned short *glo
...
 
-    // Ensure within bounds of the output image (1280 rows and 720 columns)
-    if (xIndex >= rows || yIndex >= cols)
-    {
-        return;
-    }
 
-    unsigned short *inRow = (unsigned short *)((char *)in + (yIndex * pitchIn));
-    unsigned short depthValue = *(inRow + xIndex);
+    int2 newCellIndex = coordinate_to_indices(
```
TomaszTB, line 138: Seems like you could just return here. I don't see the purpose of the lines of code after this function call

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -207,25 +181,34 @@ extern "C" __global__ void heightMapUpdateKernel(unsigned short *in, size_t pitc
...
     // Compute grid cell center in Z-Up frame
     float2 xyCoords = indices_to_coordinate(make_int2(xIndex, yIndex),
                                             make_float2(0.0f, 0.0f),
-                                            params[LOCAL_CELL_SIZE],
+                                            params[CELL_SIZE],
                                             params[LOCAL_CENTER_INDEX]);
 
-    cellCenterInZUp.x = xyCoords.x + params[GRID_OFFSET_X];
+    cellCenterInZUp.x = xyCoords.x + params[HALF_LOCAL_WIDTH_IM_METERS];
     cellCenterInZUp.y = xyCoords.y;
```
TomaszTB: Just out of curiosity: Why is `HALF_LOCAL_WIDTH_IM_METERS` added to X but not to Y?

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -2,48 +2,41 @@
...
-#define FAST_SEARCH_SIZE 36
-#define GROUND_HEIGHT 37
-
+#define MODE 4
+#define DEPTH_CX 5
+#define DEPTH_CY 6
+#define DEPTH_FX 7
+#define DEPTH_FY 8
+#define GLOBAL_CENTER_INDEX 9
+#define HALF_LOCAL_WIDTH_IM_METERS 10
```
TomaszTB: IM
```suggestion
#define HALF_LOCAL_WIDTH_IN_METERS 10
```

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -207,25 +181,34 @@ extern "C" __global__ void heightMapUpdateKernel(unsigned short *in, size_t pitc
...
-                                            params[LOCAL_CELL_SIZE],
+                                            params[CELL_SIZE],
                                             params[LOCAL_CENTER_INDEX]);
 
-    cellCenterInZUp.x = xyCoords.x + params[GRID_OFFSET_X];
+    cellCenterInZUp.x = xyCoords.x + params[HALF_LOCAL_WIDTH_IM_METERS];
     cellCenterInZUp.y = xyCoords.y;
 
-    float halfCellWidth = params[LOCAL_CELL_SIZE] / 2.0f;
+    float halfCellWidth = params[CELL_SIZE] / 2.0f;
```
TomaszTB, line 202: Also, what's the difference between this `halfCellWidth` and `params[HALF_LOCAL_WIDTH_IM_METERS]`?

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -471,6 +549,6 @@ __global__ void planOffsetKernel(unsigned short * matrixToModify, size_t pitchMa
     if (abs( (int) skipRow[indexX] - resetOffset) >= 2)
```
TomaszTB: Just a weird space here
```suggestion
    if (abs((int) skipRow[indexX] - resetOffset) >= 2)
```

TomaszTB (changes requested):
Very nice! Just a few things I noticed

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
PotatoPeeler3000: Excellent question, I've added a comment. But it shifts the zUp cell into the area that has depth data. If we don't do that. We assume that the center of the local grid is at the same center as the global map, which isn't true. Its a smaller map. And if we shift it forward in X we can search that entire local map and expect depth data.

Something like that, I've actually had problems playing with this as it seems to break easily.

```
    // Shift forward in X so that this cell appears in front of the robot, meaning it likely has depth data
    cellCenterInZUp.x = xyCoords.x + params[HALF_LOCAL_WIDTH_IM_METERS];
    cellCenterInZUp.y = xyCoords.y;

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
PotatoPeeler3000: The `halfCellWidth` is 0.02 cm / 2 = 0.01 cm. The `params[HALF_LOCAL_WIDTH_IN_METERS]` is a parameter that can get changed. The local map at the moment is like 2 meters. (2.0). So that parameter is 2.0 / 2. But it could be 4.0 / 2. Just depends on that parameter.

TomaszTB (approved):
Nice :+1:

PR #841, opened 2025-06-05
Added the ZED to the height map by adding a queue to the ImageSensor

PotatoPeeler3000:
image

### DIFF
--- a/ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapThread.java
+++ b/ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapThread.java
@@ -27,8 +27,7 @@ import us.ihmc.ros2.ROS2Publisher;
 import us.ihmc.perception.heightMap.HeightMapData;
 import us.ihmc.perception.heightMap.HeightMapParameters;
 
-import java.util.Collection;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.BlockingQueue;
 
 public class RapidHeightMapThread extends RepeatingTaskThread
 {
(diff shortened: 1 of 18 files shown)


ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapThread.java
```diff
@@ -32,32 +37,23 @@ public class RapidHeightMapThread extends RepeatingTaskThread
...
    private final Object heightMapLock = new Object();
 
-   private final ImageSensor imageSensor;
-   private final ReferenceFrame cameraFrame;
-   private final ReferenceFrame zUpSensorFrame;
    private final HeightMapParameters heightMapParameters;
-   private final int depthImageKey;
    private final CUDACompressionTools cudaCompressionTools = new CUDACompressionTools();
    private final ROS2Publisher<ImageMessage> filteredDepthPublisher;
+   private final Collection<RawImage> rawImageCollection;
```
TomaszTB: I think this should be of type `BlockingQueue` so you don't have to cast it later

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapThread.java
```diff
@@ -32,32 +37,23 @@ public class RapidHeightMapThread extends RepeatingTaskThread
...
 
    public RapidHeightMapThread(ROS2Node ros2Node,
                                ROS2SyncedRobotModel syncedRobotModel,
                                RobotCollisionModel robotCollisionModel,
-                               ImageSensor imageSensor,
-                               int depthImageKey,
+                               Collection<RawImage> rawImageCollection,
```
TomaszTB: Same here. The whole thing will break if someone passes in a different type of collection, so there's no advantage to being generic about it

ihmc-interfaces/src/main/messages/ihmc_interfaces/perception_msgs/msg/HeightMapMessage.msg
TomaszTB, line 1: Are these sizes needed for the bigger height map? I hope it doesn't use too much RAM... It wouldn't be a problem with jros2 though

TomaszTB (changes requested):

ihmc-interfaces/src/main/messages/ihmc_interfaces/perception_msgs/msg/HeightMapMessage.msg
PotatoPeeler3000: Its still tiny relative to the ImageMessage

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/RapidHeightMapThread.java
PotatoPeeler3000: Yep good call

TomaszTB (approved):
Beautiful

PR #846, opened 2025-06-09
[Bugfix] Speedup YOLO

TomaszTB:
Comparing YOLO annotated image vs ZED left color image visualizers
Before:
YOLOBeforeCut

After:
YOLOAfterCut

There's still delay but it's significantly reduces.


LuigiPenco93 (approved):
approved this. It would still be worth understanding what is causing the delay. Also why the yolo mask stays for the door panel and trash can for quite a while after the objects are gone?

TomaszTB:
> It would still be worth understanding what is causing the delay.

It was the maximum queue size. It was allowing too long of a queue to form, so it was running old jobs with a lot of delay. 

>  Also why the yolo mask stays for the door panel and trash can for quite a while after the objects are gone?

Because that particular model hadn't been updated in a while. Since we're cycling between models, each model runs at a lower frequency. I was running 10 models in the GIF, so each one was getting updated very slowly. Also, occasionally if the queue is full it'll reject running a model. Then the update gets delayed even more.

ds58 (approved):

PR #847, opened 2025-06-09
Create a height map center that doesn't have to be the origin sensor in the case where we have multiple sensors generating a height map

PotatoPeeler3000:


ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -269,8 +269,8 @@ public void update(GpuMat latestDepthImageGPU,
 
       // ---------- Run the translate kernel ---------
       {
-         int currentCellX = (int) Math.round(sensorOrigin.getX32() / heightMapParameters.getCellSizeInMeters());
-         int currentCellY = (int) Math.round(sensorOrigin.getY32() / heightMapParameters.getCellSizeInMeters());
+         int currentCellX = (int) Math.round(heightMapFrameToWorldFrame.getX32() / heightMapParameters.getCellSizeInMeters());
```
PotatoPeeler3000, line 272: No longer tied to a sensor, can be any frame that gets passed into the construtor

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapManager.java
```diff
@@ -76,7 +77,13 @@ public RapidHeightMapManager(ROS2Node ros2Node,
 
    public void updateAndPublishHeightMap(GpuMat latestDepthImage, CameraIntrinsics depthIntrinsics, ReferenceFrame cameraFrame, ReferenceFrame cameraZUpFrame)
    {
-      updateInternal(latestDepthImage, depthIntrinsics, cameraFrame, cameraZUpFrame);
+      // Update the sensor origin here with the latest reference frame
```
PotatoPeeler3000, line 80: Pfffffff this is some nonsense. I made the sensorOrigin local so have to set it here, And now I'm setting the heightMapFrameToWorldFrame as well, so two Point3D's being used

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapManager.java
```diff
@@ -152,8 +163,6 @@ private void updateInternal(GpuMat latestDepthImage, CameraIntrinsics depthIntri
       // Update the Z translation of the sensor to match the world transform (to handle the sensor's vertical position)
       sensorToGround.getTranslation().setZ(sensorToWorld.getTranslation().getZ());
 
-      sensorOrigin.set(sensorToWorld.getTranslation());
```
PotatoPeeler3000, line 155: Happening locally now

TomaszTB (approved):
This is really good. Didn't even take that many changes

PR #852, opened 2025-06-10
Feature/hardware bringup

stefanfasano:
Branch for all things Alexander bringup. Main contributions include:

- Alexander controller and state estimator parameter tuning
- Cleanup and debugging of abstracted humanoid controller/multi-threading classes to be used for both Alexander and H1


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/AvatarEstimatorThread.java
```diff
@@ -95,6 +99,7 @@ public void run()
 
       try
       {
+         updateIteration.increment();
```
stefanfasano: @rjgriffin42 I believe we added this as a debug variable when the sensor processor was not updating. Should I go ahead and remove?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -109,33 +129,40 @@ public AvatarMultiThreadingManager(String prefix,
...
       periodicParameters = new PeriodicParameters(period);
 
       // Set up the master thread
       if (useRealtimeThreads)
       {
-         masterThread = new RealtimeThread(affinity.getSchedulerPriority(), periodicParameters, this::doControl, prefix + "-master-thread");
-         masterThread.setAffinity(affinity.getSchedulerThreadProcessor());
-         threadScheduler = null;
+         masterThread = new RealtimeThread(affinity.getSchedulerPriority(), periodicParameters, this::runRealtime, prefix + "-master-thread");
+         ((RealtimeThread) masterThread).setAffinity(affinity.getSchedulerThreadProcessor());
```
stefanfasano: I have a bit of a dual naming convention here, where I interchangeably use "Master Thread" and "Scheduler Thread".  Is there a preference for which to use? I'd rather keep it consistent

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/highLevelStates/SmoothTransitionControllerState.java
```diff
@@ -47,10 +47,11 @@ public SmoothTransitionControllerState(String namePrefix,
       this.finalControllerState = finalControllerState;
       this.commandInputManager = commandInputManager;
 
+      // FIXME (20250608 rgriffin) this can be true, but we want it to be false for alexander
```
stefanfasano: @rjgriffin42 how should we deal with this? An Alexander-specific version of this class?

calvertdw (approved):

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
james-p-foster: I'd slightly prefer Master because there's a whole notion of updating the "Master context" in `HumanoidRobotContextData`, but don't have much of a dog in this fight

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -501,25 +511,6 @@ public void addCustomControlState(HighLevelControllerStateFactory customControll
...
-         }
-         else
-         {
-            joints.add(joint);
-         }
-      }
-
-      return joints;
-   }
-
```
james-p-foster, line 522: This seems like an important guy to get rid of, no? If not, disregard.

james-p-foster (approved):

stephenmcc (approved):

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
Beomyeong: I'd like to raise my hand for the `masterThread`. From the `AvatarMultiThreadingManager`, we use `masterThread` for the realtimeThread. and `threadSchduler` for the not realtime threa. Using `masterThread` would be more coincident with current classes

stefanfasano: will do! Thanks for the input guys

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -228,33 +246,24 @@ public void buildThreads()
                                                            useRealtimeThreads,
                                                            useMultiThreading,
                                                            yoVariableServer,
-                                                           registry));
-
-      // Do some YoVariable stuff
-      if (yoVariableServer != null)
```
Beomyeong, line 234: Was it removed becuase now we alway run the yoVarialbleServer?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
Beomyeong: I'd like to raise my hand for `masterThread`

Beomyeong (approved):

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
stefanfasano: I'm not totally sure. We replaced it with this: `estimatorThread.get().getFullRobotModel().getRootJoint().subtreeList()`. I know we have a four bar joint but it is 1-to-1 with the actuator, so we have no knee four bar mechanism math or kinematic specifications.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/AvatarEstimatorThread.java
rjgriffin42: You can remove, but it also makes a logical variable to set the sequence ID in a publish message. If that doesn't already exist, that is. If it does already exist, feel free to remove.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/highLevelStates/SmoothTransitionControllerState.java
rjgriffin42: So these parameters load from XML files for the robot. I would make sure that the associated variables in the valkyrie xml are true. Otherwise, I think this is a good change, making the default behavior more conservative.

open-alexander/src/main/java/us/ihmc/alexander/parameters/controller/AlexanderStateEstimatorParameters.java
```diff
@@ -163,7 +248,7 @@ public boolean trustCoPAsNonSlippingContactPoint()
    @Override
    public double getPelvisLinearVelocityAlphaNewTwist()
    {
-      return 0.15;
+      return 0.2;
```
rjgriffin42: I think it would be helpful to add a comment as to what tuning this variable does. That is, does increasing it make it trust the IMU more or less than the kinemtics?

open-alexander/src/main/java/us/ihmc/alexander/parameters/controller/AlexanderStateEstimatorParameters.java
```diff
@@ -211,4 +296,11 @@ public List<IMUBasedJointStateEstimatorParameters> getIMUBasedJointStateEstimato
    {
       return imuBasedJointStateEstimatorParameters;
    }
+
+   @Override
+   public MomentumEstimatorMode getMomentumEstimatorMode()
+   {
+      return MomentumEstimatorMode.SIMPLE;
+   }
```
rjgriffin42: We want to switch away from "Simple", right? We should add that as a comment.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarAffinityInterface.java
```diff
@@ -11,19 +11,19 @@
  */
 public interface AvatarAffinityInterface
 {
-   Processor getSchedulerThreadProcessor();
+   Processor getMasterThreadProcessor();
```
rjgriffin42, line 14: where is this used? This seems like a change that is going to propagate across _a lot_ of repos. I also preferred the name scheduler to master, since it's intended to be the scheduler thread of the barrier scheduler

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/AvatarEstimatorThread.java
stefanfasano: a publish message to the low-level process? In that case, we don't have a sequence ID but we do publish the timestamp of the message (I believe). If you want us to add a sequence ID as well, should that be in the ROS communication class?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -97,9 +90,11 @@ public class AvatarMultiThreadingFactory
 
    // The remaining constructor arguments
    private final MonotonicTime period;
-   private final double schedulerDt;
+   private final double masterThreadDt;
```
rjgriffin42, line 93: I prefer the old convention, since it is the scheduler thread of the barrier scheduler. Why change to masterThreadDt?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarAffinityInterface.java
stefanfasano: this is used only in this repo. I created this interface just for alex and H1

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -167,7 +165,7 @@ public void join()
 
    public void stop()
    {
-      LogTools.info("Calling shutdown in the controller factory");
+      System.out.println("Calling stop in the multi-threading factory");
```
rjgriffin42, line 168: why would you change this from a LogTools call? You lose the stack trace using println, which makes it next to impossible to find where it's printing from.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -176,10 +174,12 @@ public void stop()
 
    public void destroy()
    {
+      this.stop();
+      System.out.println("Calling destroy in the multi-threading factory");
```
rjgriffin42, line 178: why would you change this from a LogTools call? You lose the stack trace using println, which makes it next to impossible to find where it's printing from.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
rjgriffin42: It got moved above, but you're right, the if case was lost.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
stefanfasano: I put up a vote and James and Beomyeong voted to switch from calling it scheduler to master. I prefer that as well. I think the main idea is that it is the dt for the master thread. the barrier scheduler gets called at that dt in the master thread loop, but it is really the update dt of the master thread itself

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
stefanfasano: Dex said something about logtools not always working during a shutdown hook (which this is a part of)

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
stefanfasano: I think for now it is ok to keep the "if" removed, since we are always running with the server

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
rjgriffin42: empirically it works. And I think it works because it looks like @stefanfasano is correct, from my understanding.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -79,23 +97,24 @@ public AvatarMultiThreadingManager(String prefix,
                                       AvatarControllerThread controllerThread,
                                       AvatarStepGeneratorThread stepGeneratorThread,
                                       AvatarAffinityInterface affinity,
-                                      double schedulerDt,
+                                      double masterThreadDt,
```
rjgriffin42, line 101: See my earlier comments.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -209,26 +246,51 @@ else if (!useRealtimeThreads && useMultiThreading)
...
       {
-         threadScheduler.shutdown();
+         ((RealtimeThread) masterThread).join();
       }
+      else
+         ((RepeatingTaskThread) masterThread).stopRepeating();
+
+      ThreadTools.sleep(1000L);
+
+      System.out.println("MASTER THREAD SHUTDOWN COMPLETE -- EXITING");
```
rjgriffin42, line 292: I'd love this to be LogTools and not println

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -209,26 +246,51 @@ else if (!useRealtimeThreads && useMultiThreading)
...
+
+      running = false;
+
+      if (useRealtimeThreads)
       {
-         threadScheduler.shutdown();
+         ((RealtimeThread) masterThread).join();
       }
+      else
+         ((RepeatingTaskThread) masterThread).stopRepeating();
```
rjgriffin42, line 288: this is kind of a preference and/or style guide thing, but if you do the brackets on one, it's preferred to do them on both.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -241,61 +303,77 @@ public void resume()
...
    }
 
-   private void doControl()
+   private void runRealtime()
+   {
+      while(running)
+      {
+         long idleTime = ((RealtimeThread) masterThread).waitForNextPeriod();
+         if (idleTime > 0)
+            run();
```
rjgriffin42, line 311: should there be an else block where you increment the missed deadlines counter?

stefanfasano: I thought missed deadlines was something specific to EtherCat. If not, I can add a missed deadlines counter and add that here

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -241,61 +303,77 @@ public void resume()
...
-      masterContext.setTimestamp(monotonicTimeProvider.getTimestamp()); //TODO(sfasano, 20250530) should this be before the hasReceivedFirstState check?
+      if (hardwareCommunicationInterface.hasReceivedFirstState())
+      {
+         masterContext.setTimestamp(monotonicTimeProvider.getTimestamp());
+
+         // Update estimator thread
+         long estimatorThreadStartTime = System.nanoTime();
+         estimatorThread.run();
+         estimatorThreadComputeTime.set(System.nanoTime() - estimatorThreadStartTime);
+         estimatorThreadFrequencyCalculator.ping();
```
rjgriffin42, line 339: does this do anything that the masterThreadFrequencyCalculator does not?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
stefanfasano: see my response above

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -241,61 +303,77 @@ public void resume()
...
+         estimatorThreadUpdateRate.set(estimatorThreadFrequencyCalculator.getFrequency());
+
+         // Update all master thread runnables
+         for (int i = 0; i < masterThreadRunnables.size(); i++)
+            masterThreadRunnables.get(i).run();
+
+         // Run the thread scheduler
+         long barrierSchedulerStartTime = System.nanoTime();
+         threadScheduler.doControl();
+         threadSchedulerComputeTime.set(System.nanoTime() - barrierSchedulerStartTime);
```
rjgriffin42, line 349: I want to make sure, if you call the barrier scheduler do control call, it does _not_ also call the state estimator thread, does it? Otherwise, you're calling compute twice on the state estimator thread every tick.

open-alexander/src/main/java/us/ihmc/alexander/AlexanderSensorInformation.java
```diff
@@ -35,14 +31,45 @@ public class AlexanderSensorInformation implements HumanoidRobotSensorInformatio
    private final String rightForearmIMUSensor = "right_wrist_yaw_imu";
    private final String rightHandIMUSensor = null; // "right_gripper_yaw_imu";
 
+   // Torso IMUs
+   private final String torsoIMU = "torso_imu";
+
+   // Pelvis IMUs
+   private final String pelvisIMU = "pelvis_stim_imu";
```
rjgriffin42, line 38: we really want to rename this here and in the URDF. It's not the stim_imu, since it's not a stim.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
stefanfasano: it does not because we don't add it to the thread scheduler. I had initially added it in the H1 setup so all three threads were handled in the thread scheduler, but I removed that

open-alexander/src/main/java/us/ihmc/alexander/AlexanderSensorInformation.java
stefanfasano: I can do that in a separate PR once this is merged

open-alexander/src/main/java/us/ihmc/alexander/parameters/controller/AlexanderHighLevelControllerParameters.java
```diff
@@ -7,7 +7,6 @@
 
 import us.ihmc.alexander.AlexanderJointMap;
```
rjgriffin42, line 8: I would like to see this class get copied into the AlexanderSoftware repo, and this class be renamed to OpenAlexander, so that we're not tuning across repos. Same thing for basically everything in the controller project here.

stefanfasano: I can rename it here, and throw the copy in another PR. In what instance would we not want all these parameters to be identical though? If the true and most up-to-date values are in the Alexander-Software repo, would we not have to update the open version of the parameters each time we make a change to make sure they remain accurate?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
calvertdw: I really preferred scheduler because it is so much more specific. Master is such a generic word. This is why git changed "master" to "main"

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
calvertdw: That's correct, it won't print at all with LogTools, but you could add "(AvatarMultiThreadingFactory.java:168):" manually

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
stefanfasano: The main difference is that this one won't ping if we haven't received our first message from the robot via ros. So the master thread frequency calculator can still show we are running at 1kHz, but the estimator thread frequency calculator will be flatlined until we receive our first message and the estimator actually starts running

open-alexander/src/main/java/us/ihmc/alexander/parameters/controller/AlexanderStateEstimatorParameters.java
stefanfasano: looking into it a bit, this variable is used in this line: `rootJointLinearVelocity.scaleAdd(-scaleFactor * alphaVelocityUpdate, footVelocityInWorld, rootJointLinearVelocity);` where it is `alphaVelocityUpdate`. I believe this indicates that setting it to 0.0 means not using the kinematic estimate at all, and setting it to 1.0 means fully using kinematics? Not totally sure though, this doesn't really make sense to me at all.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/AvatarEstimatorThread.java
rjgriffin42: I think a sequence ID should be added to the pubilshed state, along with a timestamp, yeah. And pulling that sequenceID from here makes the most sense to me.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
rjgriffin42: I don't think so, but you should look. I'm pretty sure it's a "controller on a realtime clock" thing. Regardless, the fact that you're skipping the scheduler thread because it overran (idle time being 0) should be captured via a variable. Otherwise, you're skipping a scheduler tick, which would delay everything by that dt.

open-alexander/src/main/java/us/ihmc/alexander/parameters/controller/AlexanderHighLevelControllerParameters.java
rjgriffin42: this is the conversation we had yesterday in the meeting, and why I'm half thinking of renaming the robot that's contained in open robotics to just avoid the confusion. 

We don't care about updating the stuff contained in open-robotics, because that's not what the robot runs from. it is meant to have the tests for the controller be contained next to the controller, as a representative system, rather than the true thing.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
rjgriffin42: I looked at the conversation that you had about it, and it seems to overlook that it controls the barrier scheduler. I'm not sure if that was completely understood when folks were discussing it.

To me, I don't really understand the desire to call it a "master dt" and "master thread". I get it when the context of the barrier scheduler is ignored, but I think that actually loses a lot of clarity when you remember that it's literally the DT that runs the barrier scheduler.

I'm also fine with changing it to estimator DT and estimator thread, since that's what it is. However, "master" doesn't tell me either that it's running the barrier scheduler in a multi-threaded context, or that it is the thing running the estimator.  Both provide more information as to what is going on than "master".

open-alexander/src/main/java/us/ihmc/alexander/parameters/controller/AlexanderHighLevelControllerParameters.java
calvertdw: Why don't we on CI, clone private alexander and run those tests?

calvertdw: and remove open alexander

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
calvertdw: I completely agree with Robert here, it should be called something with scheduler. I think as well when you're talking about the ethercat thread, just say ethercat instead of master or at least "ethercatMaster". Master is such an unspecific word.

calvertdw: Not to mention it's probably good to move away from the "master"-"slave" terminology in general. :sweat_smile:

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/AvatarEstimatorThread.java
stefanfasano: but the published state comes from the low-level process, not this one. we can publish it with the desired state?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
stefanfasano: ok I can add that

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
stefanfasano: One reason why I don't love schedulerThread is because we have the barrier scheduler, and the single-threaded version of it, both represented by a shared variable called threadScheduler (I think that is an appropriate name since that is basically what it is, and I don't want to call the variable the barrierScheduler... because when we are running single threaded and not using the BarrierSchedulerController that variable name would not make sense). With this in mind, if we called the master thread the scheduler thread, we now have the schedulerThread and the threadScheduler which is confusing. 

Moreover, the masterDt does more than just update the barrier scheduler. It is the Dt that defines the update rate of the primary thread (the master thread), in doing so it updates the barrier scheduler but also the hardware manager/interface, the estimator thread, the yovariable server, and other stuff. It is not just reserved for updating the barrier scheduler. 

I also don't like the idea of calling it the estimator thread because in theory that would not support an alternate setup where the estimator thread is not tied into the master thread as it is currently.

I also do not like the master-slave thing, but we don't really uses the term slave in this convention. The name we choose should indicate this is the primary/main/master thread that handles every single other thing in the entire mult-threaded setup. If I didn't know better, the name 'schedulerThread' would not really indicate that to me at all. We don't have to call it master, but it should be called something along those lines I think, same way the master context is called the master context

calvertdw: > schedulerThread and the threadScheduler 

Why is that confusing? That makes sense to me.

calvertdw: It makes sense to me that the scheduler has to run at the highest rate, because that's the only way it can do it's job.

rjgriffin42 (approved):

PR #860, opened 2025-06-12
Bugfix/use alexander legs

Beomyeong:
This enables the open-robotis use Alexander only legs urdf. 
The V0_LEGS_ROBOT only includes all the leg joints and spine Z. This doesn't include any neck, arm joints. 
All the test can run with thie V0_LEGS_ROBOT version. 
The test fails with V0_FULL_ROBOT still fails with V0_LEGS_ROBOT. 
A few passed tests with V0_FULL_ROBOT may fail with V0_LEGS_ROBOT


open-alexander/src/main/java/us/ihmc/alexander/AlexanderJointMap.java
```diff
@@ -59,10 +67,31 @@ public AlexanderJointMap(AlexanderPhysicalProperties alexanderPhysicalProperties
       this.hasArms = hasArms;
 
       List<String> jointNameList = new ArrayList<>();
-      for (Joints joint : Joints.values())
```
reesep5, line 62: Why switch from using Joints.values() to the individual body segments? It feels like the same thing, just with more steps

open-alexander/src/main/java/us/ihmc/alexander/parameters/AlexanderPresetArmConfigurations.java
```diff
@@ -0,0 +1,89 @@
...
+   private static final SideDependentList<double[]> WIDE_ARMS_CYCLOID_4DOF = new SideDependentList<>();
+   private static final SideDependentList<double[]> TUCKED_UP_ARMS_CYCLOID_4DOF = new SideDependentList<>();
+   private static final SideDependentList<double[]> DOOR_AVOIDANCE_CYCLOID_4DOF = new SideDependentList<>();
+
+   private static final SideDependentList<double[]> INITIAL_SETUP_CYCLOID_7DOF = new SideDependentList<>();
+   private static final SideDependentList<double[]> STAND_PREP_CYCLOID_7DOF = new SideDependentList<>();
+   private static final SideDependentList<double[]> HOME_CYCLOID_7DOF = new SideDependentList<>();
+   private static final SideDependentList<double[]> TUCKED_UP_ARMS_CYCLOID_7DOF = new SideDependentList<>();
+
+   // To tune these values, run AlexanderModelViewer as it allows you to tune joint angles with sliders
```
reesep5: May just want to put this as a TODO to tune these values

open-alexander/src/test/java/us/ihmc/alexander/AlexanderToeOffTest.java
```diff
@@ -87,7 +87,7 @@ public double getMaxStepLength()
     @Override
     public DRCRobotModel getRobotModel()
     {
-        OpenAlexanderRobotModel robotModel = new OpenAlexanderRobotModel(OpenAlexanderVersion.V0_FULL_ROBOT, RobotTarget.SCS)
+        OpenAlexanderRobotModel robotModel = new OpenAlexanderRobotModel(OpenAlexanderVersion.V0_LEGS_ROBOT, RobotTarget.SCS)
```
reesep5: Do you still want this as V0 legs?

open-alexander/src/main/java/us/ihmc/alexander/AlexanderJointMap.java
Beomyeong: The reason for this is for the Alexander lower body. 
The hasArmConfiguration only check the robot has elbow and wrist joints. It still contains and tries to set the shoulder joints. So, I made changes for this reason.

open-alexander/src/main/java/us/ihmc/alexander/parameters/AlexanderPresetArmConfigurations.java
Beomyeong: Good point!

Beomyeong: I would delete this class. Because, Luigi's merged PR include this file, even it is in the other place.

open-alexander/src/test/java/us/ihmc/alexander/AlexanderToeOffTest.java
Beomyeong: It doens't matter. I roll backed this.

james-p-foster (approved):

PR #867, opened 2025-06-19
Make ZED FPS and resolution settable.

calvertdw:


ds58 (approved):

TomaszTB (approved):
That's cool. Just gotta remember that we can't pass in any arbitrary FPS value. Gotta be something like 60, 30, 15.

PR #880, opened 2025-06-25
Add option to stream spine and leg jointspace from KST

calvertdw:


stephenmcc:
You'll need to unpack the spine joint trajectory in `MessageUnpackingTools.createWholeBodyStreamingMessageUnpacker` to get the trajectory to the controller.

For the case where pelvis taskspace is disabled and spine joints are enabled, the pelvis translation won't be controlled, which is unwanted. In other words we just have the one boolean (`isPelvisTaskspaceOutputEnabled`) for both pelvis orientation and translation. Another complication is that when WholeBodyStreamingMessage is unpacked in MessageUnpackingTools the full selection matrix is assumed.

I think a work-around could be to unpack the pelvis taskspace trajectory as-is in  `MessageUnpackingTools.createWholeBodyStreamingMessageUnpacker`, then afterwards if a spine trajectory is provided then cancel the angular selection of the pelvis taskspace trajectory.

stephenmcc (approved):

PR #904, opened 2025-07-22
Using foot polygon for CoP-based touchdown detection

stephenmcc:


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/WrenchBasedFootSwitch.java
```diff
@@ -133,9 +134,8 @@ public WrenchBasedFootSwitch(String namePrefix,
 
       footWrench = new Wrench(measurementFrame, (ReferenceFrame) null);
 
-      footMinX = computeMinX(contactablePlaneBody);
```
stefanfasano, line 136: i guess in theory these computeMinX, computeMaxX, etc could be deleted? may be worth keeping in case they are used in tests or something

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/WrenchBasedFootSwitchFactory.java
```diff
@@ -75,7 +75,7 @@ public FootSwitchInterface newFootSwitch(String namePrefix,
       if (contactThresholdForceParameter == null)
       {
          contactThresholdForceParameter = new DoubleParameter("ContactThresholdForce", registry, defaultContactThresholdForce);
-         copThresholdFractionParameter = new DoubleParameter("CoPThresholdFraction", registry, defaultCoPThresholdFraction);
+         copThresholdFractionParameter = new DoubleParameter("CoPThresholdFraction", registry, defaultCoPThresholdDistance);
```
stefanfasano, line 78: should the parameter variable and parameter name be switched from "fraction" to "distance" too? or would that screw too many things up with the old parameter xmls and all?

open-alexander/src/main/java/us/ihmc/openAlexander/parameters/controller/OpenAlexanderStateEstimatorParameters.java
```diff
@@ -259,9 +259,9 @@ public boolean useGroundReactionForcesToComputeCenterOfMassVelocity()
    public FootSwitchFactory getFootSwitchFactory()
    {
       WrenchBasedFootSwitchFactory factory = new WrenchBasedFootSwitchFactory();
-      factory.setDefaultContactThresholdForce(80.0);
-      factory.setDefaultCoPThresholdFraction(0.02);
-      factory.setDefaultSecondContactThresholdForceIgnoringCoP(Double.POSITIVE_INFINITY);
+      factory.setDefaultContactThresholdForce(50.0);
```
stefanfasano, line 288: any reason why these force thresholds changed?

stefanfasano (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/WrenchBasedFootSwitchFactory.java
stephenmcc: changed it to "distance"

open-alexander/src/main/java/us/ihmc/openAlexander/parameters/controller/OpenAlexanderStateEstimatorParameters.java
stephenmcc: The previous CoP touchdown detector only checked the CoP X position against the foot polygon, while now it checks along Y as well. Some simulations were falling because the foot would register a late touchdown, waiting for the CoP Y to get inside the foot, so I switched to the lower force threshold for considering the CoP and started using the higher force threshold to register touchdown for cases where the CoP still isn't inside the foot.

The simulations are working ok but these values should be tuned for hardware.

PR #908, opened 2025-07-25
Fix ZED & Controller Log Synchronization

calvertdw:


ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
```diff
@@ -72,6 +72,7 @@ public class ZEDImageSensor extends ImageSensor
    private long grabSequenceNumber = 0L;
    private Instant lastGrabTime;
    private boolean lastGrabFailed = false;
+   private long currentSensorTimestamp;
```
TomaszTB: Seems more like a...
```suggestion
   private long lastGrabTimestamp;
```

If you want the actual current timestamp, you should be able to just call sl_get_current_timestamp in the getCurrentSensorTimestamp method, right?

calvertdw: Fixed! Don't need a getCurrentSensorTimestamp method

TomaszTB (approved):

PotatoPeeler3000 (approved):

PR #935, opened 2025-08-15
Speed up the GPU main kernel for the height map based on suggestions from Nsight Compute

PotatoPeeler3000:
image


ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/SnappingTerrainExtractor.java
```diff
@@ -26,8 +25,12 @@
 
 public class SnappingTerrainExtractor
 {
-   private static final int BLOCK_SIZE_XY = 32;
    private static final boolean PRINT_TIMING_FOR_KERNELS = false;
+   /**
+    * The choice of 16 here is to utilize more SMs (Multi Processors) on the GPU.
+    * This was chosen based on GPU profiling and significantly effects performance.
+    */
+   private static final int BLOCK_SIZE_XY = 16;
```
PotatoPeeler3000, line 33: Jetson AGX Orin 64 GB: SM Count & Architecture
The Jetson AGX Orin 64GB features an Ampere GPU composed of:

2 Graphic Processing Clusters (GPCs)

8 Texture Processing Clusters (TPCs)

16 Streaming Multiprocessors (SMs), each with 128 CUDA cores and 4 third-generation Tensor Cores 
RidgeRun Developer Wiki
NVIDIA

Total: 16 SMs 
RidgeRun Developer Wiki
OpenZeka | NVIDIA Embedded Distribütörü

So unlike your desktop RTX 4080 with around 80 SMs, the Jetson Orin AGX only has 16 SMs.

What This Means for Your Launch Configuration
With your 201×201 grid:

Using BLOCK_SIZE_XY = 32, you get a grid of (7 × 7) = 49 blocks, which is still over 3× the number of SMs. That’s already good headroom—not underutilizing the GPU.

Using BLOCK_SIZE_XY = 16, you’d get a grid of (13 × 13) = 169 blocks, which is over 10× the SM count, giving even more opportunities for overlap.

In both cases, you’re above the SM count. With the 16 SMs, both 49 and 169 blocks will provide sufficient concurrency to keep the GPU busy, though the smaller blocks (16×16) would allow even more fine-grained scheduling and might help better hide memory or compute stalls.

Quick Comparison Table
Setting	Block Size	Grid Blocks	SMs Available	Blocks per SM (approx.)
Current desktop (RTX 4080)	32×32	49	~80	~0.6 (underutilized)
Jetson AGX Orin 64 GB	32×32	49	16	~3.1 (oversubscribed)
Jetson AGX Orin 64 GB	16×16	169	16	~10.6

On the Jetson AGX Orin, both 49 and 169 blocks exceed the number of SMs—this is ideal for keeping SMs busy.

Smaller block size tends to boost parallelism and can lead to better latency hiding. But there's a sweet spot—too tiny blocks can increase scheduling overhead, though 16×16 is usually still good.

Summary & Recommendations
Yes, your Jetson AGX Orin 64 GB has 16 SMs—far fewer than your RTX 4080, but your current grid size is already sufficient to keep them occupied.

49 blocks (from 32×32) give ~3 blocks per SM: decent utilization.

169 blocks (from 16×16) gives ~10 blocks per SM: even better for parallelism and latency hiding—though you might incur slightly higher scheduling overhead.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -60,70 +60,25 @@ __device__ int2 spherical_projection(float3 cellCenter, float *params)
...
 
-__device__ int2 perspective_projection(float3 point, float *params)
+__device__ int2 perspective_projection(float3 point, const float *params)
 {
     float x = point.x / point.z * params[DEPTH_FX] + params[DEPTH_CX];
     float y = point.y / point.z * params[DEPTH_FY] + params[DEPTH_CY];
     return make_int2(static_cast<int>(x), static_cast<int>(y));
 }
 
-__device__ float3 back_project_spherical(int yaw_index, int pitch_index, float depth, float *params)
```
PotatoPeeler3000, line 70: Unused

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -60,70 +60,25 @@ __device__ int2 spherical_projection(float3 cellCenter, float *params)
...
-__device__ float3 back_project_perspective(int2 pos, float Z, float *params)
+__device__ float3 back_project_perspective(int2 pos, float Z, const float *params)
 {
     float X = (pos.x - params[DEPTH_CX]) / params[DEPTH_FX] * Z;
     float Y = (pos.y - params[DEPTH_CY]) / params[DEPTH_FY] * Z;
     float3 point = make_float3(Z, -X, -Y);
     return point;
 }
 
-__device__ float get_spatial_average(int xIndex, int yIndex, unsigned short *globalHeightMap, size_t pitchGlobal, float *params)
```
PotatoPeeler3000, line 95: Unused

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -151,157 +105,149 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, float *zUpCameraTo
...
-                                      float *previousGlobalHeightMap, size_t pitchGlobal,
-                                      float *localMeanMap, size_t pitchLocalMean,
-                                      float *localVarianceMap, size_t pitchLocalVariance,
-                                      float *localMotionVarianceMap, size_t pitchLocalMotionVariance,
-                                      unsigned short *localSampleCountMap, size_t pitchLocalSampleCount,
-                                      float *params, float *sensorToZUpFrameTf, float *zUpToSensorFrameTf,
-                                      float *zUpCameraToWorldAlignedGround,
-                                      float linearMotionMagnitude, float angularMotionMagnitude,
-                                      float resetOffset)
+extern "C"
```
PotatoPeeler3000, line 108: Cached frequently used params into local variables → fewer repeated accesses.

Replaced most float3 temporaries with scalar floats when possible.

Added __restrict__ for pointers → compiler can optimize memory access.

Reduced live variables in the depth loop by reusing temporary variables.

Motion variance computation moved outside loop since it doesn’t depend on individual points.

This should reduce register usage significantly, allowing more warps per SM and higher occupancy.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -151,157 +105,149 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, float *zUpCameraTo
...
+    float resetOffset)
 {
     // Thread indices
     int xIndex = blockIdx.x * blockDim.x + threadIdx.x;
     int yIndex = blockIdx.y * blockDim.y + threadIdx.y;
 
-    // Grid dimensions (from params)
-    int depthWidth = static_cast<int>(params[DEPTH_INPUT_WIDTH]);
-    int depthHeight = static_cast<int>(params[DEPTH_INPUT_HEIGHT]);
+    // Cache params into local variables to reduce repeated global memory accesses
```
PotatoPeeler3000, line 126: 1. const pointers in CUDA
When you declare a pointer as const float* params:

cpp
Copy
Edit
__global__ void myKernel(const float* params)
It tells the compiler:

You won’t modify the memory pointed to by this pointer.

The compiler can assume these values are read-only, which enables several optimizations:

How const helps performance:
Caching in registers

The compiler can safely load values from params into registers at the start of the kernel because it knows no other thread will write to it.

This reduces repeated global memory accesses, which are slow (hundreds of cycles).

Use of constant memory

CUDA hardware has a small read-only constant cache (64 KB per GPU in many architectures).

If you mark a pointer const __restrict__, the compiler may automatically place it in the constant cache, which is very fast, often only 1-2 cycles latency.

Better compiler optimization

The compiler can reorder loads, eliminate redundant reads, and perform loop invariant code motion more aggressively.

2. __restrict__ pointers
When you declare a pointer as float* __restrict__ ptr, you are telling the compiler:

“This pointer is the only way to access this memory in this scope; no other pointer will alias it.”

Why this matters:
CUDA (and C/C++ compilers) cannot assume two pointers are independent unless you tell it.

Without __restrict__, every write via one pointer forces the compiler to assume all other pointers might point to the same memory, preventing some optimizations.

With __restrict__, the compiler knows that loads and stores to this pointer don’t affect any other pointer, which enables:

Reordering memory accesses

Keeping values in registers without worrying about stale values

Vectorization and instruction-level parallelism

3. const + __restrict__ together
Combining both:

cpp
Copy
Edit
__global__ void myKernel(const float* __restrict__ params)
Means:

Memory is read-only → compiler can aggressively cache values.

No aliasing → compiler knows reading params cannot be affected by other writes.

This combination often reduces register pressure and memory traffic, increasing occupancy and overall kernel throughput.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -151,157 +105,149 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, float *zUpCameraTo
...
+                            angularMotionMagnitude * distance * varPerRotationSpeed;
 
-    // Search within the window in the depth image
-    for (int pitchOffset = -static_cast<int>(params[SEARCH_WINDOW_HEIGHT] / 2);
-         pitchOffset < static_cast<int>(params[SEARCH_WINDOW_HEIGHT] / 2 + 1); pitchOffset += skip)
+    // Search depth image
+    for (int pitchOffset = -searchWindowHeight / 2; pitchOffset <= searchWindowHeight / 2; pitchOffset += searchSkip)
     {
         int pitchIdx = projectedPoint.y + pitchOffset;
+        if (pitchIdx < 0 || pitchIdx >= depthHeight) continue;
```
PotatoPeeler3000: Optimize breaking out of the loop sooner, don't even bother going into the other loop

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -151,157 +105,149 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, float *zUpCameraTo
...
-        for (int yawOffset = -static_cast<int>(params[SEARCH_WINDOW_WIDTH] / 2);
-             yawOffset < static_cast<int>(params[SEARCH_WINDOW_WIDTH] / 2) + 1; yawOffset += skip)
+        unsigned short *rowPtr = (unsigned short *)((char *)depthImage + pitchIdx * pitchDepth);
+
+        for (int yawOffset = -searchWindowWidth / 2; yawOffset <= searchWindowWidth / 2; yawOffset += searchSkip)
         {
             int yawIdx = projectedPoint.x + yawOffset;
+            if (yawIdx < 0 || yawIdx >= depthWidth) continue;
+
+            float depth = rowPtr[yawIdx] * 0.001f; // scale to meters
```
PotatoPeeler3000, line 230: Multiplication is way way way faster then division

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -151,157 +105,149 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, float *zUpCameraTo
...
+        if (zFwd < 0.0f) return;
+        projectedPoint = perspective_projection(make_float3(xFwd, yFwd, zFwd), params);
     }
 
-    // Distance from camera origin
-    float distance = length(cellCenterInSensor);
+    // Distance from camera
+    float distance = sqrtf(cellCenterInSensor.x * cellCenterInSensor.x +
+                           cellCenterInSensor.y * cellCenterInSensor.y +
+                           cellCenterInSensor.z * cellCenterInSensor.z);
```
TomaszTB, line 198: Is `length` part of `MathUtils.cuh`? I kinda prefer using `length` purely because it's more readable. It should be doing the exact same thing within the function, right? Or maybe it uses doubles... then sqrtf would be faster

TomaszTB (approved):
I'm assuming these changes don't break the algorithm 😆 Amazing speed up though

PR #946, opened 2025-08-21
[Feature] ihmc_hands_ros2 Support

TomaszTB:


open-alexander/src/main/java/us/ihmc/openAlexander/OpenAlexanderVersion.java
```diff
@@ -120,6 +121,21 @@ public boolean hasArm(RobotSide robotSide)
       };
    }
 
+   @Override
+   public boolean hasHandWithFingers(RobotSide side)
+   {
+      return this == V1_FULL_ROBOT;
```
LuigiPenco93, line 127: not sure this has any hand, might just have the nubs for the moment

LuigiPenco93 (approved):

open-alexander/src/main/java/us/ihmc/openAlexander/OpenAlexanderVersion.java
TomaszTB: It's the Inspire hands

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/hands/RDXHandManager.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Manages the UI for a humanoid robot's hands. A hand configuration is like "open", "closed", etc.
+ */
+public class RDXHandManager
+{
+   private final SideDependentList<RDXHandInterface> rdxHands = new SideDependentList<>();
+   private final Map<HandType, HandROS2HardwareCommunication<?, ?>> handCommunications = new HashMap<>();
+
+   private final Map<String, RDXHandQuickAccessButtons> quickAccessButtons = new HashMap<>();
```
calvertdw: Why not use a SideDependentList for the two second maps? Also, usually when you have a bunch of Map fields like this, it means you should extract a class. In this case it would be probably a RDXHandManager that manages just one hand, then whatever the parent class, has a SideDependentList of those

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/hands/RDXHandManager.java
```diff
@@ -0,0 +1,97 @@
...
+
+      for (RobotSide side : rdxHands.sides())
+         quickAccessButtons.put(rdxHands.get(side).getIdentifier(), new RDXHandQuickAccessButtons(baseUI, rdxHands.get(side)));
+   }
+
+   public void update()
+   {
+      // Update the hands
+      rdxHands.values().forEach(RDXHandInterface::update);
+      quickAccessButtons.values().forEach(RDXHandQuickAccessButtons::update);
```
calvertdw: Would prefer not starting to use forEaches, I like the standard for

ds58 (approved):

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/hands/RDXHandManager.java
TomaszTB: The map was residual from when we thought we'd support more than just two hands, so I was avoiding side dependent lists. Switched it to a SideDependentList

TomaszTB: It's residual from when we thought we'd support more than just two hands, so I was avoiding side dependent lists. That idea went away so I updated it to a side depenent list. Will keep the two side dependent lists for now; we can clean it up later.

open-alexander/src/main/java/us/ihmc/openAlexander/OpenAlexanderVersion.java
LuigiPenco93: this is alexander

PR #948, opened 2025-08-21
CSG Terrain Snapping

stephenmcc:
Integrates the HeightMap with CSG. Removes various unused features for terrain snapping and adjustment, some being based on planar regions.

RDX-SCS Environments


PotatoPeeler3000 (approved):

rjgriffin42 (approved):
Looks good from a very high level overview.

PR #962, opened 2025-08-28
hand panel formatting

LuigiPenco93:
Screenshot 2025-08-28 111453

the red comment will stay until we support the simulation of the hands


TomaszTB (approved):
Yeah, that looks good to me

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/hands/psyonicAbilityHand/RDXAbilityHand.java
calvertdw, line 1: I would prefer not to start a trend of yelling at the user in all caps. Can we just use capitalized names for all these elements?

PR #970, opened 2025-09-03
Setting 2d footstep pose in CSG prior to adjustment calls

stephenmcc:
The CSG was using footstep adjustments to set the position and yaw. Now setting the position and yaw directly and only using the adjustment to set position Z, pitch and roll.

CSG_Adjustment_bugfix


stefanfasano (approved):

stefanfasano:
I was told that to prevent this bug we set the environmental constraints for the CSG to null, should that be un-done now that there is a proper fix?

reesep5 (changes requested):
You will need to set the environmental constraint, as it is currently still null. Once done, if it works on robot, we should be good to go

reesep5 (approved):

PR #979, opened 2025-09-04
Calculating squared error when fitting plane

stephenmcc:


ihmc-footstep-planning/src/main/resources/us/ihmc/footstepPlanning/SnappingTerrainExtractor.cu
```diff
@@ -264,6 +263,10 @@ __global__ void computeTerrainData(float *heightMap, size_t pitchHeightMap,
     int normal_z_int = static_cast<int>(255 * (normal.z + 1.0f) / 2.0f);
     int2 storage_key = make_int2(x_index, y_index);
 
+    float squared_error_scale = 0.1;
+    float alpha_squared_error = fminf(fmaxf(squared_error / squared_error_scale, 0.0f), 1.0f);
+    int squared_error_int = static_cast<int>(255 * alpha_squared_error);
```
TomaszTB: This casting is kinda crazy. Let me make sure I understand correctly:
- You multiply `squared_error` by 10, then clamp it to be between 0.0 and 1.0.
- You multiply again by 255 and cast it to an integer, so now it's between 0 and 255
- Later you cast that integer to an unsigned char. 

If that's all correct, these are my questions:
1. Could just multiply by 2550 and clamp it to be between 0 and 255, right? That'd be a bit faster. 
2. Why not cast it directly to unsigned char instead of casting to int then unsigned char? 
3. Is it worth doing all this or would it be better to just use a matrix of floats so there's no need for any casting?

TomaszTB (commented):
Can't really review the math coz that's beyond my understanding, but one thing kinda stood out to me

ihmc-footstep-planning/src/main/resources/us/ihmc/footstepPlanning/SnappingTerrainExtractor.cu
stephenmcc: > Is it worth doing all this or would it be better to just use a matrix of floats so there's no need for any casting?

I agree this makes the most since, given that the other values are packed as `unsigned char` I kept the convention. See `area_fraction` for example. Unclear why this convention is in place and whether packing floats is a bad idea, so that's the main question. I'm fine switching if floats aren't off limits. But for using `unsigned char`, I'm having the squared error by on a 0->1 scale where 1 corresponds to an RMS error of 0.1m, which could be extracted as a parameter eventually.

PotatoPeeler3000 (approved):

TomaszTB (approved):

PR #982, opened 2025-09-09
Change the height map to run through the entire depth image as threads rather then each cell on the height map

PotatoPeeler3000:
image

From develop----------------

And then...
image
From my branch-------------

So these images are from NSight Compute. What its showing is generally "how much am I using the GPU" when I ask to use the GPU. What we want to do is when I tell the process "Hey buddy, use the GPU". I want to absolutely max this out and get it over with. Not quite 100% because that can cause its own problems. But hey, I've asked to use a GPU, freaking use it. So the image from my branch shows about a double usage of the GPU. Which is great. Of course the downside of profiling still occurs where this is specific to my machine.
If you have any question about the images let me know! I can come explain it.

### DIFF
--- a/ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
+++ b/ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
@@ -41,6 +41,10 @@ public class RapidHeightMapExtractor
    private final dim3 blockSize;
 
    // These are the mats required to extract the depth data
+   private final GpuMat tempSumMap;
+   private final GpuMat tempCountMap;
+   private final GpuMat tempSumOfSquaresMap;
+   private final GpuMat tempMotionVarianceMap;
    private final GpuMat localMeanMap;
    private final GpuMat localVarianceMap;
    private final GpuMat localMotionVarianceMap;
@@ -53,7 +57,8 @@ public class RapidHeightMapExtractor
    private final GpuMat terrainCroppedHeightMap;
    private final GpuMat emptyGlobalHeightMap;
 
-   private final CUDAKernel updateKernel;
+   private final CUDAKernel updateTempMapsKernel;
+   private final CUDAKernel localMapKernel;
    private final CUDAKernel translateKernel;
    private final CUDAKernel registerKernel;
    private final CUDAKernel terrainCroppingKernel;
@@ -103,19 +108,26 @@ public class RapidHeightMapExtractor
       {
          heightMapProgram = new CUDAProgram(kernelPath, heightMapUtilsHeaderPath, mathUtilsHeaderPath);
 
-         updateKernel = heightMapProgram.loadKernel("heightMapUpdateKernel");
+         updateTempMapsKernel = heightMapProgram.loadKernel("heightMapUpdateDataKernel");
+         localMapKernel = heightMapProgram.loadKernel("computeLocalMap");
          translateKernel = heightMapProgram.loadKernel("translateHeightMapKernel");
          registerKernel = heightMapProgram.loadKernel("heightMapRegistrationKernel");
          terrainCroppingKernel = heightMapProgram.loadKernel("terrainCroppingHeightMapKernel");
          planOffsetKernel = heightMapProgram.loadKernel("planOffsetKernel");
          emptyRegisterKernel = heightMapProgram.loadKernel("heightMapEmptyRegistrationKernel");
 
-         updateKernel.enableKernelTimings(PRINT_TIMING_FOR_KERNELS);
+         updateTempMapsKernel.enableKernelTimings(PRINT_TIMING_FOR_KERNELS);
+         localMapKernel.enableKernelTimings(PRINT_TIMING_FOR_KERNELS);
          translateKernel.enableKernelTimings(PRINT_TIMING_FOR_KERNELS);
          registerKernel.enableKernelTimings(PRINT_TIMING_FOR_KERNELS);
          terrainCroppingKernel.enableKernelTimings(PRINT_TIMING_FOR_KERNELS);
          planOffsetKernel.enableKernelTimings(PRINT_TIMING_FOR_KERNELS);
          emptyRegisterKernel.enableKernelTimings(PRINT_TIMING_FOR_KERNELS);
+
+         tempSumMap = new GpuMat(cellsPerAxisLocal, cellsPerAxisLocal, opencv_core.CV_32FC1);
+         tempCountMap = new GpuMat(cellsPerAxisLocal, cellsPerAxisLocal, opencv_core.CV_32FC1);
+         tempSumOfSquaresMap = new GpuMat(cellsPerAxisLocal, cellsPerAxisLocal, opencv_core.CV_32FC1);
+         tempMotionVarianceMap = new GpuMat(cellsPerAxisLocal, cellsPerAxisLocal, opencv_core.CV_32FC1);
 
          // Initialize matrices and images
          localMeanMap = new GpuMat(cellsPerAxisLocal, cellsPerAxisLocal, opencv_core.CV_32FC1);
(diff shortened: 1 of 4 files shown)


ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -222,40 +228,56 @@ public void update(GpuMat latestDepthImageGPU,
...
-         updateKernel.withPointer(groundToSensorTransformDevicePointer);
-         updateKernel.withPointer(zUpCameraToWorldAlignedGroundDevicePointer);
-         updateKernel.withFloat(linearMotionMagnitude);
-         updateKernel.withFloat(angularMotionMagnitude);
-         updateKernel.withFloat(resetOffset);
-
-         updateKernel.run(stream, updateKernelGridDim, blockSize, 0);
-
-         updateKernelGridDim.close();
+         // Clearing all the temp maps so they are ready for the next update call
```
PotatoPeeler3000, line 236: Comment explains what this does, but we are going to re-use these temp maps each update. And this seems to be the best way to clear the mat without looping through it in the GPU kernel.

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -222,40 +228,56 @@ public void update(GpuMat latestDepthImageGPU,
...
+         cudaMemset2D(tempSumOfSquaresMap.data(), tempSumOfSquaresMap.step(), 0, (long) tempSumOfSquaresMap.cols() * Float.BYTES, tempSumOfSquaresMap.rows());
+         cudaMemset2D(tempMotionVarianceMap.data(),
+                      tempMotionVarianceMap.step(),
+                      0,
+                      (long) tempMotionVarianceMap.cols() * Float.BYTES,
+                      tempMotionVarianceMap.rows());
+
+         int gridDimX = (latestDepthImageGPU.cols() + BLOCK_SIZE_XY - 1) / BLOCK_SIZE_XY;
+         int gridDimY = (latestDepthImageGPU.rows() + BLOCK_SIZE_XY - 1) / BLOCK_SIZE_XY;
+         dim3 updateTempMapsDim = new dim3(gridDimX, gridDimY, 1);
```
PotatoPeeler3000, line 248: This is an important change. We now launch a thread per pixel in the image. There are pros and cons of this.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -31,41 +31,6 @@ extern "C"
 #define SEARCH_SKIP_SIZE 26
 #define GROUND_HEIGHT 27
 
-#define VERTICAL_FOV 1.5707963267948966f
```
PotatoPeeler3000, line 34: Old code for the oyster, removed

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -97,165 +62,137 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, const float *zUpCa
...
-                                      float *__restrict__ localVarianceMap, size_t pitchLocalVariance,
-                                      float *__restrict__ localMotionVarianceMap, size_t pitchLocalMotionVariance,
-                                      const float *__restrict__ params,
-                                      const float *__restrict__ sensorToZUpFrameTf,
-                                      const float *__restrict__ zUpToSensorFrameTf,
-                                      const float *__restrict__ zUpCameraToWorldAlignedGround,
-                                      const float linearMotionMagnitude,
-                                      const float angularMotionMagnitude,
-                                      float resetOffset)
+__global__ void heightMapUpdateDataKernel(const unsigned short* __restrict__ depthImage, size_t pitchDepth,
```
PotatoPeeler3000, line 72: Honestly I found this fairly simple to follow. Given a pixel in the deph image. back project it and transform it into the right frame. Then find the cell where its coordinates fall into. And add the depth to the sum of that cell.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -97,165 +62,137 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, const float *zUpCa
...
-    // Transform cell center from Z-Up to sensor frame
-    float3 cellCenterInSensor = transformPoint3D(make_float3(cellX, cellY, cellZ), zUpToSensorFrameTf);
+    // Pointers to the target cell in each temporary map
+    float* sumPtr = (float*)((char*)sumMap + cellIndex.y * pitchSum) + cellIndex.x;
+    float* countPtr = (float*)((char*)countMap + cellIndex.y * pitchCount) + cellIndex.x;
+    float* sumSqPtr = (float*)((char*)sumOfSquaresMap + cellIndex.y * pitchSumSq) + cellIndex.x;
+    float* motionVarPtr = (float*)((char*)motionVarianceSumMap + cellIndex.y * pitchMotionVar) + cellIndex.x;
+
+    // Atomically add this pixel's contribution
+    atomicAdd(sumPtr, queryPointInZUpFrame.z);
```
PotatoPeeler3000, line 126: This is a cuda thing. It allows to share the pointer between threads and add to it in a safe manner. (atomicAdd())

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -97,165 +62,137 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, const float *zUpCa
...
-        unsigned short *rowPtr = (unsigned short *)((char *)depthImage + pitchIdx * pitchDepth);
+    // Only compute if one or more points landed in this cell
+    if (count > 0.5f)
+    {
+        float sum = *sumPtr;
+        mean = sum / count;
+        motionVariance = *motionVarPtr / count;
 
-        for (int yawOffset = -searchWindowWidthHalf; yawOffset <= searchWindowWidthHalf; yawOffset += searchSkip)
+        if (count > 1.5f)
```
PotatoPeeler3000: Why this? Now that's a good question

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -97,165 +62,137 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, const float *zUpCa
...
-    // Distance from camera
-    float distance = sqrtf(cellCenterInSensor.x * cellCenterInSensor.x +
-                           cellCenterInSensor.y * cellCenterInSensor.y +
-                           cellCenterInSensor.z * cellCenterInSensor.z);
+/**
+ * @brief Compute Local Map KERNEL: The threads correspond to cell indices
+ * Takes the maps that hold all the sums, and computes the mean, variance, and motion variance per cell
+ */
+extern "C"
+__global__ void computeLocalMap(const float* __restrict__ sumMap, size_t pitchSum,
```
PotatoPeeler3000, line 144: We could do this in the kernel above but it gets a little messy

PotatoPeeler3000: So we split it up into two kernels

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -97,165 +62,137 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, const float *zUpCa
...
-                count++;
-                float delta = queryPointInZUp.z - meanZ;
-                meanZ += delta / count;
-                float delta2 = queryPointInZUp.z - meanZ;
-                m2 += delta * delta2;
-            }
+            float sumSq = *sumSqPtr;
+            // Stable one-pass variance formula: (E[X^2] - (E[X])^2 * N) / (N-1)
+            variance = (sumSq - (sum * sum) / count) / (count - 1.0f);
+            variance = fmaxf(0.0f, variance); // Clamp to zero to avoid negatives from float error
```
PotatoPeeler3000, line 185: @PotatoPeeler3000 Should probably do a pass on this, to verify it

ihmc-perception/src/main/resources/us/ihmc/perception/heightMap/HeightMapParameters.json
```diff
@@ -50,7 +50,7 @@
     "upperBound": 1.0
   },
   "Variance per meter": {
-    "value": 0.01,
+    "value": 0.1,
```
PotatoPeeler3000, line 53: Little less noisy

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -222,40 +228,56 @@ public void update(GpuMat latestDepthImageGPU,
...
-         updateKernel.withFloat(resetOffset);
-
-         updateKernel.run(stream, updateKernelGridDim, blockSize, 0);
-
-         updateKernelGridDim.close();
+         // Clearing all the temp maps so they are ready for the next update call
+         cudaMemset2D(tempSumMap.data(), tempSumMap.step(), 0, (long) tempSumMap.cols() * Float.BYTES, tempSumMap.rows());
+         cudaMemset2D(tempCountMap.data(), tempCountMap.step(), 0, (long) tempCountMap.cols() * Float.BYTES, tempCountMap.rows());
+         cudaMemset2D(tempSumOfSquaresMap.data(), tempSumOfSquaresMap.step(), 0, (long) tempSumOfSquaresMap.cols() * Float.BYTES, tempSumOfSquaresMap.rows());
+         cudaMemset2D(tempMotionVarianceMap.data(),
```
TomaszTB: Should use cudaMemset2DAsync

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -97,165 +62,137 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, const float *zUpCa
...
-    {
-        float *globalHeight = (float *)((char *)previousGlobalHeightMap + globalIndex.x * pitchGlobal) + globalIndex.y;
-        if (*globalHeight != resetOffset)
-        {
-            cellZ = *globalHeight;
-        }
-    }
+    // Early exit for invalid depth
+    if (depth < 0.5f)
+        return;
```
TomaszTB: 0.5 seems quite high to me. The height map won't register anything half a meter in from of the sensor? It's probably fine for most cases but I might reduce it to 0.25 or something

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -97,165 +62,137 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, const float *zUpCa
...
-    float maxY = cellY + halfCellWidth;
+    // Bounds check against the local map dimensions because we are scanning the entire depth image
+    if (cellIndex.x < 0 || cellIndex.x >= localCellsPerAxis || cellIndex.y < 0 || cellIndex.y >= localCellsPerAxis)
+        return;
 
-    // Transform cell center from Z-Up to sensor frame
-    float3 cellCenterInSensor = transformPoint3D(make_float3(cellX, cellY, cellZ), zUpToSensorFrameTf);
+    // Pointers to the target cell in each temporary map
+    float* sumPtr = (float*)((char*)sumMap + cellIndex.y * pitchSum) + cellIndex.x;
+    float* countPtr = (float*)((char*)countMap + cellIndex.y * pitchCount) + cellIndex.x;
```
TomaszTB: Can we make the count map a 8UC1 GpuMat? That way it's unsigned char* and you can do integer math, instead of doing floating point math/comparison on values that are always integers

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -97,165 +62,137 @@ __device__ int2 getGlobalIndexFromLocalIndex(int2 localIndex, const float *zUpCa
...
-            continue;
+    float count = *countPtr;
+    float mean = 0.0f;
+    float variance = 0.0f;
+    float motionVariance = 0.0f;
 
-        // This is created outside the inner loop because its cheaper to only create once per loop
-        unsigned short *rowPtr = (unsigned short *)((char *)depthImage + pitchIdx * pitchDepth);
+    // Only compute if one or more points landed in this cell
+    if (count > 0.5f)
```
TomaszTB: Like here you could just do `if (count > 0)` if you use 8UC1 (unsigned char)

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
TomaszTB: Because count is a floating point value for no reason?

TomaszTB (commented):
This makes the code much more readable for me. It's awesome. Got a few minor complains though 😆

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
PotatoPeeler3000: Ya agreed, I'll try that

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
PotatoPeeler3000: Nice, done!

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
PotatoPeeler3000: Ya this makes sense now, because if we need to compute the variance we need to do this formula

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
PotatoPeeler3000: lol changed to > 1

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
PotatoPeeler3000: Yep makes sense

PotatoPeeler3000:
```
Note #3326-D: function "atomicAdd(int *, int)" does not match because argument #1 does not match parameter
Note #3326-D: function "atomicAdd(unsigned int *, unsigned int)" does not match because argument #1 does not match parameter
Note #3326-D: function "atomicAdd(unsigned long long *, unsigned long long)" does not match because argument #1 does not match parameter
Note #3326-D: function "atomicAdd(float *, float)" does not match because argument #1 does not match parameter
Note #3326-D: function "atomicAdd(double *, double)" does not match because argument #1 does not match parameter
```

PotatoPeeler3000: I made it an int, that is one of the compatible types

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
PotatoPeeler3000: Changed it to 0.25, and seems fine

PR #991, opened 2025-09-22
Fix timestamps for exported parquet & video.

calvertdw:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/logProcessor/leRobot/LeRobotDataset.java
```diff
@@ -48,7 +48,6 @@ public class LeRobotDataset
    private final List<String> taskNames = new ArrayList<>();
    private final List<LeRobotDatasetEpisode> episodes = new ArrayList<>();
    private long totalFrames = 0L;
-   private boolean usePerfectTimestamps = true;
```
calvertdw, line 51: Remove option to do perfect timestamps. We have to do it always.

ds58 (approved):

PR #998, opened 2025-09-24
Disable ROS 1 support, removing all transitive dependencies but keeping old code compiling.

calvertdw:
Remove vendored netty classes, because we actually use Netty in some places.


ihmc-ros-tools/build.gradle.kts
calvertdw, line 1: This is the main benefit. This code pulls in a specific old versions of Netty and apache commons, which is very undesirable.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/logProcessor/leRobot/LeRobotDatasetTools.java
```diff
@@ -1,6 +1,6 @@
 package us.ihmc.avatar.logProcessor.leRobot;
 
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
```
calvertdw, line 3: Apparently I accidentally used an old version of apache here transitively provided through rosjava.

ds58 (approved):

TomaszTB (approved):

PR #999, opened 2025-09-24
Fixing failning tests

Beomyeong:
Adding registry variable of neverholdposition and neverholdorientaion… to be turend off. Turning off is better when there are external forces.

This should be investigated because, for flat walking, turning on is better. Also, pushing left force should have plus sign.


calvertdw (approved):

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/pushRecovery/AvatarQuickPushRecoveryWalkingTest.java
```diff
@@ -241,7 +241,7 @@ public void testOutwardPushInitialTransferToLeftStateAndLeftMidSwing()
       setupTest();
 
       // setup all parameters
-      Vector3D forceDirection = new Vector3D(0.0, -1.0, 0.0);
+      Vector3D forceDirection = new Vector3D(0.0, 1.0, 0.0);
```
james-p-foster, line 244: We should switch this back to -1.0 to comply with the semantics in the test method name, which require the robot to be pushed outwards.

james-p-foster (changes requested):

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/pushRecovery/AvatarQuickPushRecoveryWalkingTest.java
Beomyeong: Yes, that's the reason I changed. that. 
image

With that "+1" the robot pushed to outward for the leftFootsupport.

james-p-foster (approved):

PR #1011, opened 2025-09-30
Cleanup/terrain map

PotatoPeeler3000:
Ok so most of this is attempting to put all the gpu kernels in the same class path. We compute the height map, and we compute the terrain map. Those are pretty much the two gpu kernels we expect to run for our current footstep planner.

Previously it was setup so the height map was computed, We would turn that into `HeightMapData`, and pass that to the terrain map kernels which would then pack the `TerrainMapData`. And both of those data structures were sent to the footstep planner for planning. And both were published to the UI for visualization.

That was getting pretty complicated to maintain. Testing was complicated cause the terrain map was being computed in the footstep planner module. We were going from gpu data, back to cpu, then back to gpu for the terrain map. Both were being used in the footstep planner, even though they were pretty much the same thing.

Now I've combined the gpu kernels so they all live in the same thread (woah!). And everything is contained in the TerrainMapData so that is the only thing required for the footstep planning. Meaning the HeightMapData doesnt need to exist at all actually, but due to future networking restrictions it seems valuable to keep in case we need it.

At the end of the day, performance is pretty much the same, the maps are the same, and the planning is the same. I'm hoping its just a lot easier to maintain and expand.


ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/ros2/RDXROS2HeightMapVisualizer.java
```diff
@@ -97,14 +97,11 @@ public void acceptHeightMapMessage(HeightMapMessage heightMapMessage)
       if (!isActive())
          return;
 
-      if (!colorBasedOnTraversability.get())
```
PotatoPeeler3000, line 100: We don't really care if we are coloring via traversability. We can still receive the mesasge.

PotatoPeeler3000: Honestly should update this to depend on the terrain map entirely

PotatoPeeler3000: But in a future PR

ihmc-high-level-behaviors/src/main/java/us/ihmc/behaviors/activeMapping/ActiveMappingParameterToolBox.java
```diff
@@ -21,7 +21,7 @@ public class ActiveMappingParameterToolBox
    private final HeightMapParameters heightMapParameters;
    private final ROS2StoredPropertySetGroup ros2StoredPropertySetGroup;
    private final DepthImageFilteringParameters depthImageFilteringParameters;
-   private final SteppableRegionCalculatorParameters steppableRegionCalculatorParameters;
+   private final TerrainMapParameters terrainMapParameters;
```
PotatoPeeler3000, line 24: You will see this change all over the place. Name change

ihmc-high-level-behaviors/src/main/java/us/ihmc/perception/GpuMappingThread.java
```diff
@@ -30,26 +32,27 @@
 
 import java.util.concurrent.BlockingQueue;
 
-public class RapidHeightMapThread extends RepeatingTaskThread
+public class GpuMappingThread extends RepeatingTaskThread
```
PotatoPeeler3000, line 35: Ok so now I've got a GpuMappingThread. And that calls the GpuMappingManager. Which calls each extractor.

ihmc-interfaces/src/main/generated-java/perception_msgs/msg/dds/HeightMapStateRequestMessagePubSubType.java
PotatoPeeler3000, line 1: Handful of old things I wanted to remove

ihmc-perception/src/main/java/us/ihmc/perception/cuda/CUDAFootstepOptimizer.java
```diff
@@ -3,15 +3,14 @@
...
 import org.bytedeco.cuda.global.cudart;
-import org.bytedeco.javacpp.DoublePointer;
 import org.bytedeco.javacpp.FloatPointer;
 import org.bytedeco.javacpp.IntPointer;
 import org.bytedeco.javacpp.PointerScope;
 import us.ihmc.euclid.referenceFrame.FramePose3D;
 import us.ihmc.euclid.referenceFrame.interfaces.FramePose3DReadOnly;
 import us.ihmc.euclid.yawPitchRoll.YawPitchRoll;
 import us.ihmc.log.LogTools;
-import us.ihmc.perception.gpuHeightMap.PlanarityChecker;
```
PotatoPeeler3000, line 14: We only have the gpu height map now, so really don't need that in the file path

ihmc-perception/src/main/java/us/ihmc/perception/comms/PerceptionComms.java
```diff
@@ -37,4 +37,7 @@ public class PerceptionComms
 
    public static final StoredPropertySetROS2TopicPair HEIGHT_MAP_PARAMETERS
          = new StoredPropertySetROS2TopicPair(MODULE_NAME,"height_map_parameters");
+
+   public static final StoredPropertySetROS2TopicPair TERRAIN_MAP_PARAMETERS
```
PotatoPeeler3000, line 41: Yep

ihmc-perception/src/main/resources/us/ihmc/perception/cuda/MathUtils.cuh
```diff
@@ -118,10 +118,10 @@ __device__ float* invert3x3Matrix(float* matrix, float* result)
    result[8] = detMinor22 / det;
 }
 
-__device__ float solveForPlaneCoefficients(float* covariance_matrix, float* z_variance_vector, float zz, float* coefficients)
+__device__ double solveForPlaneCoefficients(double* covariance_matrix, double* z_variance_vector, double zz, double* coefficients)
```
PotatoPeeler3000, line 121: Ok so this as floats was causing problems. Wasn't getting the correct values so had to switch to doubles.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.cu
```diff
@@ -306,79 +273,55 @@ __global__ void heightMapRegistrationKernel(const float *__restrict__ localMeanM
 }
 
 extern "C"
-__global__ void terrainCroppingHeightMapKernel(const float *__restrict__ globalHeightMap, size_t pitchGlobal,
```
PotatoPeeler3000, line 309: Oh I forgot I removed this lol, isn't needed anymore, things are so fast we don't need a smaller map for the terrain map

ihmc-perception/src/main/java/us/ihmc/perception/gpuHeightMap/RapidHeightMapExtractor.java
```diff
@@ -18,22 +19,19 @@
...
-public class RapidHeightMapExtractor
+public class HeightMapExtractor
 {
    private static final boolean PRINT_TIMING_FOR_KERNELS = false;
    /**
     * The choice of 16 here is to utilize more SMs (Multi Processors) on the GPU.
     * This was chosen based on GPU profiling and significantly effects performance.
     */
    private static final int BLOCK_SIZE_XY = 8;
-   private static final int MODE = 1; // 0 -> Ouster, 1 -> Realsense
```
PotatoPeeler3000, line 36: Not supporting the Ouster stuff at all so removed this

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/HeightMapExtractor.java
```diff
@@ -196,18 +180,25 @@ public void update(GpuMat latestDepthImageGPU,
...
 
       // This transformation only has rotation
       RigidBodyTransform groundToWorldAlignedGround = new RigidBodyTransform(worldToGroundNoRotation);
       groundToWorldAlignedGround.multiply(groundToWorldTransform);
 
+      // Step 3: Multiply with full ground->world to keep rotation, giving aligned ground
+      RigidBodyTransform sensorToWorldAlignedGround = new RigidBodyTransform(worldToGroundNoRotation);
+      sensorToWorldAlignedGround.multiply(groundToWorldTransform);
+
+      // Step 4: Apply sensor->ground transform
```
PotatoPeeler3000, line 199: This is a pretty big change. Previously the local map was aligned in orientation with the camera. Which meant when you mapped from the local grid to the world grid you had to go through a transform because the grids didn't have the same yaw. Now everything is on the world aligned grid so we don't need to do that. Becomes a little simpler.

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/networkProcessor/footstepPostProcessing/AvatarPostProcessingTests.java
```diff
@@ -9,6 +9,7 @@
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import perception_msgs.msg.dds.HeightMapMessage;
```
TomaszTB: This import isn't used anymore
```suggestion
```

ihmc-communication/src/main/java/us/ihmc/communication/PerceptionAPI.java
```diff
@@ -148,6 +150,11 @@ public final class PerceptionAPI
    public static final ROS2Topic<Empty> RESET_HEIGHT_MAP = PERCEPTION_MODULE.withSuffix("reset_height_map").withType(Empty.class);
    public static final ROS2Topic<Empty> LOWER_HEIGHT_MAP_BACKDROP = PERCEPTION_MODULE.withSuffix("lower_height_map_backdrop").withType(Empty.class);
 
+   /*
+    * Terrain map
+    */
+   public static final ROS2Topic<TerrainMapMessage> TERRAIN_MAP_MESSAGE = TERRAIN_MAP_MODULE.withOutput().withTypeName(TerrainMapMessage.class);
```
TomaszTB, line 156: So the topic is `/ihmc/terrain_map/output/TerrainMap`? Why not just `/ihmc/terrain_map`? Or even just `/terrain_map`??? Why does our ROS2Topic class encourage weird topic names 😭

TomaszTB (approved):

ihmc-communication/src/main/java/us/ihmc/communication/PerceptionAPI.java
PotatoPeeler3000: lol not sure, I just pattern matched after the height map

PR #1019, opened 2025-10-06
Sychronize stream after GPU kernel because we are using the data from the GPU.

PotatoPeeler3000:


ds58 (approved):

TomaszTB (approved):
So you forgot to synchronize after running a kernel... Rookie mistake

ihmc-rosie:
> So you forgot to synchronize after running a kernel... Rookie mistake

Well said lmao

PR #1021, opened 2025-10-06
feature/do-not-adjust-inertias-by-default

stefanfasano:
made it so inertias are not modified when loading urdf unless explicitelly done so by author

### DIFF
--- a/ihmc-model-file-loader/src/main/java/us/ihmc/modelFileLoaders/RobotDefinitionLoader.java
+++ b/ihmc-model-file-loader/src/main/java/us/ihmc/modelFileLoaders/RobotDefinitionLoader.java
@@ -72,7 +72,6 @@ public class RobotDefinitionLoader
                robotDefinition.addSubtreeJointsToIgnore(jointName);
             adjustJointLimitStops(robotDefinition, jointNameMap);
          }
-         adjustRigidBodyInterias(robotDefinition);
 
          if (removeCollisionMeshes)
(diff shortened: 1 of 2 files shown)


ihmc-model-file-loader/src/main/java/us/ihmc/modelFileLoaders/RobotDefinitionLoader.java
```diff
@@ -224,14 +222,25 @@ else if (prismaticJoint.getKpSoftLimitStop() <= 0.0 && prismaticJoint.getKdSoftL
...
-         {
-            RigidBodyDefinition successor = joint.getSuccessor();
-            successor.getMomentOfInertia().scale(100.0);
-         }
-      }
+      RigidBodyDefinition[] rigidBodyDefinitionsToAdjust = robotDefinition.getAllRigidBodies()
+                                                                          .stream()
+                                                                          .filter(rigidBodyDefinition -> isJointInNeedOfReducedGains(rigidBodyDefinition.getParentJoint()
+                                                                                                                                                        .getName()))
+                                                                          .toArray(RigidBodyDefinition[]::new);
```
stefanfasano, line 229: this will create garbage but I'm not sure how to do it in any way that doesn't create garbage. Also this should only be done once, not every tick

TomaszTB (approved):

james-p-foster (approved):

PR #1032, opened 2025-10-10
Feature/zed odometry

LuigiPenco93:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsToolboxModule/HumanoidKinematicsToolboxController.java
```diff
@@ -533,9 +532,15 @@ public void updateInternal()
       }
 
       super.updateInternal();
+
```
LuigiPenco93, line 535: added anchor body (Chest)  pose for Beyond Mimic

PR #1034, opened 2025-10-14
Feature/setup rl control state

stefanfasano:
made generic smooth transition state. added more external api options for addding state transitions in avatar multi thread factory


ihmc-humanoid-robotics/src/main/java/us/ihmc/humanoidRobotics/communication/packets/dataobjects/HighLevelControllerName.java
```diff
@@ -34,6 +34,12 @@ public enum HighLevelControllerName
    @RosEnumValueDocumentation(documentation = "whole body force control employing IHMC fast walking algorithms")
    QUICKSTER,
    EXTERNAL_TRANSITION_STATE,
+   @RosEnumValueDocumentation(documentation = "RL control state")
+   RL_CONTROL,
+   @RosEnumValueDocumentation(documentation = "Smooth transition state from RL to either freeze or stand prep.")
+   EXIT_RL,
+   @RosEnumValueDocumentation(documentation = "Smooth transition state from stand prep to RL.")
```
rjgriffin42: I think we can actually remove all the decorators here.

Purely from a formatting perspective, I would prefer EXTERNAL_TRANSITION_STATE to be right before EXTERNAL, as well.

reesep5 (approved):

PR #1035, opened 2025-10-14
fixed adding in the soft estop

rjgriffin42:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -386,6 +399,13 @@ private void run()
       if (isPaused.getBooleanValue())
          return;
 
+      if (blockingProvider != null && listenToBlockingCondition.getBooleanValue() && blockingProvider.getValue())
+      {
+         // publish the older control state.
+         hardwareCommunicationInterface.write(lowLevelOutputProcessor.getProcessedDesiredOutput(), lowLevelOutputProcessor.getMasterGain().getValue());
+         return;
```
rjgriffin42, line 406: this is what allows the controller execution to stay synchronized with the remote simulator. In normal practice, the blocking provider would be null, so this wouldn't be called.

LuigiPenco93 (approved):

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/HardwareCommunicationInterface.java
```diff
@@ -25,10 +25,14 @@ public interface HardwareCommunicationInterface
 
    boolean hasReceivedFirstState();
 
+   boolean hasNewStateMessage();
+
    default void addSoftEStopListener(YoVariableChangedListener listener)
    {
    }
 
+   void setSoftEStop(boolean softEStop);
```
stefanfasano: i don't see this implemented in AlexanderRos2HardwareCommunication. should we make this default to do nothing like the method above it?

stefanfasano: for alex as of now, there isn't a good way to soft e-stop anyway, our version of that would be the unservoQuickly trigger, which is located in the AvatarLowLevelOutputProcessor

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -165,6 +165,26 @@ public AvatarMultiThreadingFactory(DRCRobotModel robotModel,
                                                            freezeStateFactory);
    }
 
+   public void servoRobot()
```
stefanfasano, line 168: I don't think these methods make sense here. I'd suggest either:

- put them in the AvatarMultiThreadingManager
- make getters for the lowLevelOutputProcessor and hardwareCommunicationInterface

you can pretty easily interact with this stuff through the threading manager since it has access to all these, and when you run threadingFactory.buildThreadsAndThreadingManager, it returns the threading manager and you can ensure that it has actually been created

Ideally start/stop/join would also be moved from here to the threading manager, so that this factory is not interacted with at all once it actually builds the threading manager, but it is not like that yet because of the node spinning/stopping you see in start/stop

rjgriffin42: I did it this way because the factory isn't being used as a factory. Which would require restructuring things and the way it's called, which I didn't really want to do in the scope of this PR.

stefanfasano: other than the start/stop/destroy methods, it is being used as a factory. But I can see how that's out of scope, I think duncan and I are working on a PR that could tackle that stuff

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
stefanfasano: in the multi-threading factory you set up a blocking provider so it will not be null. Is listen to blocking condition false by default? we aren't setting that to true anywhere are we

rjgriffin42: it is by default, yes. It's getting set to true in the Simulation runnable.

PR #1038, opened 2025-10-14
Use manager instead of factory after creation.

calvertdw:


stefanfasano (approved):

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -326,11 +337,19 @@ public void join()
...
    public void stop()
    {
+      estimatorROS2Node.stopSpinning();
+      controllerROS2Node.stopSpinning();
       hardwareCommunicationInterface.stop();
    }
 
    public void destroy()
    {
+      estimatorROS2Node.destroy();
```
rjgriffin42, line 361: to match the old functionality, this needs to call `this.stop()` in the first line of the method.

calvertdw: I looked and it's effectively useless.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -596,39 +513,4 @@ public void addRequestableTransition(HighLevelControllerName currentControlState
...
-   {
-      return estimatorThread.get().getYoRegistry();
-   }
-
-   public YoGraphicGroupDefinition getEstimatorYoGraphics()
-   {
-      return estimatorThread.get().getSCS2YoGraphics();
-   }
-
-   public YoRegistry getControllerRegistry()
```
stefanfasano, line 610: i don't think that these methods for getting the controller and step generator registry/graphics registry got moved to the manager class. we could add them as is or just add getters for the threads themselves, although I wouldn't prefer the latter because if you have access to the threads directly you can call .run() on them in the wrong place/time

PR #1050, opened 2025-10-20
Setup logging to be done onboard or offboard based on the network interface its connected too

PotatoPeeler3000:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -155,6 +164,8 @@ public AvatarMultiThreadingFactory(DRCRobotModel robotModel,
       // Setup state estimator factory
       estimatorThreadFactory = createStateEstimatorFactory(robotModel, fullRobotModel, sensorReaderFactory);
 
+      estimatorThreadFactory.getYoGraphicsListRegistry();
```
stefanfasano: does this line do anything

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -234,9 +245,59 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
       // Set up the block to prevent execution whenever there is no new state message.
       threadingManager.get().setBlockingProvider(() -> !hardwareCommunicationInterface.hasNewStateMessage());
 
+      ArrayList<RegistrySendBufferBuilder> builders = new ArrayList<>();
+      builders.add(new RegistrySendBufferBuilder(rootRegistry, null));
```
stefanfasano: do we wanna add estimator thread yographics here (if there are any)? or is that not how that works

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -233,9 +245,59 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
...
       return threadingManager.get();
    }
 
+   public static boolean isEthernet0Connected()
+   {
+      // Since this is a new feature, I've added an override to force a given situation
+      if (OVERRIDE_AUTOMATIC_LOGGING)
+      {
+         //return true;
+         //return false;
```
ds58: ?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -233,9 +245,59 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
...
+      builders.add(new RegistrySendBufferBuilder(controllerThread.get().getYoVariableRegistry(), null, controllerThread.get().getSCS2YoGraphics()));
+      builders.add(new RegistrySendBufferBuilder(stepGeneratorThread.get().getYoVariableRegistry(), null, stepGeneratorThread.get().getSCS2YoGraphics()));
+      // TODO need to check if IK Streaming is started, would need to add this then
+      //      builders.add(new RegistrySendBufferBuilder(ikStreamingThread.get().getYoVariableRegistry(), null, ikStreamingThread.get().getSCS2YoGraphics()));
+
+      // Setup logger
+      if (isEthernet0Connected())
+      {
+         // Do nothing, logging on of board logger
+         System.out.println("Logging Status: data is being logged LOCALLY");
```
ds58: Can you use LogTools?

ds58 (changes requested):

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -234,9 +245,59 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
       // Set up the block to prevent execution whenever there is no new state message.
       threadingManager.get().setBlockingProvider(() -> !hardwareCommunicationInterface.hasNewStateMessage());
 
+      ArrayList<RegistrySendBufferBuilder> builders = new ArrayList<>();
+      builders.add(new RegistrySendBufferBuilder(rootRegistry, null));
+      builders.add(new RegistrySendBufferBuilder(controllerThread.get().getYoVariableRegistry(), null, controllerThread.get().getSCS2YoGraphics()));
+      builders.add(new RegistrySendBufferBuilder(stepGeneratorThread.get().getYoVariableRegistry(), null, stepGeneratorThread.get().getSCS2YoGraphics()));
```
stefanfasano: i would do a check for stepGeneratorThread.hasValue before doing this line, cause it is not guaranteed. It would be good practice to do the same for controller thread, but in reality it is always gonna not be null

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -234,9 +245,59 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
...
+      // TODO need to check if IK Streaming is started, would need to add this then
+      //      builders.add(new RegistrySendBufferBuilder(ikStreamingThread.get().getYoVariableRegistry(), null, ikStreamingThread.get().getSCS2YoGraphics()));
```
stefanfasano: can just do the check I suggested for the step generator thread, something like:

if (ikStreamingThread.hasValue())
  ...

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -65,6 +70,9 @@ public class AvatarMultiThreadingFactory
    public static final boolean RUN_AUTO_DIAGNOSTIC = false;
    private static final int ROS2_PRIORITY = 25;
 
+   // Override automatic logging
+   private static final boolean OVERRIDE_AUTOMATIC_LOGGING = false;
```
calvertdw: It's not clear what this override does. I would recommend FORCE_LOG_LOCALLY or FORCE_LOG_REMOTELY, whichever it is

calvertdw: Do you really need this or is it just for testing? Maybe you could remove it.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -234,9 +245,59 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
...
+         // Logging locally on the robot
+         System.out.println("Logging Status: data is being logged REMOTELY");
+         intraprocessYoVariableLogger = new IntraprocessYoVariableLogger(getClass().getSimpleName(), robotModel.getLogModelProvider(), builders, 100000, 0.01);
+         intraprocessYoVariableLogger.start();
+
+         threadingManager.get()
+                         .addPostEstimatorThreadRunnable(() -> intraprocessYoVariableLogger.update(estimatorThread.get()
+                                                                                                                  .getHumanoidRobotContextData()
+                                                                                                                  .getTimestamp()));
+      }
```
calvertdw, line 274: This isn't general enough to be in this very core class. I think it would be worth discussing with Stefan.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -234,9 +245,59 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
...
       return threadingManager.get();
    }
 
+   public static boolean isEthernet0Connected()
+   {
+      // Since this is a new feature, I've added an override to force a given situation
+      if (OVERRIDE_AUTOMATIC_LOGGING)
+      {
+         //return true;
+         //return false;
```
stefanfasano: should one of these not be commented out

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -234,9 +245,59 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
...
+      }
+
+      try
+      {
+         NetworkInterface ni = NetworkInterface.getByName("ethernet0");
```
stefanfasano: I don't love the idea of having the interface name hardcoded here, because this class should be hardware agnostic.

I think the way to get around this is to add a method to this factory that is basically like enableRemoteLogging() and then you can pass in the interface name, and then maybe a boolean to say whether to even check if ethernet is connected, or to just override that check and just do remote logging anyway.

Make the intraProcessYoVariableLogger an optional factory field, and populate it if the method I just described is called. Then when the "build()" method of this factory is called, check if the intraProcessYoVariableLogger .hasValue, and then add is as a post-estimator thread runnable

then in alexander, you can just call avatarMultiThreadFactory. enableRemoteLogging() before you call build() of the factory

calvertdw: I would check the interface outside of this factory and the factory method to add would be `logLocally()`. It seems remote logging is enabled by default and if its not connected it will just not do it by nature.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
PotatoPeeler3000: Nope, removed

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
PotatoPeeler3000: Naa this is the rootRegistry, which is not the estimator thread

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
PotatoPeeler3000: LogTools creates garbage

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
PotatoPeeler3000: It is just for testing, probably remove it after we test outside if things go smoothly

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
PotatoPeeler3000: Naa its just for debugging if stuff goes wrong

stefanfasano (changes requested):
main thing is the interface name being in here. would be happy to pair program a fix for this

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -74,6 +77,9 @@ public class AvatarMultiThreadingFactory
    // Hardware communication API
    private final HardwareCommunicationInterface hardwareCommunicationInterface;
 
+   // Logger stuff
+   private boolean logLocally;
```
stefanfasano: I would initialize this to false cause from what I can tell if it doesn't get set by the setter it'll be null

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -107,6 +113,7 @@ public class AvatarMultiThreadingFactory
    private final boolean useRealtimeThreads;
    private final boolean useMultiThreading;
    private final YoVariableServer yoVariableServer;
+   private IntraprocessYoVariableLogger intraprocessYoVariableLogger;
```
stefanfasano: i think you can just make this a local variable now

stefanfasano (commented):
looks much better. left two minor comments, after those if all still works i say it's good to go

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
calvertdw: primitives cant be null, this will be false, but i think it's smart to add the = false

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -233,6 +240,32 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
...
+      if (logLocally)
+      {
+         // Setup logger
+         ArrayList<RegistrySendBufferBuilder> builders = new ArrayList<>();
+         builders.add(new RegistrySendBufferBuilder(rootRegistry, null));
+         builders.add(new RegistrySendBufferBuilder(controllerThread.get().getYoVariableRegistry(), null, controllerThread.get().getSCS2YoGraphics()));
+         if (stepGeneratorThread.hasValue())
+         {
+            builders.add(new RegistrySendBufferBuilder(stepGeneratorThread.get().getYoVariableRegistry(), null, stepGeneratorThread.get().getSCS2YoGraphics()));
+         }
```
calvertdw: You could remove the braces to save some space

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -233,6 +240,32 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
...
+
+         // Logging locally on the robot
+         intraprocessYoVariableLogger = new IntraprocessYoVariableLogger(getClass().getSimpleName(), robotModel.getLogModelProvider(), builders, 100000, 0.01);
+         intraprocessYoVariableLogger.start();
+
+         threadingManager.get()
+                         .addPostEstimatorThreadRunnable(() -> intraprocessYoVariableLogger.update(estimatorThread.get()
+                                                                                                                  .getHumanoidRobotContextData()
+                                                                                                                  .getTimestamp()));
```
calvertdw, line 272:
```suggestion
         threadingManager.get().addPostEstimatorThreadRunnable(() -> 
           intraprocessYoVariableLogger.update(estimatorThread.get().getHumanoidRobotContextData().getTimestamp()));
```

calvertdw: i just think you could condense a bit if you break after the ->

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -233,6 +240,32 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
...
+         System.out.println("Logging Status: data is being logged LOCALLY");
```
calvertdw:
```suggestion
         LogTools.info("Logging locally.");
```
There's no need to shout :P

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -535,4 +568,9 @@ public void addSmoothTransitionState(String transitionName, HighLevelControllerN
    {
       controllerFactory.addCustomSmoothTransitionControlState(transitionName, transitionStateEnum, currentControlStateEnum, nextControlStateEnum);
    }
+
+   public void setLocalLogging(boolean logLocally)
```
calvertdw:
```suggestion
   public void setLogLocally(boolean logLocally)
```

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
PotatoPeeler3000: Removed

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
ds58: to be fair, it would be easier to see

calvertdw: Yeah but what about all the other messages? This feels like a debugging thing to all caps it

calvertdw (approved):

PR #1054, opened 2025-10-22
Moved the JVMStatisticsGenerator into open robotics and switched to using it. Added it to local logging as well

PotatoPeeler3000:


ihmc-robot-data-visualizer/src/main/java/us/ihmc/robotDataVisualizer/logger/JVMStatisticsGenerator.java
stefanfasano, line 1: how come this class shows up as being created instead of moved? was the one in the old location deleted?

stefanfasano (approved):

ihmc-robot-data-visualizer/src/main/java/us/ihmc/robotDataVisualizer/logger/JVMStatisticsGenerator.java
PotatoPeeler3000: It was created. The old class lives in the logger, its easier to edit in Open Robotics.

PR #1062, opened 2025-10-23
General-purpose xbox-to-csg plugin for RDX and SCS

stefanfasano:


ihmc-avatar-interfaces/build.gradle.kts
```diff
@@ -27,6 +27,9 @@ mainDependencies {
    api("us.ihmc:ihmc-simulation-toolkit:source")
    api("us.ihmc:ihmc-robot-data-visualizer:source")
    api("us.ihmc:robot-environment-awareness-visualizers:source")
+
+   api("com.badlogicgames.gdx-controllers:gdx-controllers-core:2.2.3")
+   api("com.badlogicgames.gdx-controllers:gdx-controllers-desktop:2.2.3")
```
calvertdw: I would add a comment and you don't need to declare core.
```suggestion
   api("com.badlogicgames.gdx-controllers:gdx-controllers-desktop:2.2.3") // for CSG plugin
```

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/joystickBasedJavaFXController/XBoxOneCSGPlugin.java
calvertdw, line 1: If you're going to keep this one around, I would suggest using it in SCS 2 and moving the GDX one to the RDX source set, so you don't have to add the Gradle dep above

calvertdw: or something... idk

stefanfasano: yeah i was wondering the same, I wasn't sure exactly where these should live. The only thing is I want this to be usable with both SCS and RDX. Is there any reason why the GDX one wouldn't work with SCS?

calvertdw (approved):

PR #1063, opened 2025-10-27
Got local logging working and tuned the realsense and height map for real world testing

PotatoPeeler3000:


TomaszTB (approved):

ds58 (changes requested):
Can we not move the LocalLoggingTools pls?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
stefanfasano, line 1: i think we did this stuff and it is on develop but this one is a bit out of date from that still

stefanfasano (approved):

PR #1064, opened 2025-10-27
Some tools for MPC; adjustments to `DoNothingBehaviour`

james-p-foster:


open-alexander/src/main/java/us/ihmc/openAlexander/parameters/controller/OpenAlexanderHighLevelControllerParameters.java
```diff
@@ -82,17 +82,17 @@ public List<GroupParameter<JointDesiredBehaviorReadOnly>> getDesiredJointBehavio
...
       List<String> allJoint = new ArrayList<String>();
       allJoint.addAll(jointMap.getSpineJointNamesAsStrings());
       allJoint.addAll(jointMap.getNeckJointNamesAsStrings());
       allJoint.addAll(jointMap.getArmJointNamesAsStrings());
       allJoint.addAll(jointMap.getLegJointNamesAsStrings());
-      behaviors.add(new GroupParameter<JointDesiredBehaviorReadOnly>("wholeBody", new JointDesiredBehavior(EFFORT), allJoint));
+      JointDesiredBehavior jointDesiredBehavior = new JointDesiredBehavior(EFFORT);
+      jointDesiredBehavior.setMaxPositionError(0.0);
+      jointDesiredBehavior.setMaxVelocityError(0.0);
+      behaviors.add(new GroupParameter<JointDesiredBehaviorReadOnly>("wholeBody", jointDesiredBehavior, allJoint));
```
james-p-foster, line 95: If you look at the default constructor of `JointDesiredBehaviour`, you can see  that max position error and max velocity error are set to infinity. We want to set them to zero so no control action is done. This causes shaking in do nothing state for MPC, and I believe the same was seen for RL

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/MultiBodySystemMissingTools.java
```diff
@@ -248,4 +248,34 @@ public static double computeSubTreeMass(RigidBodyReadOnly rootBody)
...
+   }
+
+   public static int getStateVectorSize(JointReadOnly[] joints)
+   {
+      return getConfigurationVectorSize(joints) + getVelocityVectorSize(joints);
+   }
+
+   public static int getStateDifferentialVectorSize(JointReadOnly[] joints)
+   {
+      return 2 * getVelocityVectorSize(joints);
```
Beomyeong, line 274: Does floatingBase have quaternion as rotation? Can we just multiply 2 simply? 
Or, do we use a frameVector3D to get differentialVector?

Beomyeong (approved):

ihmc-robotics-toolkit/src/main/java/us/ihmc/robotics/MultiBodySystemMissingTools.java
james-p-foster: Yeah the floating base rotation is represented by a quaternion, but this method is getting the size of the state differential, delta_x, not the state, x. The differential of the rotation is just represented by roll pitch yaw, so 3D, which justifies the use of this multiplication by 2 of the velocity size here.

LaughingCommit (approved):

open-alexander/src/main/java/us/ihmc/openAlexander/parameters/controller/OpenAlexanderHighLevelControllerParameters.java
stefanfasano: i think same for regular controller too

stefanfasano (approved):

rjgriffin42 (approved):

rjgriffin42:
My only comment is a deep, deep objection to your choice of spelling.

PR #1067, opened 2025-10-28
Feature/alex teleop 2

LuigiPenco93:


open-alexander/src/main/java/us/ihmc/openAlexander/AlexanderModelFactory.java
```diff
@@ -121,7 +121,7 @@ public RobotDefinition getControllerRobotDefinition()
    {
       if (controllerRobotDefinition == null)
       {
-         controllerRobotDefinition = getSCS1RobotDefinition();
+         controllerRobotDefinition = getSCS1RobotDefinition(true);
```
stefanfasano: should this always be true? shouldn't it be based on the robot version

open-alexander/src/main/java/us/ihmc/openAlexander/AlexanderModelFactory.java
```diff
@@ -83,7 +83,7 @@ public RobotDefinition getSCS1RobotDefinition()
                                                                          Arrays.asList(urdfParameters.getResourceDirectories()),
                                                                          getClass().getClassLoader(),
                                                                          urdfParameters.getURDFModelName(),
-                                                                         contactPointParameters,
+                                                                         createContactPoints ? contactPointParameters : null,
```
stefanfasano: why would we need this to be null? why not keep it the old way

open-alexander/src/main/java/us/ihmc/openAlexander/parameters/controller/OpenAlexanderMomentumOptimizationSettings.java
```diff
@@ -16,7 +16,7 @@ public class OpenAlexanderMomentumOptimizationSettings extends MomentumOptimizat
 {
    // TODO Parameter values copied from Valkyrie which has different weight.
    // TODO Needs tune up.
-   private final Vector3D linearMomentumWeight = new Vector3D(0.5, 0.5, 0.02);
```
LuigiPenco93, line 19: I need to revert this

PR #1068, opened 2025-10-29
Feature/desired interpolation

reesep5:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarLowLevelOutputProcessor.java
```diff
@@ -110,6 +139,39 @@ public void update(JointDesiredOutputListReadOnly unprocessedDesireds)
...
+   public void startDesiredsInterpolation()
+   {
+      interpolationTime.set(0);
+   }
+
+   private void interpolate()
+   {
+      double ratio = interpolationTime.getValueAsDouble() / interpolateDuration.getValueAsDouble();
+      interpolationRatio.set(MathTools.clamp(ratio, 0, 1));
+      if (ratio <= 1)
```
rjgriffin42: Is this ever false? You clamp it to be within this range the line above.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarLowLevelOutputProcessor.java
```diff
@@ -110,6 +139,39 @@ public void update(JointDesiredOutputListReadOnly unprocessedDesireds)
...
+
+            jointControlBlenders[i].computeAndUpdateJointControl(processedDesireds.getJointDesiredOutput(i),
+                                                                 previousJointDesireds,
+                                                                 currentJointDesireds,
+                                                                 ratio);
+         }
+      }
+      interpolationTime.add(updateDt);
+      if (interpolationTime.getDoubleValue() >= interpolateDuration.getDoubleValue())
+         previousDesireds.overwriteWith(processedDesireds);
```
rjgriffin42: If the above if statement is always true, then does this chunk have value to include? also, technically speaking, this should be done _before_ incrementing the interpolation time on line 170 to be consistently handled as the "else" condition to the above if statement.

open-alexander/src/main/java/us/ihmc/openAlexander/OpenAlexanderRobotModel.java
```diff
@@ -72,7 +72,7 @@ public class OpenAlexanderRobotModel implements DRCRobotModel
 
    private static final double DEFAULT_SIMULATE_DT = 0.0001;
    private static final double DEFAULT_ESTIMATE_DT = 0.001;
-   private static final double DEFAULT_CONTROL_DT = 0.003;
+   private static final double DEFAULT_CONTROL_DT = 0.02;
```
rjgriffin42: This should be removed from this PR, unless you want Nick to stab you in the heart.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarLowLevelOutputProcessor.java
```diff
@@ -110,6 +139,39 @@ public void update(JointDesiredOutputListReadOnly unprocessedDesireds)
...
+            JointDesiredOutputReadOnly previousJointDesireds = previousDesireds.getJointDesiredOutput(i);
+            JointDesiredOutputReadOnly currentJointDesireds = unprocessedDesireds.getJointDesiredOutput(i);
+
+            jointControlBlenders[i].computeAndUpdateJointControl(processedDesireds.getJointDesiredOutput(i),
+                                                                 previousJointDesireds,
+                                                                 currentJointDesireds,
+                                                                 ratio);
+         }
+      }
+      interpolationTime.add(updateDt);
```
rjgriffin42: would it make sense to also clamp this to updateDt, so it doesn't run wild?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarLowLevelOutputProcessor.java
reesep5: This was initially done to fix a bug with how some things were updating. I can move the resetting of the previous desireds back to the `startDesiredsInterpolation()` method

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarLowLevelOutputProcessor.java
reesep5: Changed the logic slightly, just trying to avoid any bouncing when the new desireds are fed in

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarLowLevelOutputProcessor.java
reesep5: It continuing to grow should not be a problem, since we have the ratio clamped. Also, it gets reset every time `startDesiredsInterpolation()` is called

rjgriffin42:
I'm noticing this isn't implemented for the Alex simulations, only in the AvatarMultiThreadFactory. Can we also implement it for Alex in the SCS2AvatarSimulationFactory?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -209,6 +209,8 @@ private HumanoidRobotControlTask setupControllerTaskAndThread(DRCRobotModel robo
          controllerTask.addCallbackPostTask(() -> yoVariableServer.update(controllerThread.getHumanoidRobotContextData().getTimestamp(),
                                                                           controllerThread.getYoVariableRegistry()));
 
+      controllerTask.addCallbackPostTask(lowLevelOutputProcessor::startDesiredsInterpolation);
```
stefanfasano, line 212: this only does the interpolation if it is actually enabled in the low-level output processor right?

reesep5: Yes, the yoboolean has to be checked for it to happen

LaughingCommit: yes, i believe so

rjgriffin42 (approved):

PR #1089, opened 2025-11-12
Move all filtering to SIL

reesep5:


rjgriffin42:
Approved, with the note that this should all get absorbed into `AlexStateEstimatorParameters`

rjgriffin42 (approved):

PR #1102, opened 2025-11-16
Added a listener to master gain so that we enter DO_NOTHING_BEHAVIOR when it reaches zero

PotatoPeeler3000:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -444,16 +444,17 @@ private HighLevelHumanoidControllerFactory createHighLevelControllerFactory(DRCR
...
-                                                               transitionToDoNothingCommand.setHighLevelControllerName(DO_NOTHING_BEHAVIOR);
-                                                               controllerCommandInputManager.submitCommand(transitionToDoNothingCommand);
-                                                            }
+                                                               controllerFactory.getRequestedControlStateEnum().set(DO_NOTHING_BEHAVIOR);
                                                          });
+         // Transition to DO_NOTHING when the robot is unservoed
+         lowLevelOutputProcessor.addMasterGainListener(change ->
+                                                       {
+                                                          if (lowLevelOutputProcessor.getMasterGain().getValue() == 0.0)
+                                                             controllerFactory.getRequestedControlStateEnum().set(DO_NOTHING_BEHAVIOR);
```
PotatoPeeler3000, line 456: If the master gain goes to zero, we set the controller state to DO_NOTHING. Safety case where if you unservo, and re-servoed with out checking what state you were in, the robot could jump to a position.

rjgriffin42 (approved):

reesep5 (approved):

PR #1110, opened 2025-11-18
Bugfix/fix capture region and rate limit

rjgriffin42:


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/stepAdjustment/ErrorBasedStepAdjustmentController.java
```diff
@@ -559,12 +559,12 @@ private void projectAdjustedStepIntoCaptureRegion(FramePoint3DReadOnly pointToPr
       {
          captureRegionInWorld.orthogonalProjection(adjustedSolution);
          computeBestReachabilityConstraintToUseWhenNotIntersecting();
-         getSelectedRegion().orthogonalProjection(adjustedSolution);
+         getSelectedReachableRegion().orthogonalProjection(adjustedSolution);
       }
       else
       {
          computeBestReachabilityConstraintToUseWhenIntersecting();
-         getSelectedRegion().orthogonalProjection(adjustedSolution);
```
rjgriffin42, line 567: I switched this back to what it was previously before the previous change. As written in the code, if returning the selected region, it returned the intersection between the reachable region and the capture point. Previously, the conditions would return this if the intersections were _valid_, so that there was an intersection, and if there were no intersections, it would return _just_ the reachable region. The change in the PR returned only the selected one, so that the projection failed. 

I changed this, so that if there's an intersection, it returns the selected reachable region, while refactoring the previous method to say "selectedReachableCaptureRegion", which is what is returned if these regions are valid.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/controller/ICPController.java
```diff
@@ -417,8 +417,7 @@ private void submitSolverTaskConditions(FrameConvexPolygon2DReadOnly supportPoly
 
       UnrolledInverseFromMinor_DDRM.inv(transformedGains, inverseTransformedGains);
 
-      solver.resetCoPFeedbackConditions();
-      solver.resetFeedbackDirection();
+      solver.resetFeedbackConditions();
```
rjgriffin42, line 420: I consolidated this logic into a single reset method, while also removing the limits.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/controller/ICPController.java
```diff
@@ -442,6 +441,11 @@ private void submitSolverTaskConditions(FrameConvexPolygon2DReadOnly supportPoly
 
          solver.setMaximumFeedbackRate(feedbackGains.getFeedbackPartMaxRate(), controlDT);
       }
+      else
+      {
+         solver.removeMaximumFeedbackMagnitude();
+         solver.removeFeedbackRateLimit();
+      }
```
rjgriffin42, line 448: Previously, what happened is that we wouldn't add the magnitude limits or rate limits if it was in a reset. However, those values were never getting reset inside the ICPControllerQPSolver. So it was applying whatever they were the last time they were set. So we were never actually removing these constraints.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/controller/ICPControllerQPSolver.java
```diff
@@ -357,6 +369,15 @@ private void reshape()
       solution.reshape(problemSize, 1);
    }
 
+   public void resetFeedbackConditions()
+   {
+      resetCoPFeedbackConditions();
+      resetCMPFeedbackConditions();
+      resetFeedbackDirection();
+      removeFeedbackRateLimit();
+      removeMaximumFeedbackMagnitude();
```
rjgriffin42, line 378: we were never calling "resetCMPFeedbackConditions". While this didn't cause any problems, it is a bug. The other two are the things that reset those rate limits and magnitude limits each time so they can be selectively enforced.

reesep5 (approved):

Beomyeong (approved):

PR #1117, opened 2025-11-20
Slowed down the sim hertz rate rom 0.0001 -> 0.0002 to make SCS2 run faster

PotatoPeeler3000:


PotatoPeeler3000:
There is a chanc
e this doesn't work at all but can't seem to test it without merging. Also it does break more of the tests. Looking into that.

PR #1124, opened 2025-11-26
Fix Ability Hand slider issue

TomaszTB:


TomaszTB:
Sliders got fixed in some other PR... can't remember which

PR #1130, opened 2025-12-05
set up some singularity filtering for the joint based contact switch

rjgriffin42:


Beomyeong (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitch.java
```diff
@@ -355,6 +360,7 @@ public JacobianBasedBasedTouchdownDetector(RigidBodyBasics foot,
          this.horizontalVelocityThreshold = horizontalVelocityThreshold;
          this.verticalVelocityThreshold = verticalVelocityThreshold;
          this.verticalVelocityHighThreshold = verticalVelocityHighThreshold;
+         this.jacobianDeterminantThreshold = jacobianDeterminantThreshold;
```
stefanfasano, line 363: since there are lots of thresholds in this class, maybe change this to be called "jacobianDeterrminantSingularityThreshold" to indicate this is for singularity checking, not touchdown detection checking

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitchFactory.java
```diff
@@ -112,6 +114,11 @@ public void setDefaultVerticalVelocityThreshold(double defaultVerticalVelocityTh
       this.defaultVerticalVelocityThreshold = defaultVerticalVelocityThreshold;
    }
 
+   public void setDefaultJacobianDeterminantThreshold(double defaultJacobianDeterminantThreshold)
```
stefanfasano: same comment as I left on the other class. This could be misinterpreted as the jacobian transpose force estimate threshold

PR #1133, opened 2025-12-10
Switched to using ihmc-alex-sdk

reesep5:
Not renaming openAlexander since we have discussed renaming it to Zulu or something similar


open-alexander/build.gradle.kts
```diff
@@ -12,7 +12,7 @@ ihmc {
 }
 
 mainDependencies {
-   api("us.ihmc:ihmc-alexander-sdk:0.3.4")
+   api("us.ihmc:ihmc-alex-sdk:source")
```
TomaszTB: We'll want to make a release of ihmc-alex-sdk and change this back to a released version, right?

TomaszTB (approved):

calvertdw (approved):

open-alexander/build.gradle.kts
reesep5: Yes, this was just to make sure the rename worked. I'm gonna do a release once it's finalized

open-alexander/src/main/java/us/ihmc/openAlexander/parameters/model/OpenAlexanderURDFParameters.java
stefanfasano, line 1: Is there also an OpenAlexanderXMLParameters that needs to be checked?

stefanfasano: if it does exist, I'm not even sure if it is used or needs to be used

stefanfasano (approved):

PR #1136, opened 2025-12-12
Fix misc behavior bugs and crashes when doing mustard pickup.

calvertdw:


TomaszTB (approved):
I'm kinda confused coz I thought many of these changes were already in the hand refactor branch but this looks good

PR #1137, opened 2025-12-12
[Bugfix] Simulation UI Fixes

TomaszTB:


ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/behaviorTree/scene/RDXBehaviorTreeScene.java
```diff
@@ -181,6 +181,9 @@ private static void renderPersistentDetection(PersistentDetectionStatusMessage m
 
    private void getRenderables(Array<Renderable> renderables, Pool<Renderable> pool, Set<RDXSceneLevel> sceneLevels)
    {
+      if (!sceneLevels.contains(RDXSceneLevel.VIRTUAL))
+         return;
```
TomaszTB, line 185: The simulated sensors were rendering the frame graphics as ground truth (see chunks of point cloud missing)

Image

ihmc-perception/src/main/java/us/ihmc/perception/RawImagePublisher.java
```diff
@@ -39,7 +41,10 @@ public RawImagePublisher(ROS2Node ros2Node)
       imageMessage = new ImageMessage();
       ros2Image = new Image();
 
-      nvJPEG = new CUDAJPEGProcessor();
+      if (CUDATools.hasNVJPEG())
```
TomaszTB, line 44: James had complained that he couldn't run the simulation UI due to not having nvJPEG installed. I believe this is the only spot nvJPEG is used in sim, so I made it fall back to CPU JPEG compression for publishing

calvertdw (approved):

PR #1161, opened 2026-01-07
remove the robot data visualizer project

rjgriffin42:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -298,8 +298,6 @@ public AvatarMultiThreadingManager buildThreadsAndThreadingManager()
                                                        ikStreamingThread.get().getSCS2YoGraphics()));
          }
 
-         builders.add(new RegistrySendBufferBuilder(jvmStatisticsGenerator.getYoRegistry(), null));
```
rjgriffin42, line 301: @stefanfasano The JVM statistics generator in robot data logger adds the registry directly to the yo variable server. We can do this differently if we want, I'm just not completely sure why it was done this way to begin with.

ds58 (approved):

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
stefanfasano: I recall this being added here by @PotatoPeeler3000 when he did the whole local logging integration. May want to check with him/test on robot to make sure removing this won't break stuff

PotatoPeeler3000: How would the JVM statistics get added to the log if we don't add it to the builders? @rjgriffin42 The YoVariableServer isn't running when we do local logging on the robot. So we need this line or we don't get the JVMStatistics on the local log.

rjgriffin42: Ok, do I need to call update on the visualizer inside the JVMStasticsGenerator?

PR #1166, opened 2026-01-08
RawImageToolsTest & Improvements

TomaszTB:


ds58 (approved):

ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
```diff
@@ -472,73 +474,6 @@ public void close()
       System.out.println("Closed " + getClass().getSimpleName());
    }
 
-   private void throwOnError(int errorCode) throws ZEDException
```
calvertdw, line 475: Isn't all this stuff useful?

calvertdw (approved):

ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
TomaszTB: Yeah, we put it in zed-java-api a while back but forgot to switch where it's used in the ZEDImageSensor. We're still calling `throwOnError`, just from ZEDTools which is in zed-java-api.

PR #1177, opened 2026-01-13
Cleanup/chunked map

PotatoPeeler3000:
Hey team, so the goal with this guy is to have some sort of data structure that  works as parts of the world. In this case I went with chunks because I thought it was intuitive as it acts the same as something in Minecraft, which I think everyone is familiar with.

In this case, we take the height map and convert it to a bunch of chunks in the world. And by seeing if a chunk has changed between its previous state, and current state. We can decide whether or not to publish it over the network. This allows us to only publish parts of the height map that are updated.

Additionally on the rendering side, we can leave old parts of the map rendered. Note that they may not be part of the data structure anymore but provide a visual to the surrounding environment. 

The reason I want to do this is because it allows the height map to go back to full resolution on the real robot. The current networking restrictions require a smaller sized height map, and a smaller resolution height map. Meaning visually we are losing some value. This does a better job at managing the networking.

To check this locally and play with it. Run the `AlexActiveMappingSimulationUI` and in the `HeightMapParameters` panel check the `Enable Chunked Map`. Then to turn on the visualizer, go into the `Height Map Visualizer`, and go to `Visualization Options`. Then check `Enable Chunked Map` and turn off the `Enable Height Map` option.

-----------------------------------------
Here is some of the ROS2Stats data:
image
^^^This image shows the networking with the current settings for the height map parameters on develop. That means the cell size is 4cm by 4cm. And the size of the height map is 3 meters by 3 meters.
image
^^^This image shows the networking with height map parmeters of 2cm by 2cm cell size. And 4 meter by 4 meter height map size. This is the case when most of the data is pretty much the same.
image
^^^This image has the 2cm by 2cm cell size, and 4 meter by 4 meter height map size. It shows the case when a lot of the data has different heights, this was the case where the robot was over rough terrain.

--------------------------------------------------
Additionally, the map can look like this on the UI now when walking around. Please note that this doesn't account for any drift in the world.
image


ihmc-interfaces/src/main/messages/ihmc_interfaces/perception_msgs/msg/ChunkMessage.msg
PotatoPeeler3000, line 1: Don't need everything to be 64 bits, 32 is fine

ihmc-interfaces/src/main/messages/ihmc_interfaces/perception_msgs/msg/ChunkMessage.msg
```diff
@@ -2,22 +2,22 @@
...
 # Cell size in meters of an individual cell
-float64 cell_size_in_meters
+float32 cell_size_in_meters
 
 # Cells per axis
 int32 cells_per_axis
 
 # List of heights, which correspond to the list of keys
-float32[<=5000] heights
+byte[<=20000] heights
```
PotatoPeeler3000, line 23: Had to increase this because we aren't using floats anymore, but bytes. And when compressing that with ".png" its not guarenteed that will fit into a fixed size. Generally the largest buffer size I've seen is 10,220 bytes. So 20,000 seems reasonable. Of course that all goes out the window if the chunk width changes.

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/HeightMapMessageTools.java
```diff
@@ -5,78 +5,13 @@
...
 import perception_msgs.msg.dds.HeightMapMessageForController;
 import us.ihmc.euclid.tuple3D.Point3D;
-import us.ihmc.idl.IDLSequence;
 import us.ihmc.idl.IDLSequence.Float;
 
-import java.nio.FloatBuffer;
-
 public class HeightMapMessageTools
 {
-   public static Mat unpackMessageToMat(ChunkMessage chunkMessage)
```
PotatoPeeler3000, line 19: Moved this to the ChunkMessageTools since I'm trying to get that tests and reliable

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/ChunkMessageTools.java
PotatoPeeler3000, line 1: These are the helper methods that pack and unpack the message. It gets compressed as well

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/ChunkTools.java
PotatoPeeler3000, line 1: Helper methods to go from a Mat object, to the Chunk object, and the opposite.

ihmc-perception/src/test/java/us/ihmc/perception/gpuMapping/TerrainMapExtractorTest.java
```diff
@@ -10,12 +10,6 @@
 import us.ihmc.euclid.tuple3D.Point3D;
 import us.ihmc.euclid.tuple3D.Vector3D;
 import us.ihmc.log.LogTools;
-import us.ihmc.perception.gpuMapping.HeightMapParameters;
```
PotatoPeeler3000, line 13: I guess those were unused lol

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/ChunkedMapManager.java
```diff
@@ -40,20 +38,20 @@ public class ChunkedMapManager
    public ChunkedMapManager(ROS2Node ros2Node, HeightMapParameters heightMapParameters)
    {
       this.heightMapParameters = heightMapParameters;
-      chunkMessage = new ChunkMessage();
       chunkMessagePublisher = ros2Node.createPublisher(PerceptionAPI.CHUNK);
+      chunkMessage = new ChunkMessage();
```
PotatoPeeler3000, line 45: Epic stuff

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/ChunkedMapManager.java
```diff
@@ -40,20 +38,20 @@ public class ChunkedMapManager
...
       if (heightMapParameters.getEnableChunkedMap())
       {
-         addHeightMap(latestHeightMap, heightMapCenterPoint, heightMapParameters.getWidthInMeters(), heightMapParameters.getCellSize());
+         addHeightMap(latestHeightMap, heightMapCenterPoint, heightMapParameters.getWidthInMeters(), (float) heightMapParameters.getCellSize());
          publishChunkedMap(chunkMessagePublisher);
       }
    }
 
-   public void addHeightMap(Mat heightMap, Point3DReadOnly heightMapCenter, double gridSize, double resolution)
+   public void addHeightMap(Mat heightMap, Point3DReadOnly heightMapCenter, double gridSize, float resolution)
```
PotatoPeeler3000, line 57: Floats please!

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/ChunkedMapManager.java
```diff
@@ -83,21 +80,16 @@ public void addHeightMap(Mat heightMap, Point3DReadOnly heightMapCenter, double
             chunks.add(chunk);
          }
       }
-
-      for (Chunk chunk : chunks)
```
PotatoPeeler3000, line 87: Chunks aren't using Mats in the object anymore, so this method got removed.

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/ChunkedMapManager.java
```diff
@@ -142,19 +134,19 @@ private void publishChunkedMap(ROS2Publisher<ChunkMessage> publisher)
...
+         ChunkMessageTools.toMessage(chunk, chunkMessage);
          publisher.publish(chunkMessage);
+         // After it's been published, we reset the dirty value so we don't keep publishing it even
+         // if it hasn't changed
+         chunk.setDirty(false);
       }
    }
 
-   private Collection<Chunk> getChunks()
+   public Collection<Chunk> getChunks()
```
PotatoPeeler3000, line 162: For testing

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/Chunk.java
```diff
@@ -25,76 +19,45 @@ public class Chunk
...
-
-      setDefaultHeight(cellsPerAxis);
-   }
-
-   /**
-    * We shouldn't have t odo this, but there is a memory problem where the new Mats are pointing to an old address or something
-    * and that is causing the new data to be filled with old data. Even though its a new object.
-    * So... TODO remove this hehe
-    */
-   private void setDefaultHeight(int cellsPerAxis)
```
PotatoPeeler3000, line 73: That method doesn't need to exist anymore, bug went away

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/Chunk.java
```diff
@@ -25,76 +19,45 @@ public class Chunk
...
     * Width of a chunk in meters
     */
-   public static final double CHUNK_WIDTH = 1.0;
-
-   private double originX;
-   private double originY;
-   private double cellSize;
+   public static final float CHUNK_WIDTH = 1.0f;
 
+   private float originX;
```
PotatoPeeler3000, line 24: Floats baby floatz

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/RDXChunkedMapRenderer.java
PotatoPeeler3000, line 1: The renderer seems to work, Not sure how to write a test for it though, I'd like too

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/RDXChunkedMapRenderer.java
```diff
@@ -132,24 +129,32 @@ public void destroy()
...
       private final Chunk chunk;
       private final RDXChunkRenderer renderer;
 
-      public ChunkRenderer(ChunkMessage chunkMessage)
+      public ChunkRenderer(Chunk chunk)
       {
          renderer = new RDXChunkRenderer();
-         chunk = new Chunk(chunkMessage);
+         this.chunk = new Chunk(chunk.getOriginX(), chunk.getOriginY(), chunk.getCellSize(), chunk.getCellsPerAxis());
+         map = new Mat(chunk.getCellsPerAxis(), chunk.getCellsPerAxis(), opencv_core.CV_32FC1);
```
TomaszTB, line 141: Ever close this Mat?

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/Chunk.java
```diff
@@ -137,14 +94,23 @@ public void setHeightAt(double xCord, double yCord, float height, double resolut
 
       if (i >= 0 && i < cellsPerAxis && j >= 0 && j < cellsPerAxis)
       {
-         chunkHeights[i * cellsPerAxis + j] = height;
+         float previousHeight = chunkHeights[i * cellsPerAxis + j];
+
+         if (previousHeight != height)
```
TomaszTB: Hmm floating point equality comparison... Generally don't wanna do that. I think the better comparison would be something like this. You can tune the tolerance however you like.

```suggestion
         if (Math.abs(previousHeight - height) > 1E-4)
```

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/Chunk.java
```diff
@@ -137,14 +94,23 @@ public void setHeightAt(double xCord, double yCord, float height, double resolut
...
-   public void commitHeightsToMat()
+   public void setHeight(int key, double height)
    {
-      FloatBuffer buffer = chunk.createBuffer();
-      buffer.put(chunkHeights);
+      if (key >= 0 && key < chunkHeights.length)
+      {
+         chunkHeights[key] = (float) height;
+      }
    }
```
TomaszTB, line 114: No marking dirty in this method?

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/ChunkedMapManager.java
TomaszTB: If it's just for testing and you don't want the method exposed outside of the package you could do package-private

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/ChunkMessageTools.java
```diff
@@ -0,0 +1,57 @@
...
+      messageToPack.setHashCodeOfChunk(chunk.hashCode());
+      messageToPack.setOriginX(chunk.getOriginX());
+      messageToPack.setOriginY(chunk.getOriginY());
+      messageToPack.setWidthInMeters(chunk.getWidthInMeters());
+      messageToPack.setCellSizeInMeters(chunk.getCellSize());
+      messageToPack.setCellsPerAxis(chunk.getCellsPerAxis());
+
+      // Get the heights in native memory as a 16U Mat
+      FloatPointer dataPointer = new FloatPointer(chunk.getChunkHeights());
+      Mat chunkMap16U = new Mat(chunk.getCellsPerAxis(), 2 * chunk.getCellsPerAxis(), opencv_core.CV_16U, dataPointer);
```
TomaszTB, line 41: That's so nasty. Who thought of this???

ihmc-perception/src/test/java/us/ihmc/perception/gpuMapping/worldmodel/ChunkMessageToolsTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+   private static final float WIDTH_IN_METERS = 1.0f;
+   private static final float CELL_RESOLUTION = 0.02f;
+
+   /**
+    * We are testing that when a message is packed and sent. That when its unpacked, its values are all the same.
+    * So we pack a {@link ChunkMessage} and check that when we unpack the {@link ChunkMessage} all the values match
+    * the original data.
+    */
+   @RepeatedTest(10)
```
TomaszTB, line 26: What's the benefit to using RepeatedTest here? Usually I use it to check for any race conditions, but all this code runs in a single thread, right?

ihmc-perception/src/test/java/us/ihmc/perception/gpuMapping/worldmodel/ChunkToolsTest.java
```diff
@@ -0,0 +1,124 @@
...
+import us.ihmc.perception.gpuMapping.worldModel.ChunkTools;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class ChunkToolsTest
+{
+   private final int iterations = 1000;
+   private final static float MILLISECOND_TOLERANCE = 1.0f;
+
+   @RepeatedTest(10)
```
TomaszTB, line 20: Same question as above

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/ChunkedMapManager.java
PotatoPeeler3000: Yep good call on this one

ihmc-perception/src/test/java/us/ihmc/perception/gpuMapping/worldmodel/ChunkMessageToolsTest.java
PotatoPeeler3000: Ehhh I don't think there is one lol, just for fun

ihmc-perception/src/test/java/us/ihmc/perception/gpuMapping/worldmodel/ChunkToolsTest.java
PotatoPeeler3000: No reason

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/ChunkMessageTools.java
PotatoPeeler3000: You showed me that lol, I copied it from the height map message tools class

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/Chunk.java
PotatoPeeler3000: Sounds good!

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/RDXChunkedMapRenderer.java
PotatoPeeler3000: Good call

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/ChunkMessageTools.java
TomaszTB: I can write some really ugly code sometimes :sweat_smile:

ihmc-perception/src/test/java/us/ihmc/perception/gpuMapping/worldmodel/ChunkMessageToolsTest.java
TomaszTB: My only concern is that it adds to CI time when it's not really necessary. Not really a massive problem tho

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/Chunk.java
calvertdw, line 1: I didn't realize this class already existed in our codebase, but I think it should have some domain qualification in the name. Is it specific to height maps? So HeightMapChunk. And prefix the other Chunk classes with HeightMap.

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/Chunk.java
```diff
@@ -137,14 +94,23 @@ public void setHeightAt(double xCord, double yCord, float height, double resolut
 
       if (i >= 0 && i < cellsPerAxis && j >= 0 && j < cellsPerAxis)
       {
-         chunkHeights[i * cellsPerAxis + j] = height;
+         float previousHeight = chunkHeights[i * cellsPerAxis + j];
+
+         if (Math.abs(previousHeight - height) > 1E-4)
+         {
+            // If we updated a height value that didn't match what we already had, we say the chunk is dirty, and needs to be published.
+            dirty = true;
```
calvertdw, line 102: :laughing: The verb "dirty" for this has always been hilarious to me

PotatoPeeler3000: True lol

ihmc-perception/src/main/java/us/ihmc/perception/gpuMapping/worldModel/Chunk.java
PotatoPeeler3000: That's a fair point, I'm gonna leave it for now cause I'm on my phone, but can come back to that when I do another rev

ihmc-perception/src/test/java/us/ihmc/perception/gpuMapping/worldmodel/ChunkMessageToolsTest.java
PotatoPeeler3000: The 10 tests take less then a second, so I'm not really worried about it

TomaszTB (approved):

ihmc-perception/src/test/java/us/ihmc/perception/gpuMapping/worldmodel/ChunkMessageToolsTest.java
calvertdw: image

calvertdw (approved):

PotatoPeeler3000:
Without any notion of throttling a given chunk to a specific publish rate:
image
And this is after the throttling per chunk:
image

PR #1180, opened 2026-01-15
Feature/changeable rate

rjgriffin42:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/BarrierSchedulerTools.java
```diff
@@ -1,33 +0,0 @@
-package us.ihmc.avatar;
```
rjgriffin42, line 1: Class was unused

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/AvatarStepGeneratorThread.java
```diff
@@ -36,7 +37,7 @@
 import us.ihmc.yoVariables.variable.YoDouble;
 import us.ihmc.yoVariables.variable.YoLong;
 
-public class AvatarStepGeneratorThread implements AvatarControllerThreadInterface
+public class AvatarStepGeneratorThread implements SCS2YoGraphicHolder
```
rjgriffin42, line 40: change so we don't have to implement a non-useful method

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/StepGeneratorTask.java
```diff
@@ -1,18 +1,84 @@
...
+import us.ihmc.commonWalkingControlModules.controllerCore.command.CrossRobotCommandResolver;
 import us.ihmc.robotModels.FullHumanoidRobotModel;
+import us.ihmc.robotics.time.ThreadTimer;
+import us.ihmc.yoVariables.variable.YoLong;
 
-public class StepGeneratorTask extends ControllerTask
+import java.util.ArrayList;
+import java.util.List;
+
+public class StepGeneratorTask extends HumanoidRobotControlTask
```
rjgriffin42, line 13: it's now different than the controller task.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/StepGeneratorTask.java
```diff
@@ -1,18 +1,84 @@
...
+      return super.initialize();
+   }
+
+   @Override
+   protected void execute()
    {
-      super(prefix, controllerThread, divisor, schedulerDt, masterFullRobotModel);
+      timer.start();
+      long schedulerTick = thread.getHumanoidRobotContextData().getSchedulerTick();
+      ticksBehindScheduled.set(schedulerTick - timer.getTickCount() * divisor);
```
Beomyeong, line 54: Is the reason this method does not include the below of `controllerTask.java`

```
int divisor = (int) Math.round(controllerThread.getCurrentDT() / schedulerDt);
      if (divisor != oldDivisor)
         setDivisor(divisor);
      ticksBehindScheduled.set(schedulerTick - timer.getTickCount() * oldDivisor);
```
because only ControllerDt will be changed and dt for stepGeneratorThread woudn't. Correct?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/drcRobot/DRCRobotModel.java
```diff
@@ -105,41 +98,53 @@ public default SideDependentList<HandModel> getHandModels()
...
-   public default SimulatedHandSensorReader createSimulatedHandSensorReader(OneDegreeOfFreedomJointHolder robot, List<String> fingerJointNames)
+
+   public abstract double getSimulateDT();
+
+   public abstract double getEstimatorDT();
+
+   default double getStepGeneratorDT()
    {
-      return new DefaultSimulatedHandSensorReader(robot, fingerJointNames);
+      return 10.0 * getFastestControllerDT();
```
Beomyeong, line 142: It seems DT for stepGeneratorThread will change. So, should we need to calculate the divisor in execute method from `StepGeneratorTask`?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/logProcessor/LogDataProcessorHelper.java
```diff
@@ -81,7 +81,7 @@ public LogDataProcessorHelper(DRCRobotModel model, SimulationConstructionSet scs
 
       inverseDynamicsStructure = new FullInverseDynamicsStructure(fullRobotModel.getElevator(), fullRobotModel.getPelvis(), fullRobotModel.getRootJoint());
 
-      controllerDT = model.getControllerDT();
+      controllerDT = model.getFastestControllerDT();
```
Beomyeong, line 84: I just wonder, if we set the DT as a fasetesControllerDT, and then if the controlDT changed to slower than this, 
The logData just save same data until controller which has slower DT is updated?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/stateTransitions/FeetLoadedToWalkingStandTransition.java
```diff
@@ -7,6 +7,7 @@
 import us.ihmc.robotics.robotSide.SideDependentList;
 import us.ihmc.robotics.sensors.FootSwitchInterface;
 import us.ihmc.robotics.sensors.ForceSensorDataHolderReadOnly;
+import us.ihmc.yoVariables.providers.DoubleProvider;
```
Beomyeong: Shouldn't the input value for the constructor be the DoubleProvider controlDT not double controlDT?

Beomyeong: If when this is constructed, double type for controlDT is input from the DoubleProvider, why we need to import this DoubleProvider here?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/stateTransitions/FeetLoadedTransition.java
```diff
@@ -10,6 +10,7 @@
 import us.ihmc.robotics.sensors.ForceSensorDataReadOnly;
 import us.ihmc.robotics.stateMachine.core.StateTransitionCondition;
 import us.ihmc.yoVariables.filters.SimpleMovingAverageFilteredYoVariable;
+import us.ihmc.yoVariables.providers.DoubleProvider;
```
Beomyeong, line 13: Same comment with the `FeetLoadedToWalkingStandTransition`

Beomyeong (approved):

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/logProcessor/diagnostic/DiagnosticAnalysisProcessor.java
```diff
@@ -36,12 +36,12 @@ public class DiagnosticAnalysisProcessor implements LogDataProcessorFunction
    private final YoVariableHolder logYoVariableHolder;
    private final double dt;
 
-   private DiagnosticParameters diagnosticParameters;
+   private final DiagnosticParameters diagnosticParameters;
 
-   public DiagnosticAnalysisProcessor(LogDataProcessorHelper logDataProcessorHelper, DRCRobotModel drcRobotModel)
+   public DiagnosticAnalysisProcessor(LogDataProcessorHelper logDataProcessorHelper, DRCRobotModel drcRobotModel, DiagnosticParameters diagnosticParameters)
```
stefanfasano, line 41: I understand making this final, but what is the need for adding the extra constructor argument?. Could add a second constructor that is the same as the old one and passes in DiagnosticParameters to this new constructor via drcRobotModel.get...

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/logProcessor/LogDataProcessorHelper.java
stefanfasano: this is a good point. Maybe getFastestControllerDT can return a double provider instead? that way it can always be the fastestcontrollerdt

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsStreamingToolboxModule/IKStreamingRTPluginFactory.java
```diff
@@ -293,6 +293,12 @@ public IKStreamingRTThread(String robotName,
 
       private long initialTime = -1L;
 
+      @Override
+      public double getCurrentDT()
+      {
+         return kinematicsStreamingToolboxController.getDT();
```
stefanfasano, line 299: should this be double provider?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsStreamingToolboxModule/KinematicsStreamingToolboxController.java
```diff
@@ -85,6 +86,7 @@ public KinematicsStreamingToolboxController(CommandInputManager commandInputMana
    {
       super(statusOutputManager, parentRegistry);
 
+      dt = parameters.getToolboxUpdatePeriod();
```
stefanfasano, line 89: should this be double provider? what if it changes in toolbox

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -201,11 +201,7 @@ private HumanoidRobotControlTask setupControllerTaskAndThread(DRCRobotModel robo
                                                                  YoVariableServer yoVariableServer)
    {
       // Set up Controller Task
-      int controllerDivisor = (int) Math.round(robotModel.getControllerDT() / masterThreadDt);
-      if (!Precision.equals(robotModel.getControllerDT() / masterThreadDt, controllerDivisor))
-         throw new RuntimeException("Controller DT must be multiple of master thread DT.");
-
-      controllerTask = new ControllerTask("Controller", controllerThread, controllerDivisor, masterThreadDt, masterFullRobotModel);
+      controllerTask = new ControllerTask("Controller", controllerThread, masterThreadDt, masterFullRobotModel);
```
stefanfasano: why does this one not need divisor but does need masterFullRobotModel

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -243,15 +239,14 @@ else if (!useRealtimeThreads && useMultiThreading)
...
-                                                                    FullHumanoidRobotModel masterFullRobotModel,
                                                                     YoVariableServer yoVariableServer)
    {
       // Set up Step Generator Task
       int stepGeneratorDivisor = (int) Math.round(robotModel.getStepGeneratorDT() / masterThreadDt);
       if (!Precision.equals(robotModel.getStepGeneratorDT() / masterThreadDt, stepGeneratorDivisor))
          throw new RuntimeException("Step generator DT must be multiple of master thread DT.");
 
-      stepGeneratorTask = new StepGeneratorTask("StepGenerator", stepGeneratorThread, stepGeneratorDivisor, masterThreadDt, masterFullRobotModel);
+      stepGeneratorTask = new StepGeneratorTask("StepGenerator", stepGeneratorThread, stepGeneratorDivisor, masterThreadDt);
```
stefanfasano: Why does this one need divisor but not need masterFullRobotModel. Ideally the task constructor is consistent between this, the controller task and the estimator task.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/AvatarControllerThread.java
```diff
@@ -169,6 +173,12 @@ public AvatarControllerThread(String robotName,
       ParameterLoaderHelper.loadParameters(this, robotModel, registry);
    }
 
+   @Override
+   public double getCurrentDT()
+   {
+      return controllerManager.getCurrentControlDT();
```
stefanfasano, line 179: add null check here

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/ControllerTask.java
```diff
@@ -17,24 +17,28 @@ public class ControllerTask extends HumanoidRobotControlTask
...
+   private final double schedulerDt;
+
+   public ControllerTask(String prefix,
+                         AvatarControllerThreadInterface controllerThread,
+                         double schedulerDt,
+                         FullHumanoidRobotModel masterFullRobotModel)
    {
-      super(divisor);
-      this.divisor = divisor;
+      super((int) Math.round(controllerThread.getCurrentDT() / schedulerDt));
```
stefanfasano, line 28: since there is nothing here that requires a double provider, or a changeable rate in the constructor, I would revert this back to the old way where you do the divisor calculation before construction and pass it in. That way you can do the check of whether the control dt is a whole multiple of scheduler dt. You can still use the .getCurrentDT() for that calculation, and you can still do the checks in the execute() method below as you currently have them

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/StepGeneratorTask.java
stefanfasano: if you decide to do what I was saying in the comment I left on ControllerTask, you can revert this back to before where is extends ControllerTask, and you can just override the execute() method cause that is the only real different part

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/BalanceManager.java
```diff
@@ -239,8 +240,10 @@ public BalanceManager(HighLevelHumanoidControllerToolbox controllerToolbox,
                          WalkingControllerParameters walkingControllerParameters,
                          CoPTrajectoryParameters copTrajectoryParameters,
                          SplitFractionCalculatorParametersReadOnly splitFractionParameters,
+                         double controlDT,
```
stefanfasano, line 243: should this be double provider?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/CenterOfMassHeightManager.java
```diff
@@ -64,6 +64,7 @@ public class CenterOfMassHeightManager implements SCS2YoGraphicHolder
 
    public CenterOfMassHeightManager(HighLevelHumanoidControllerToolbox controllerToolbox,
                                     WalkingControllerParameters walkingControllerParameters,
+                                    double controlDT,
```
stefanfasano, line 67: double provider?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/captureRegion/PushRecoveryControlModule.java
```diff
@@ -71,10 +72,13 @@ public class PushRecoveryControlModule implements SCS2YoGraphicHolder
...
-   public PushRecoveryControlModule(BipedSupportPolygons bipedSupportPolygons, HighLevelHumanoidControllerToolbox controllerToolbox,
-                                    WalkingControllerParameters recoveryControllerParameters, YoRegistry parentRegistry)
+   public PushRecoveryControlModule(BipedSupportPolygons bipedSupportPolygons,
+                                    HighLevelHumanoidControllerToolbox controllerToolbox,
+                                    WalkingControllerParameters recoveryControllerParameters,
+                                    double controlDT,
+                                    YoRegistry parentRegistry)
    {
-      controlDT = controllerToolbox.getControlDT();
+      this.controlDT = controlDT;
```
stefanfasano, line 81: should this be double provider or no? If so, maybe the toolbox can just return a double provider for current control dt

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controlModules/foot/FeetManager.java
```diff
@@ -83,16 +83,20 @@ public class FeetManager implements SCS2YoGraphicHolder
...
 
+   private final double controlDT;
+
    public FeetManager(HighLevelHumanoidControllerToolbox controllerToolbox,
                       WalkingControllerParameters walkingControllerParameters,
                       PIDSE3GainsReadOnly swingFootGains,
                       PIDSE3GainsReadOnly holdFootGains,
                       PIDSE3GainsReadOnly toeOffFootGains,
                       SideDependentList<RigidBodyControlManager> flamingoFootControlManagers,
+                      double controlDT,
```
stefanfasano, line 94: should this be double provider?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controlModules/foot/FootControlHelper.java
```diff
@@ -54,6 +56,7 @@ public FootControlHelper(RobotSide robotSide,
                             ExplorationParameters explorationParameters,
                             YoPartialFootholdModuleParameters footholdRotationParameters,
                             SupportStateParameters supportStateParameters,
+                            double controlDT,
```
stefanfasano, line 59: should this be double provider?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controlModules/foot/FootControlModule.java
```diff
@@ -123,6 +123,7 @@ public FootControlModule(RobotSide robotSide,
                             DoubleProvider minWeightFractionPerFoot,
                             DoubleProvider maxWeightFractionPerFoot,
                             DoubleProvider unloadedFinalRhoWeight,
+                            double controlDT,
```
stefanfasano, line 126: should this be double provider?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/HumanoidHighLevelControllerManager.java
```diff
@@ -138,7 +142,10 @@ public HumanoidHighLevelControllerManager(CommandInputManager commandInputManage
 
       pluginFactories.forEach(this::addControllerPluginFactory);
 
-      inertialParameterManager = managerFactory.getOrCreateInertialParameterManager();
+      //
```
stefanfasano, line 145: empty comment and deadspace

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/inverseKinematics/InverseKinematicsQPSolver.java
```diff
@@ -139,7 +140,7 @@ public void reset()
 
    private void addJointAccelerationRegularization()
    {
-      double factor = jointAccelerationRegularization.getDoubleValue() / (dt * dt);
+      double factor = jointAccelerationRegularization.getDoubleValue() / (dt.getValue() * dt.getValue());
```
stefanfasano, line 143: MathTools.Sqaure

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/momentumBasedController/optimization/InverseDynamicsQPSolver.java
```diff
@@ -276,7 +277,7 @@ public void resetRateRegularization()
 
    private void addJointJerkRegularization()
    {
-      double factor = dt * dt / jointJerkRegularization.getDoubleValue();
+      double factor = dt.getValue() * dt.getValue() / jointJerkRegularization.getDoubleValue();
```
stefanfasano, line 280: Mathtools.square

stefanfasano (approved):
Left some comments but I will approve anyway cause there's no way that I'm looking through all these again lol

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/drcRobot/DRCRobotModel.java
rjgriffin42: The goal is for the step generator rate to not change.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/StepGeneratorTask.java
rjgriffin42: correct

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/stateTransitions/FeetLoadedToWalkingStandTransition.java
rjgriffin42: no, this transition is pinned to a certain state. Each state has its own, unchanging DT, so you don't need a DoubleProvider. the import was added when I originally put it as a double provider but changed that back.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
rjgriffin42: Why would the constructor need to be consistent when the thread that it's executing is strongly typed and inconsistent?

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
rjgriffin42: In the code change of the controller task, it is computing its divisor internally, since it has a changing divisor.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/ControllerTask.java
rjgriffin42: I don't understand why you would advocate for that. To me, if I'm constructing the Task and passing in the divisor, I expect that divisor to be unchanging, which it's not.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/BalanceManager.java
rjgriffin42: no, since the balance manager only exists in one high level state.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/capturePoint/CenterOfMassHeightManager.java
rjgriffin42: no, since the height manager only exists in one high level state.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/captureRegion/PushRecoveryControlModule.java
rjgriffin42: no, and this class isn't even created currently.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controlModules/foot/FeetManager.java
rjgriffin42: no, since the feet manager only exists in one high level state.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controlModules/foot/FootControlHelper.java
rjgriffin42: no, since the helper only exists in one high level state.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controlModules/foot/FootControlModule.java
rjgriffin42: no, since the module only exists in one high level state.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/StepGeneratorTask.java
rjgriffin42: It's more complicated that that. There are chagnes to the underlying thread interface that this is implementing to make it a changeable rate thread. To me, that's out of scope of what the step generator was intended to do. So rather than further bastardize the Controller Task to work for the step generator, we can create a separate task, and prune out the stuff that isn't even needed.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/logProcessor/diagnostic/DiagnosticAnalysisProcessor.java
rjgriffin42: In an attempt to lightly clean the DRCRobotModel, I removed getting Diagnostic Parameters out of it. This processor is never actively used anywhere, and is never launched as part of the controller, even in its intended usage, so I wanted to get the Diagonstic Paramters out of the model.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsStreamingToolboxModule/IKStreamingRTPluginFactory.java
rjgriffin42: we talked about this in person, I dont' think we'll be running the plug in at a changeable rate, which would also be out of scope of this PR.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsStreamingToolboxModule/KinematicsStreamingToolboxController.java
rjgriffin42: we talked about this in person, I dont' think we'll be running the plug in at a changeable rate, which would also be out of scope of this PR.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/logProcessor/LogDataProcessorHelper.java
rjgriffin42: I'm not sure I understand what we're talking about. As it stands, the YoVariableServer publishes updates to the yo variables at its thread rate, which is pinned to the scheduler thread. I don't think making it change rate here makes a difference to that. This means that all the variables are logged at 1kHz. 

This class is for analyzing already existing logs, too, so I'm not sure what you're referring to is in scope of this class.

PR #1183, opened 2026-01-15
Change the TerrainMapExtractor to only use floats. We do this by implementing a Cholesky decompasition for solving the coefficients of the plane during least squares

PotatoPeeler3000:
Still want to go through and learn the math of all this, but vibe coding this seems pretty good. Did some performance changes and it was faster. Everything seems good in the algorithm, and all the tests pass.


ihmc-perception/src/main/resources/us/ihmc/perception/gpuMapping/TerrainMapExtractor.cu
```diff
@@ -23,6 +23,41 @@ extern "C"
 #define TOO_STEEP 4
 #define CLIFF_TOP 5
 
+__device__ bool choleskySolve3x3(const float A[9], const float b[3], float x[3])
```
PotatoPeeler3000: I need to go through and understand this math, this was straight vibe coded.

ihmc-perception/src/main/resources/us/ihmc/perception/gpuMapping/HeightMapUtils.cuh
```diff
@@ -20,5 +20,5 @@ __device__ int2 coordinate_to_indices(float2 coordinates, float2 center, float r
 
 __device__ int compute_center_index(float gridSize, float resolution)
 {
-    return round(0.5 * gridSize / resolution);
+    return round(0.5f * gridSize / resolution);
```
PotatoPeeler3000, line 23: This change gives a speed up of about 14%
image
image
What's happening here is the 0.5 is a double, so it requires using the FP64 register on the GPU, which is much slower then the regular ALU register which supports floats

ihmc-perception/src/main/resources/us/ihmc/perception/gpuMapping/TerrainMapExtractor.cu
```diff
@@ -184,34 +219,67 @@ __global__ void computeTerrainData(float *heightMap, size_t pitchHeightMap,
 
     // Perform best fit plane
     float3 normal = make_float3(0.0f, 0.0f, 1.0f);
-    double coefficients[3] = {0.0, 0.0, 0.0};
+    // Output coefficients (A, B, C)
```
PotatoPeeler3000, line 168: Changing all this to use floats gives about a 50% speed up of the kernel execution time.
image
image
In the bottom image the blue is the improved kernel without all the doubles. We still had some but that was in the HeightMapUtils.cuh file. That change isn't reflected in the speed up here.
The top image shows the microsecond execution time of the kernel, and how we went from like around 40 us to 18us.

ihmc-perception/src/test/java/us/ihmc/perception/gpuMapping/TerrainMapExtractorTest.java
```diff
@@ -143,9 +143,6 @@ public void testNormalCalculationForFlatInclinedTerrain()
       byte[] snapNormalYMap = terrainMapData.getSnapNormalYMap();
       byte[] snapNormalZMap = terrainMapData.getSnapNormalZMap();
 
-      LogTools.info("normal: " + terrainMapData.getNormal(0.3, 0.0));
```
PotatoPeeler3000, line 140: Cleanup

ihmc-perception/src/main/resources/us/ihmc/perception/gpuMapping/HeightMapUtils.cuh
PotatoPeeler3000: The blue is the version with 0.5f. So it doesn't have any of the double registers. The green one is the old slower one which is 14% slower

TomaszTB (approved):
Ay, as long as the tests pass 😆

ihmc-perception/src/main/resources/us/ihmc/perception/gpuMapping/TerrainMapExtractor.cu
stephenmcc: I think experimenting with faster matrix inversions is great, but this isn't very compatible with the existing function. If it's meant to replace invert3x3Matrix, that one should be removed. But I think what's better is to have functions with the same inputs that can be easily swapped. For example

`__device__ void invert3x3MatrixClosedForm(const float A[9], float Ainverse[9])` (existing one)
`__device__ void invert3x3MatrixCholesky(const float A[9], float Ainverse[9])` (this one)

Or if you prefer to have is go ahead and solve for a given x, then

`__device__ void solve3x3ClosedForm(const float A[9], const float b[3], float x[3])` (existing one)
`__device__ void solve3x3Cholesky(const float A[9], const float b[3], float x[3])` (this one)

stephenmcc (changes requested):

stephenmcc (approved):

PR #1189, opened 2026-01-20
Change OpenAlexander -> Zulu and Alexander -> Zulu for open robotics

PotatoPeeler3000:


.github/workflows/gradle-test-fast.yml
PotatoPeeler3000, line 1: This is for CI

simulation-construction-set-tools/src/main/java/us/ihmc/simulationConstructionSetTools/tools/CITools.java
PotatoPeeler3000, line 1: More CI stuff

zulu/build.gradle.kts
```diff
@@ -12,7 +12,6 @@ ihmc {
 }
 
 mainDependencies {
-   api("us.ihmc:ihmc-alex-sdk:0.3.7")
```
PotatoPeeler3000, line 15: Don't need to depend on this if we have the URDF and mesh's locally

zulu/gradle.properties
```diff
@@ -1,5 +1,5 @@
-kebabCasedName = open-alexander
-pascalCasedName = OpenAlexander
+kebabCasedName = zulu
```
PotatoPeeler3000, line 1: package rename

zulu/src/main/resources/us/ihmc/zulu/urdf/zulu-full-body.urdf
PotatoPeeler3000, line 1: Full body URDF for tests on CI

.github/workflows/gradle-test-fast.yml
rjgriffin42: Is the declaration of the subproject necessary, now that there's no dependency on the SDK?

zulu/src/main/java/us/ihmc/zulu/ZuluVersion.java
```diff
@@ -1,27 +1,28 @@
...
+import us.ihmc.zulu.parameters.model.ZULUURDFParameters;
 import us.ihmc.euclid.transform.RigidBodyTransform;
 import us.ihmc.robotics.robotSide.RobotSide;
 import us.ihmc.robotics.robotSide.SideDependentList;
 
 import java.util.Arrays;
 import java.util.Collection;
 
-public enum OpenAlexanderVersion implements AlexanderVersionInterface
+public enum ZuluVersion implements ZuluVersionInterface
```
rjgriffin42: is this necessary to be an interface anymore? That was done to allow inheritance between the two repos, but that got broken up. 

Honestly, if there's only a single version, this whole enum can be removed.

.github/workflows/gradle-test-fast.yml
PotatoPeeler3000: I don't know

PR #1196, opened 2026-01-22
Quick Footstep Planner

calvertdw:


ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/simplePlanners/QuickFootstepPlanner.java
```diff
@@ -0,0 +1,135 @@
+package us.ihmc.footstepPlanning.simplePlanners;
+
```
LuigiPenco93, line 2: there are quite a few hard coded parameters that could be extracted into a parameter class

calvertdw: My hope is that this produces walking plans suitable for any adult-sized humanoid, such that tuning parameters is not needed. If different functionality is desired, I'd like to take a look at a heuristic for it instead.

ds58 (approved):

LuigiPenco93 (approved):

PR #1203, opened 2026-01-23
Add demo for synced ZED svo and RealSense bag

ds58:


TomaszTB (approved):
Nice. Bonus points if you calculate the transform between the ZED and RealSense image origins to render them exactly where they should be in the demo

PR #1205, opened 2026-01-23
[Feature] Add Next Frame & Previous Frame Buttons to SVO Panel

TomaszTB:
[Screencast from 2026-01-23 16-51-07.webm](https://github.com/user-attachments/assets/be6c86e0-a822-4974-bbc5-632843d6fcdd)

### DIFF
--- a/ihmc-perception/src/main/java/us/ihmc/sensors/zed/ROS2ZEDSVOPlaybackSensor.java
+++ b/ihmc-perception/src/main/java/us/ihmc/sensors/zed/ROS2ZEDSVOPlaybackSensor.java
@@ -19,7 +19,7 @@ public class ROS2ZEDSVOPlaybackSensor extends ZEDSVOPlaybackSensor
       ros2.subscribeViaCallback(PerceptionAPI.ZED_SVO_PLAY, this::play);
       ros2.subscribeViaCallback(PerceptionAPI.ZED_SVO_SET_POSITION, position ->
       {
-         setCurrentPosition((int) position.getData() + 1);
+         setCurrentPosition((int) position.getData() + 1); // +1 is required to set the SVO to the correct frame. Otherwise, sets to 1 frame behind. Idk why.
          if (getGrabThread().getScheduled() == 0)
             getGrabThread().setScheduled(1);
       });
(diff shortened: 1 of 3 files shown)


ihmc-perception/src/main/java/us/ihmc/sensors/zed/ROS2ZEDSVOPlaybackSensor.java
```diff
@@ -1,57 +1,44 @@
...
       super(cameraID, zedModel, slDepthMode, svoFileName);
 
+      this.ros2 = ros2;
+
       ros2.subscribeViaCallback(PerceptionAPI.ZED_SVO_PAUSE, this::pause);
       ros2.subscribeViaCallback(PerceptionAPI.ZED_SVO_PLAY, this::play);
       ros2.subscribeViaCallback(PerceptionAPI.ZED_SVO_SET_POSITION, position ->
       {
-         setCurrentPosition((int) position.getData());
+         setCurrentPosition((int) position.getData() + 1);
```
TomaszTB: Don't ask why the +1. I don't know. It doesn't work without it though

ihmc-perception/src/main/java/us/ihmc/sensors/zed/ROS2ZEDSVOPlaybackSensor.java
```diff
@@ -1,57 +1,44 @@
...
 
    @Override
-   public void close()
+   public boolean grab()
    {
-      publishInfoThread.blockingKill();
-
-      super.close();
+      boolean result = super.grab();
+      publishSVOInfo();
```
TomaszTB, line 41: Simplified to publish after each grab instead of starting a thread when the sensor starts. I guess it wouldn't publish when the sensor is paused, but I haven't found that to be an issue. Let me know if I should fix this.

ds58 (approved):

ihmc-perception/src/main/java/us/ihmc/sensors/zed/ROS2ZEDSVOPlaybackSensor.java
ds58: I think I remember this quirk

PotatoPeeler3000: Can you leave a comment on that we need the +1 but don't fully understand it.

PotatoPeeler3000 (approved):

PR #1209, opened 2026-01-26
Fix for color in RealSense bag playback

ds58:


TomaszTB (approved):
You got some variables name *Ptr which I guess should be fully typed out to *Pointer but I don't think it's a big deal :shrug:

PR #1215, opened 2026-01-28
Add CUDA kernels for counting points in shapes and behavior condition action.

TomaszTB:


ihmc-perception/src/main/resources/us/ihmc/perception/cuda/ShapePointCounter.cu
```diff
@@ -52,3 +50,73 @@ __global__ void countPointsInSphere(unsigned short* depthImage,
...
+                // Closest point C on segment AB to P
+                float3 c = capsulePointA + (ab * t);
+
+                // Squared distance from P to C
+                float3 pc = worldFramePoint - c;
+                distance2 = normSquared(pc);
+            }
+
+            if (distance2 <= radius2)
+                atomicAdd(count, 1);
```
TomaszTB: This gets super float when there are a bunch of points because it's synchronized across the entire kernel. I think we could speed it up significantly with a bit of work

calvertdw (approved):

PR #1239, opened 2026-02-09
Speed up the registration kernel from 3.3 -> 3.0 microseconds

PotatoPeeler3000:
The speed increase of the kernel depends on the hardware. But the memory efficiency of the kernel is way better.

### DIFF
--- a/ihmc-perception/src/main/resources/us/ihmc/perception/gpuMapping/HeightMapExtractor.cu
+++ b/ihmc-perception/src/main/resources/us/ihmc/perception/gpuMapping/HeightMapExtractor.cu
@@ -218,7 +218,7 @@ __global__ void heightMapRegistrationKernel(const float *__restrict__ localMeanM
                                             const float globalMapCenterX,
 											const float globalMapCenterY,
                                             const float *__restrict__ groundToWorldTranslation,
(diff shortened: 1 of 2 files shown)


ihmc-perception/src/main/resources/us/ihmc/perception/gpuMapping/HeightMapExtractor.cu
```diff
@@ -218,56 +218,66 @@ __global__ void heightMapRegistrationKernel(const float *__restrict__ localMeanM
                                             const float globalMapCenterX,
 											const float globalMapCenterY,
                                             const float *__restrict__ groundToWorldTranslation,
```
TomaszTB, line 220: Something looks funky with the formatting here. Is it just GitHub?

TomaszTB (approved):
Yup, that's great

ihmc-perception/src/main/resources/us/ihmc/perception/gpuMapping/HeightMapExtractor.cu
PotatoPeeler3000: Yes

PR #1246, opened 2026-02-11
[Feature] VLM YOLO Image

TomaszTB:


ihmc-perception/src/main/java/us/ihmc/perception/detections/PersistentDetection.java
```diff
@@ -302,4 +302,23 @@ public ReferenceFrame getFilteredDetectionFrame()
    {
       return filteredDetectionFrame.getReferenceFrame();
    }
+
+   private static class IdProvider
```
calvertdw, line 306: whaattt, why can't you use an AtomicInt getAndIncrement and have it monotonically increasing?

TomaszTB: I'm worried the IDs will become quite long after a while of running the robot, and the images we send to the VLM will be polluted with numbers. Just seems better to have single digit IDs for the VLM

calvertdw (approved):

LuigiPenco93 (approved):

PR #1260, opened 2026-02-18
Re-enable behavior tree logging.

calvertdw:


ihmc-interfaces/src/main/messages/ihmc_interfaces/behavior_msgs/msg/BehaviorTreeYoDataMessage.msg
```diff
@@ -0,0 +1,42 @@
...
+uint8 number_of_persistent_detections
+
+uint8 number_of_scene_objects
+
+float32[3] scene_object_x
+float32[3] scene_object_y
+float32[3] scene_object_z
+float32[3] scene_object_yaw
+float32[3] scene_object_pitch
+float32[3] scene_object_roll
```
TomaszTB: Why not use a `geometry_msgs/Pose` message?

TomaszTB (approved):
Just one question

ihmc-interfaces/src/main/messages/ihmc_interfaces/behavior_msgs/msg/BehaviorTreeYoDataMessage.msg
calvertdw: cause its 56 bytes per pose and this is only 24. I'm very scared of the issue we were having so keeping it to an absolute minimum

PR #1271, opened 2026-02-25
Introduce re-calculation of glitch filter window sizes depending on c…

james-p-foster:
…ontroller / estimator thread dt and a nominal window length in seconds; preserve WBCC develop functionality where possible

### DIFF
--- a/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitch.java
+++ b/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitch.java
@@ -40,11 +40,18 @@ import java.util.List;
 public class JointTorqueBasedFootSwitch implements FootSwitchInterface
 {
    private static final double GRAVITY_Z = -9.81;
+   private static final int FORCE_THRESHOLD_WINDOW_SIZE = 2;
+   private static final int COP_THRESHOLD_WINDOW_SIZE = 3;
+   private static final int MINIMUM_WINDOW_SIZE = Math.min(FORCE_THRESHOLD_WINDOW_SIZE, COP_THRESHOLD_WINDOW_SIZE);
    private final YoRegistry registry;
 
    private final BooleanProvider useJacobianTranspose;
    private final JointTorqueBasedTouchdownDetector touchdownDetector;
    private final JacobianBasedBasedTouchdownDetector wrenchDetector;
+
+   private final YoInteger contactThresholdWindowSize;
+   private final YoDouble contactThresholdWindow;
+   private final DoubleProvider switchDT;
 
    private final MovingReferenceFrame soleFrame;
 
@@ -57,7 +64,8 @@ public class JointTorqueBasedFootSwitch implements FootSwitchInterface
                                      DoubleProvider contactForceThresholdLow,
                                      DoubleProvider contactForceThresholdHigh,
                                      DoubleProvider contactCoPThreshold,
-                                     YoInteger contactThresholdWindowSize,
+                                     double contactWindowLength,
+                                     DoubleProvider switchDT,
                                      BooleanProvider compensateGravity,
                                      DoubleProvider horizontalVelocityThreshold,
                                      DoubleProvider verticalVelocityThreshold,
@@ -95,6 +103,14 @@ public class JointTorqueBasedFootSwitch implements FootSwitchInterface
       }
 
       registry = new YoRegistry(jointToRead.getName() + getClass().getSimpleName());
+
+      this.switchDT = switchDT;
+
+      contactThresholdWindow = new YoDouble(namePrefix + "ContactThresholdWindow", registry);
+      contactThresholdWindow.set(contactWindowLength);
+      contactThresholdWindowSize = new YoInteger(namePrefix + "ContactThresholdWindowSize", registry);
+      contactThresholdWindowSize.set(Math.max(1, (int) Math.ceil(contactWindowLength / switchDT.getValue())));
+
       touchdownDetector = new JointTorqueBasedTouchdownDetector(namePrefix,
                                                                 jointToRead,
                                                                 true,
@@ -131,6 +147,9 @@ public class JointTorqueBasedFootSwitch implements FootSwitchInterface
    @Override
    public void update()
    {
+      // Account for any changes in the DT by updating the window size to maintain a consistent time window for contact detection.
+      contactThresholdWindowSize.set(Math.max(MINIMUM_WINDOW_SIZE, (int) Math.ceil(contactThresholdWindow.getDoubleValue() / switchDT.getValue())));
+
       touchdownDetector.update();
 
       wrenchDetector.calculate();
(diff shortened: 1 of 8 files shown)


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitchFactory.java
```diff
@@ -22,7 +20,7 @@ public class JointTorqueBasedFootSwitchFactory implements FootSwitchFactory
    private double defaultContactThresholdForceLow = 50.0;
    private double defaultContactThresholdForceHigh = 100.0;
    private double defaultContactCoPThreshold = 0.01;
-   private int defaultContactWindowSize = 10;
+   private double defaultContactWindow = 0.05; // slightly higher than previous WBCC-specific window of 0.03s
```
Beomyeong: Is it righr? It was 10 but the comment says it was 0.03

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitchFactory.java
```diff
@@ -67,9 +64,9 @@ public void setDefaultHigherContactThresholdTorque(double defaultHigherContactTh
       this.defaultHigherContactThresholdTorque = defaultHigherContactThresholdTorque;
    }
 
-   public void setDefaultContactWindowSize(int defaultContactWindowSize)
+   public void setDefaultContactWindow(double window)
```
Beomyeong: Also, isn't windowsize the int not the double?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitchFactory.java
james-p-foster: The previous window size was 10 ticks, now we have changed to specifying the window length in seconds -- this change is 0.05, which the comment says is "slightly higher" than the previous setting of 0.03s. I think this will be near enough to make no difference for the WBCC, but happy to hear your opinion otherwise.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitchFactory.java
james-p-foster: See previous comment -- this is now a window length in seconds, not a size in ticks

Beomyeong (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitch.java
```diff
@@ -95,6 +103,14 @@ public JointTorqueBasedFootSwitch(String namePrefix,
       }
 
       registry = new YoRegistry(jointToRead.getName() + getClass().getSimpleName());
+
+      this.switchDT = switchDT;
+
+      contactThresholdWindow = new YoDouble(namePrefix + "ContactThresholdWindow", registry);
```
rjgriffin42: Is this meant to be a duration? If so, could you change the name to reflect that it's a duration?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitch.java
```diff
@@ -95,6 +103,14 @@ public JointTorqueBasedFootSwitch(String namePrefix,
...
+      contactThresholdWindow.set(contactWindowLength);
+      contactThresholdWindowSize = new YoInteger(namePrefix + "ContactThresholdWindowSize", registry);
+      contactThresholdWindowSize.set(Math.max(1, (int) Math.ceil(contactWindowLength / switchDT.getValue())));
```
stefanfasano: should this just use MINIMUM_WINDOW_SIZE instead of 1? If that is the case, you could tie the calculation of this window size as a change listener for if the threshold window duration variable changes, however that would not capture if the DT changes (cause you can't add a listener to a double provider), so it is probably best to keep that in the update loop like you already have

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitch.java
```diff
@@ -391,10 +410,12 @@ public JacobianBasedBasedTouchdownDetector(RigidBodyBasics foot,
          verticalVelocity = new YoDouble(namePrefix + "VerticalVelocity", registry);
 
          isPastForceThresholdLow = new YoBoolean(namePrefix + "IsPastForceThresholdLow", registry);
-         isPastForceThresholdLowFiltered = new GlitchFilteredYoBoolean(namePrefix + "IsPastForceThresholdLowFiltered", registry, isPastForceThresholdLow, 2);
+         isPastForceThresholdLowFiltered = new GlitchFilteredYoBoolean(namePrefix + "IsPastForceThresholdLowFiltered", registry, isPastForceThresholdLow,
+                                                                       FORCE_THRESHOLD_WINDOW_SIZE);
```
stefanfasano, line 414: should FORCE_THRESHOLD_WINDOW_SIZE and COP_THRESHOLD_WINDOW_SIZE get the same treatment as contactThresholdWindowSize?

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitchFactory.java
stefanfasano: re-robert's point, I'd add the word "duration" to this variable. Also if you are adding a comment about how that is slightly higher than the controller window of 0.03, I'd also indicate it is also higher than estimator window of 0.01 since I think estimator made one of these objects as well

stefanfasano (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitch.java
james-p-foster: Renamed

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitch.java
james-p-foster: Good catch, it's now the min size. Like you say, I'm gonna keep it in the update loop

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitch.java
james-p-foster: I thought about this, but decided against it. They're already low, at 2-3 with the WBCC @ 333Hz. If we cut them down to e.g MPC rate at 50Hz, these will <1, so not possible. Therefore, I left them, and made sure that the overall window size was no less than the largest of these two.

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitchFactory.java
james-p-foster: Appended duration and added estimator window comment, important to note that this variable is always overwritten in `AlexWalkingControllerParameters` and `AlexStateEstimatorParameters` tho so all good

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/sensors/footSwitch/JointTorqueBasedFootSwitchFactory.java
james-p-foster: Appended `Duration`

PR #1276, opened 2026-03-02
Fixing broken Alex tests

Beomyeong:
AlexEndToEndChestTrajectoryMessageTest, added overrides returning 0 for:
      - getExpectedTaskspaceWaypointsAfterQueueWithWrongPreviousId()
      - getExpectedTaskspaceWaypointsAfterQueueWithUsingDifferentTrajectoryFrameWithoutOverride()
      - getExpectedTaskspaceWaypointsAfterStopAllTrajectory()


ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/controllerAPI/EndToEndChestTrajectoryMessageTest.java
```diff
@@ -79,6 +79,10 @@ public abstract class EndToEndChestTrajectoryMessageTest implements MultiRobotTe
 {
    private static final SimulationTestingParameters simulationTestingParameters = SimulationTestingParameters.createFromSystemProperties();
    private static final double EPSILON_FOR_DESIREDS = 5.0e-4;
+   /**
+    * SO3TrajectoryMessage.taskspace_trajectory_points is IDL-bounded to 50.
+    */
+   private static final int MAX_SO3_TRAJECTORY_POINTS_PER_MESSAGE = 50;
```
PotatoPeeler3000, line 85: Good find!

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/controllerAPI/EndToEndChestTrajectoryMessageTest.java
```diff
@@ -934,6 +938,21 @@ public Axis3D getAxisToMoveForQueuedMessages()
       return Axis3D.Y;
    }
 
+   public int getExpectedTaskspaceWaypointsAfterQueueWithWrongPreviousId()
```
PotatoPeeler3000, line 941: I think we can get rid of these, is there ever a case where 1 would be expected? I don't think so.

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/controllerAPI/EndToEndChestTrajectoryMessageTest.java
```diff
@@ -1218,7 +1237,10 @@ else if (getAxisToMoveForQueuedMessages() == Axis3D.Z)
          defaultControlMode = RigidBodyControlMode.JOINTSPACE;
       }
       assertEquals(defaultControlMode, EndToEndTestTools.findRigidBodyControlManagerState(chest.getName(), simulationTestHelper));
-      EndToEndTestTools.assertTotalNumberOfWaypointsInTaskspaceManager(chest.getName(), prefix, 1, simulationTestHelper);
+      EndToEndTestTools.assertTotalNumberOfWaypointsInTaskspaceManager(chest.getName(),
+                                                                       prefix,
+                                                                       getExpectedTaskspaceWaypointsAfterQueueWithWrongPreviousId(),
```
PotatoPeeler3000, line 1242: Can't we just use '0' here in the tests? This behavior is part of the WBCC which isn't dependent on the robot parameters

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/controllerAPI/EndToEndChestTrajectoryMessageTest.java
```diff
@@ -1298,7 +1320,10 @@ public void testQueueWithUsingDifferentTrajectoryFrameWithoutOverride() throws E
          defaultControlMode = RigidBodyControlMode.JOINTSPACE;
       }
       assertEquals(defaultControlMode, EndToEndTestTools.findRigidBodyControlManagerState(chest.getName(), simulationTestHelper));
-      EndToEndTestTools.assertTotalNumberOfWaypointsInTaskspaceManager(chest.getName(), prefix, 1, simulationTestHelper);
+      EndToEndTestTools.assertTotalNumberOfWaypointsInTaskspaceManager(chest.getName(),
+                                                                       prefix,
+                                                                       getExpectedTaskspaceWaypointsAfterQueueWithUsingDifferentTrajectoryFrameWithoutOverride(),
```
PotatoPeeler3000, line 1325: Comment above, should just use '0' and remove this helper method

zulu/src/test/java/us/ihmc/zulu/controllerAPI/ZuluEndToEndChestTrajectoryMessageTest.java
```diff
@@ -35,6 +35,24 @@ public Axis3D getAxisToMoveForQueuedMessages()
       return Axis3D.Z;
    }
 
+   @Override
```
PotatoPeeler3000, line 38: I don't think we need these, just use '0' in the assert of the test

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/roughTerrainWalking/EndToEndCinderBlockFieldTest.java
```diff
@@ -122,7 +122,7 @@ public void testWalkingOverCinderBlockField() throws Exception
       simulationTestHelper.publishToController(footsteps);
 
       WalkingControllerParameters walkingControllerParameters = getRobotModel().getWalkingControllerParameters();
-      success = simulationTestHelper.simulateNow(EndToEndTestTools.computeWalkingDuration(footsteps, walkingControllerParameters) + 1.0);
+      success = simulationTestHelper.simulateNow(EndToEndTestTools.computeWalkingDuration(footsteps, walkingControllerParameters) + 10.0);
```
PotatoPeeler3000, line 125: Is this test fixed when you increase it to 10.0 seconds? It was still broken for me.

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/controllerAPI/EndToEndChestTrajectoryMessageTest.java
Beomyeong: Alex set the defualt control mode as JointSpace mode, so it would be '0'. Also the reason for me to use that getter not using simple number '0' is I want to input what the meaning of that is. 
Previous way '0', or '1' looks simple, but it doesn't introduce correctly what that number means.

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/controllerAPI/EndToEndChestTrajectoryMessageTest.java
Beomyeong: Same reason with above comment

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/controllerAPI/EndToEndChestTrajectoryMessageTest.java
Beomyeong: Same reason with above comment

zulu/src/test/java/us/ihmc/zulu/controllerAPI/ZuluEndToEndChestTrajectoryMessageTest.java
Beomyeong: Same reason with above comment. Isn't that one the universal / common class for the other robots? I wrote down this for some case that other robot use the taskspace control is the defualt control.

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/roughTerrainWalking/EndToEndCinderBlockFieldTest.java
Beomyeong: With this and there are changed in Alex repository to change the swing foot height and the  ratio for the stepping point. 
There are two causes for this broken. 
1. Too small simulation time to get stable. 
2. There's a inclided cinderblock. But the center height of that cinder block is similar height with the next cinderbloc (Robot support foot is there). So, the siwng height isn't enought high, becuase the API consider the ground is similar to flat. But the higher part of the inclided cinderblock blocking the swing foot. .

ihmc-avatar-interfaces/src/test/java/us/ihmc/avatar/controllerAPI/EndToEndChestTrajectoryMessageTest.java
PotatoPeeler3000: Is there ever a case where the value of the assert isn't `0`? I thought it always has to be?

PotatoPeeler3000: Also we don't need getters to explain the variable. We can have something like:
```
int expectedTaskspaceWaypointsAfterBadQueueID = 0;
assert(bla, expectedTaskspaceWaypointsAfterBadQueueID, bla, bla);
```

rjgriffin42: I guess this implementation is more future proof, as it enables a robot that can be non-zero, and you can override this return in the child test. We do that later thing a long.

rjgriffin42 (approved):

PR #1280, opened 2026-03-03
Quick Footstep Planner: Support turning, side stepping, and waypoints with new algorithms.

calvertdw:


ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/simplePlanners/QuickFootstep.java
ds58, line 1: Can we get away without it? Also, the naming is problematic. Should either be `QuickFootstepPlannerFootstep` or just `Footstep` because nothing in here is specific to `QuickFootstepPlanner`

ds58 (changes requested):

ihmc-footstep-planning/src/main/java/us/ihmc/footstepPlanning/simplePlanners/QuickFootstep.java
TomaszTB: Maybe it could be a private inner class of the `QuickFootstepPlanner`? Also, if you want to make it small, I think it could be a record class.

ds58: An inner private class would be good but it also has usage in the test

calvertdw: Is it really that problematic? The field are specific to this planner, by the inclusion of swing distance, there's other footstep classes like this too

Image

TomaszTB (approved):

PR #1281, opened 2026-03-05
Set reference frame of camera as soon as we grab the frame

PotatoPeeler3000:


ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
```diff
@@ -313,7 +316,7 @@ protected boolean grab()
          }
 
          throwOnError(returnCode);
-         lastGrabTime = Instant.now();
+         lastGrabTime = grabTime;
```
PotatoPeeler3000, line 319: By setting the grab time as soon as we get the image, the time is more correct

ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
```diff
@@ -290,6 +290,9 @@ protected boolean grab()
       {
          // Grab images now
          returnCode = sl_grab(cameraID, zedRuntimeParameters);
+         RigidBodyTransform leftSensorTransformAtGrab = leftSensorFrame.getTransformToWorldFrame();
+         RigidBodyTransform rightSensorTransformAtGrab = rightSensorFrame.getTransformToWorldFrame();
+         Instant grabTime = Instant.now();
```
PotatoPeeler3000, line 295: Need to set these here because this is when we grab the image, if we wait till we pack the RawImage to set the transform, it will have changed already.

ihmc-perception/src/main/java/us/ihmc/sensors/zed/ZEDImageSensor.java
```diff
@@ -290,6 +290,9 @@ protected boolean grab()
       {
          // Grab images now
          returnCode = sl_grab(cameraID, zedRuntimeParameters);
+         RigidBodyTransform leftSensorTransformAtGrab = leftSensorFrame.getTransformToWorldFrame();
+         RigidBodyTransform rightSensorTransformAtGrab = rightSensorFrame.getTransformToWorldFrame();
```
PotatoPeeler3000, line 294: `.getTRansformToWorldFrame()` creates a new object, so that seems fine. It also seems like it will generally be faster then `getTransformToRoot()`. So that's what I used

TomaszTB: > It also seems like it will generally be faster then `getTransformToRoot()`

It's slower. `getTransformToWorldFrame()` does a couple extra checks then calls `getTransformToRoot()` internally. I think creating a new transform and calling `getTransformToRoot()` would be faster. But that's a minor optimization that doesn't really matter

TomaszTB (approved):

PR #1293, opened 2026-03-11
fix hand robot config data in kst

LuigiPenco93:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsToolboxModule/KinematicsToolboxController.java
```diff
@@ -434,24 +438,26 @@ public class KinematicsToolboxController extends ToolboxController implements SC
...
-    * @param parentRegistry          registry to attach {@code YoVariable}s to.
+    * @param commandInputManager        the message/command barrier used by this controller. Submit
+    *                                   messages or commands to be processed to the
+    *                                   {@code commandInputManager} from outside the controller.
+    * @param statusOutputManager        the output interface used by this controller.
+    * @param rootJoint                  the underactuated floating root joint of the multi-body system.
+    *                                   Can be {@code null} in the case all the joints are actuated.
+    * @param desiredOneDoFJoints        the actuated joints of the system. The inverse kinematics will
+    *                                   only use these joints during the optimization.
+    * @param desiredOneDoFJointIndices  the indices of the actuated joints of the system.
```
calvertdw, line 449: May be null, right? Might want to add that if so
```suggestion
    * @param desiredOneDoFJointIndices  the indices of the actuated joints of the system.
    *                                                             Can pass null if there are no fingers in the robot model.
```

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/networkProcessor/kinematicsToolboxModule/KinematicsToolboxHelper.java
```diff
@@ -168,13 +169,21 @@ public static void setRobotStateFromControllerCoreOutput(ControllerCoreOutputRea
     */
    public static void setRobotStateFromRobotConfigurationData(RobotConfigurationData robotConfigurationData,
                                                               FloatingJointBasics rootJoint,
-                                                              OneDoFJointBasics[] oneDoFJoints)
+                                                              OneDoFJointBasics[] oneDoFJoints,
+                                                              List<Integer> oneDoFJointsIndices)
    {
       TFloatArrayList newJointAngles = robotConfigurationData.getJointAngles();
 
+      List<Float> filteredJoints = new ArrayList<>();
```
calvertdw: I would use array to avoid allocating. not super necessary. not sure if KST is allocationfree
```suggestion
      float[] filteredJoints = new float[oneDoFJointsIndices.size()];
```

ihmc-robot-models/src/main/java/us/ihmc/robotModels/FullRobotModelUtils.java
```diff
@@ -24,6 +24,32 @@ public static OneDoFJointBasics[] getAllJointsExcludingHands(FullHumanoidRobotMo
       return joints.toArray(new OneDoFJointBasics[joints.size()]);
    }
 
+   public static List<Integer> getAllJointsExcludingHandsIndices(FullHumanoidRobotModel model)
```
calvertdw, line 27: idk, maybe it would be better to use int[] so it could be used in allocation free places?

calvertdw (approved):

PR #1300, opened 2026-03-16
[Cleanup] ROS2HumanoidFrames

TomaszTB:
I spent some more time trying to reproduce the `NotARotationMatrixException` we see with the base footprint frame, but couldn't reproduce it. 

I did clean up some code in the process and fix the base footprint frame to actually match the specification given by ROS 2.

### DIFF
--- a/ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/drcRobot/CommunicationsSyncedRobotModel.java
+++ b/ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/drcRobot/CommunicationsSyncedRobotModel.java
@@ -139,7 +139,6 @@ public abstract class CommunicationsSyncedRobotModel
    {
       try
       {
-         fullRobotModel.getElevator().updateFramesRecursively();
          referenceFrames.updateFrames();
       }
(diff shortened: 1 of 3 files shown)


calvertdw (approved):

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/drcRobot/CommunicationsSyncedRobotModel.java
```diff
@@ -139,7 +139,6 @@ private void updateFramesForFullRobotModel()
    {
       try
       {
-         fullRobotModel.getElevator().updateFramesRecursively();
```
ds58, line 142: How does this change affect things?

TomaszTB: Oh this wasn't meant to be part of this PR but I guess it may as well be 🤷

The referenceFrames.updateFrames() call updates the fullRobotModel internally. So it was updating the fullRobotModel twice.

ds58 (approved):

PR #1304, opened 2026-03-17
[Feature] ControllerAPI Low Frequency Topics

TomaszTB:
Add low frequency output topics to the ControllerAPI. Each controller output now has two topics: a normal unthrottled topic and a throttled low frequency topic. 

The low frequency topics are simply the normal topic with `/lf` appended. For example, the normal topic
`/ihmc/alex/humanoid_control/output/footstep_queue_status`
has its low frequency alternative as 
`/ihmc/alex/humanoid_control/output/footstep_queue_status/lf`. 

Currently I set an arbitrary publish frequency limit to 50hz (matching RCD publish frequency). Let me know if you'd like to see a different number/have it parametrized somehow.

Tested this on the robot using the shakeout behavior, a tiny bit of continuous hiking, and control ring walking.

### DIFF
--- a/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controllerAPI/input/ControllerNetworkSubscriber.java
+++ b/ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controllerAPI/input/ControllerNetworkSubscriber.java
@@ -4,6 +4,7 @@ import controller_msgs.msg.dds.InvalidPacketNotificationPacket;
 import ihmc_common_msgs.msg.dds.MessageCollection;
 import ihmc_common_msgs.msg.dds.MessageCollectionNotification;
 import us.ihmc.commonWalkingControlModules.controllerAPI.input.MessageCollector.MessageIDExtractor;
+import us.ihmc.commons.thread.Throttler;
 import us.ihmc.communication.controllerAPI.CommandInputManager;
 import us.ihmc.communication.controllerAPI.ControllerAPI;
 import us.ihmc.communication.controllerAPI.MessageUnpackingTools.MessageUnpacker;
@@ -34,6 +35,7 @@ import java.util.concurrent.atomic.AtomicReference;
  */
 public class ControllerNetworkSubscriber
 {
+   private static final double LOW_FREQUENCY_PUBLISH_RATE = 50.0;
    private static final boolean DEBUG = false;
 
    /** The input API to which the received messages should be submitted. */
@@ -58,6 +60,7 @@ public class ControllerNetworkSubscriber
     * communication thread.
     */
    private final Map<Class<? extends Settable<?>>, ROS2Publisher<?>> statusMessagePublisherMap = new HashMap<>();
+   private final Map<Class<? extends Settable<?>>, ThrottledROS2Publisher<?>> lowFrequencyStatusMessagePublisherMap = new HashMap<>();
 
    private final ROS2Node ros2Node;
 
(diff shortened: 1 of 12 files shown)


ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controllerAPI/input/ControllerNetworkSubscriber.java
```diff
@@ -34,6 +35,7 @@
  */
 public class ControllerNetworkSubscriber
 {
+   private static final double LOW_FREQUENCY_PUBLISH_RATE = 50.0;
```
ds58, line 38: Can we make it 30.0 and if required, increase it?

ds58 (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/controllerAPI/input/ControllerNetworkSubscriber.java
TomaszTB: Sure. I've also considered going 10th of controller frequency, which would be 33hz. We use that for some other stuff.. I don't remember what though. Could that be better?

calvertdw: would prefer 50

calvertdw: but idc too much

calvertdw: im not sure about interactions of things if you go too low. code shouldn't care but sometimes it does

PotatoPeeler3000: 50 please

PotatoPeeler3000 (approved):

LuigiPenco93 (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/plugin/HumanoidSteppingPluginFactory.java
```diff
@@ -39,7 +39,7 @@ public interface HumanoidSteppingPluginFactory extends HighLevelHumanoidControll
 
    default void createStepGeneratorNetworkSubscriber(String robotName, RealtimeROS2Node realtimeROS2Node)
    {
-      ROS2Topic<?> baseTopic = ControllerAPI.getBaseTopic(HumanoidControllerAPI.HUMANOID_CONTROL_MODULE_NAME, robotName);
+      ROS2Topic<?> baseTopic = HumanoidControllerAPI.getBaseTopic(robotName);
```
stefanfasano, line 42: are these changes equivalent? just wanna make sure this doesn't break any CSG ros stuff. If they are not, there are some publishers that interface with this that will need to be changed to maintain compatibility

stefanfasano: looks like they are the same

stefanfasano (approved):

ihmc-common-walking-control-modules/src/main/java/us/ihmc/commonWalkingControlModules/highLevelHumanoidControl/plugin/HumanoidSteppingPluginFactory.java
TomaszTB: Yup, I made sure they're equivalent. We should use HumanoidControllerAPI if we're accessing humanoid topics. Makes it easier to find where we use what topics

PR #1308, opened 2026-03-17
rdx startup shutdown

reesep5:


ihmc-interfaces/src/main/messages/ihmc_interfaces/controller_msgs/msg/EStopMasterGainStatusMessage.msg
```diff
@@ -1,5 +1,11 @@
 # A message for allowing operator to observe the status of various startup/shutdown variables often required for robot operation
 
+# Current startup status
+bool startup_in_progress
```
stefanfasano: it looks like the in_progress ones didn't get generated in the other classes. Do we want those guys?

reesep5: Oh, that was accidental. I need to remove those

stefanfasano (approved):

PR #1347, opened 2026-04-16
removed some fault listeners in avatar multi thread factory

stefanfasano:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -658,23 +658,6 @@ private HighLevelHumanoidControllerFactory createHighLevelControllerFactory(DRCR
                                                                                controllerFactory,
                                                                                !highLevelControllerParameters.automaticallyTransitionToWalkingWhenReady()));
 
-         // Transition to DO_NOTHING in the event of a fault
-         hardwareCommunicationInterface.addFaultListener(change ->
```
stefanfasano, line 662: some of this was already duplicated in AlexROS2HardwareCommunication. I think it would be best for the robot-specific implementations of hardware communication interface to handle these fault conditions

PR #1361, opened 2026-05-19
Add a prefix string to PushRobotControllerSCS2 so it is possible to h…

james-p-foster:
…ave multiple pushers on one joint

### DIFF
--- a/ihmc-simulation-toolkit/src/main/java/us/ihmc/simulationToolkit/controllers/PushRobotControllerSCS2.java
+++ b/ihmc-simulation-toolkit/src/main/java/us/ihmc/simulationToolkit/controllers/PushRobotControllerSCS2.java
@@ -1,7 +1,5 @@
 package us.ihmc.simulationToolkit.controllers;
 
-import java.awt.event.ActionEvent;
-import java.awt.event.ActionListener;
 import java.util.LinkedList;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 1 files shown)


ihmc-simulation-toolkit/src/main/java/us/ihmc/simulationToolkit/controllers/PushRobotControllerSCS2.java
```diff
@@ -66,23 +65,29 @@ public PushRobotControllerSCS2(DoubleProvider time, Robot pushableRobot, String
    }
 
    public PushRobotControllerSCS2(DoubleProvider time, Robot pushableRobot, String jointNameToApplyForce, Vector3DReadOnly forcePointOffset, double visualScale)
+   {
+      this(time, pushableRobot, jointNameToApplyForce, jointNameToApplyForce, forcePointOffset, visualScale);
+   }
+
+   public PushRobotControllerSCS2(DoubleProvider time, Robot pushableRobot, String jointNameToApplyForce, String name, Vector3DReadOnly forcePointOffset, double visualScale)
```
Beomyeong, line 72: How this change works for multiple push to the joints?
It seems like adding a duplicated variable, name, into a constructor. And use that only for the defining yoVariables.

james-p-foster: Before, you only passed in the joint name string, so if you wanted multiple pushes on a joint, e.g. a couple instead of a force, you'd get a duplicate name error. The additional string let's you disambiguate

Beomyeong (approved):

PR #1384, opened 2026-06-20
Only create the SCS2StateEstimatorDebugVariables if the GUI is shown for SCS2 because otherwise the user won't be debugging anything

PotatoPeeler3000:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/scs2/SCS2AvatarSimulationFactory.java
```diff
@@ -376,12 +376,16 @@ else if (useBulletPhysicsEngine.hasValue() && useBulletPhysicsEngine.get())
...
       robot = simulationConstructionSet.addRobot(robotDefinition);
-      robot.addThrottledController(new SCS2StateEstimatorDebugVariables(simulationConstructionSet.getInertialFrame(),
-                                                                        gravity.get(),
-                                                                        robotModel.getEstimatorDT(),
-                                                                        robot.getControllerManager()
-                                                                             .getControllerInput()),
-                                   robotModel.getEstimatorDT());
+
+      // The debug variables don't need to be updated when the GUI isn't shown as users won't have a way of debugging
+      if (showGUI.get())
```
PotatoPeeler3000, line 381: Added this check here. The debug variables are useful to the user when using the GUI. But don't provide anything when running headless like in CI

PR #1403, opened 2026-07-28
Feature/ground mimic init

LuigiPenco93:


ihmc-simulation-toolkit/src/main/java/us/ihmc/sensorProcessing/simulatedSensors/SCS2SensorReader.java
```diff
@@ -37,6 +37,7 @@
 import us.ihmc.sensorProcessing.stateEstimation.IMUSensorReadOnly;
 import us.ihmc.sensorProcessing.stateEstimation.SensorProcessingConfiguration;
 import us.ihmc.yoVariables.registry.YoRegistry;
+import us.ihmc.yoVariables.variable.YoBoolean;
```
LuigiPenco93: These changes exist so perfect sensors can be toggled at runtime while the state estimator still exists — needed for Ground Prep → RL in sim.

On develop: perfect sensors is a constructor-only mode (usePerfectSensor). That path builds no state estimator. You can’t leave Ground Prep (perfect) and later run NORMAL estimation without rebuilding.

On this branch:
Always build the estimator path.
Add runtime YoBoolean usePerfectRootJointSensors + setUsePerfectSensors(...).
When on: copy sim root pose (applyPerfectRootJointState) and 1-DoF joints onto the controller model; pair with disabling the main estimator so it doesn’t overwrite that.
When off: normal sensor → estimator path.
That lets RLSimulationFactory use perfect sensors only in Ground Prep, then switch to the real estimator for RL — without the old “perfect = no estimator forever” constraint.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/drcRobot/DRCRobotModel.java
```diff
@@ -42,6 +43,25 @@ public default RobotTarget getTarget()
 
    public abstract RobotInitialSetup<HumanoidFloatingRootJointRobot> getDefaultRobotInitialSetup();
```
LuigiPenco93, line 44: Supporting more initial setups

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/AvatarEstimatorThread.java
```diff
@@ -131,6 +136,78 @@ public void run()
       estimatorThreadTimer.stopMeasurement();
    }
 
+   /**
```
LuigiPenco93: these are needed to switch from perfect sensors to state estimator

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/drcRobot/DRCRobotModel.java
rjgriffin42: The way we have done this in the past is to override the return of the robot model with the one you want to use. This can be done with a setter added to AlexRobotModel, or it could be done by directly overriding the method when creating the class. Then you are changing it at the place of use, rather than for all the typ4es, like here, and you wouldn't have to do changes in HumanoidRobotInitialSetup.

LuigiPenco93: fixed

ihmc-high-level-behaviors/src/libgdx/java/us/ihmc/rdx/ui/graphics/RDXRobotPerceptionVisualizersPanel.java
```diff
@@ -58,6 +58,14 @@ protected RDXRobotPerceptionVisualizersPanel(ROS2Node ros2Node, ROS2SyncedRobotM
       // Additional visualizers instantiated in robot specific class
    }
 
+   /**
+    * Hook for robot-specific visualizers (e.g. collision mesh overlays).
+    */
+   protected RDXROS2RobotVisualizer createRobotVisualizer(ROS2Node ros2Node, ROS2SyncedRobotModel syncedRobot)
```
ds58, line 64: I cant see why this needed to be a method

ds58 (changes requested):

ds58 (approved):

PR #1412, opened 2026-08-04
Gr00t implementation

EAOZONE:
## Summary

This PR generalizes and hardens the existing OpenPI websocket client so it can support GR00T policies with embodiment-specific state and action layouts.

It also adds a `Gr00tClient` that validates the deployed policy contract before sending observations. This prevents an incompatible checkpoint, joint ordering, action horizon, or sampling rate from being used accidentally.

## Changes

- Generalized `OpenpiClient` to support:
  - Independent state and action dimensions
  - Configurable image, state, prompt, and action wire keys
  - Configurable image dimensions and action chunk length
  - Runtime prompt updates
  - Policy-provided action horizons
- Made MessagePack response parsing independent of map field ordering.
- Added strict validation of action tensor shape, dtype, payload size, and ndarray metadata.
- Clear buffered actions after malformed responses so stale commands cannot be reused.
- Improved websocket failure handling:
  - Pending requests complete exceptionally when the connection closes.
  - Metadata waits are unblocked by connection failures.
  - Failed writes are removed from the pending-response queue.
  - Connection errors are exposed to callers for diagnostics.
- Added `Gr00tClient`, which validates:
  - Semantic layout ID
  - State and action sizes
  - Action horizon
  - Action sampling rate
- Updated the existing VLA caller for the stricter client behavior.

## Compatibility

GR00T-specific contract validation is implemented by the new client through the generic metadata-validation hook.

## Testing

Added focused coverage for:

- Independent state and action dimensions
- Unordered response fields
- Action horizon decoding
- Malformed tensor rejection and buffer clearing
- Duplicate wire-key rejection
- Little-endian numeric buffers
- GR00T metadata compatibility
- Disconnect handling during metadata and inference requests

All 10 focused GR00T/OpenPI tests pass.


ihmc-high-level-behaviors/src/main/java/us/ihmc/gr00t/Gr00tHandController.java
```diff
@@ -0,0 +1,37 @@
+package us.ihmc.gr00t;
+
+import us.ihmc.robotics.robotSide.RobotSide;
+
+/** Hand telemetry and optional policy/manual command output used by a humanoid GR00T task. */
+public interface Gr00tHandController
```
TomaszTB, line 6: This interface is only useful if we want to have policies that use different hands.. are we ever going to use, say, the SAKE EZGrippers with Gr00t?

ihmc-high-level-behaviors/src/main/java/us/ihmc/gr00t/Gr00tHumanoidArmPlanner.java
```diff
@@ -0,0 +1,359 @@
...
+import us.ihmc.mecano.multiBodySystem.interfaces.OneDoFJointBasics;
+import us.ihmc.robotModels.FullHumanoidRobotModel;
+import us.ihmc.robotModels.FullRobotModelUtils;
+import us.ihmc.robotics.partNames.NeckJointName;
+import us.ihmc.robotics.robotSide.RobotSide;
+
+import java.util.List;
+
+/** Converts validated Cartesian policy chunks into fixed-base arm and neck trajectories. */
+final class Gr00tHumanoidArmPlanner
```
TomaszTB, line 22: Is this intentionally package-private?

TomaszTB (approved):
Just did a scan of the code. Mostly looks fine. Can't say I did a thorough review of the logic.. that would take me a while

LuigiPenco93 (approved):

ihmc-high-level-behaviors/src/main/java/us/ihmc/gr00t/Gr00tHandController.java
EAOZONE: aren't we making new hands? I also have tried to make this as generalizable as possible to be able to go from one machine to another with ease.

ihmc-high-level-behaviors/src/main/java/us/ihmc/gr00t/Gr00tHumanoidArmPlanner.java
EAOZONE: Nope.

PR #1413, opened 2026-08-04
Integrate Invariant State Estimator into Framework

lucaslibshutz:
# List of Things to Change/Fix

### Priority 1: Structural
- [ ] Contact Detection Pipeline (WBCC vs. estimator)
- [ ] **_Implement ContactNet ONNX Runtime here_**
- [ ] All state estimator parameters should be in an Alex specific file
- [x] Split `JointLevelKFPreFilter.java` into more organized components (merged in #1421 )
- [ ] Remove ProprioceptivePreFilter factory?
- [ ] Move rotor inertia values to be somewhere that is referenced, or don't use them at all
- [ ] Change state estimator factory existing to accept InEKF as mode rather than override (PR in progress, open once #1421 is merged for brevity)
    - [ ] Also ensure that the old `imuBiasStateEstimator` is **not removed**.
- [ ] Add `allocate` functionality to constructors for classes, such that we don't make new objects each time in an unsafe manner.
- [x] Check `DMatrixRMaj` usage, should not be `Matrix3D`.
    - Euclid needs to return `Matrix3D`, a `DMatrixRMaj` data type doesn't make sense here for rotation matrices and lie group items.
- [ ] Whole codebase needs tuning vars to be Yo-Variable-ized.
- [ ] **Needs to be documented**: InEKF might need live pelvis bias estimation, both linear and angular - **make a note of this, *especially* for 002.**

### Priority 2: Realtime Ops / Legibility
- [x] **JointKF**: Look at `JointIndexHandler` versus the `LinkedHashMap<OneDoFJointBasics, ...>`, especially as that was using `Integer`, and not `TObjectInt...`.
   - Can't use this with the current state of the filter, as we need IMUs linked to joints, AND pre-constructing the whole tree would cost `dim^3` for rows that are zero, which wastes computational overhead for pretty much no reason in my opinion.
- [ ] Fix import of `createForKinematicsEstimator` in `AvatarEstimatorThreadFactory.java`, line 590, and line 598.
- [x] Check `Integer` versus `TInteger` uses (via `Trove` Java library)
- [x] All variables should be full names, not abbreviations
- [x] Change `Lambda` to `lambda`
- [ ] Ensure the `symmetrize(lambda)` function is not being done automatically
- [ ] Check if `MatrixTools.addDiagonal()` can be used to add rotor inertia values for the `lambda` process noise Schur complement.
- [ ] `ContactUpdater.java`: Lines 171-172 should populate the full span of the Jacobian, if not you might be able to remove the `zero()` call there. Check that  it does so, same with `GravityLevelingUpdater.java`:235-238.
    - Pretty sure that this is not the case, `GravityLevelingUpdater.java` is confirmed to be fine, but I will check for `ContactUpdater.java`.
 - [ ] `FootSwitchProbabilityProvider.java` lines 156-159 might be too strict, for "going up and down" a bunch of times.
 - [ ] `InvariantEKF.java`, line 177: you should not be making a new `Vector3D()`, as that causes allocation at runtime among other things.

#### `KinematicContactDetector.java`
- [ ] Change `logistic()` function in the `KinematicContactDetector` to be `Math.log`, same with `Math.clamp()` for `clamp()`.
- [ ] Line 116 of KD should use `EuclidCoreTools.interpolate()`.
- [ ] Line 132 of KD should use:
```java
FixedFramePoint3DBasics soleWorld = soleInWorld.get(side);
soleWorld.setMatchingFrame(soleFrames.get(side));
```
- [ ] Line 102 `frameUpdater.run()` might have to be under a sim-only flag, not used on hardware.
- [ ] Line 46-49, from Robert:
> This is used only as a temp object. Because of that, you could instead create `tempSoleInWorld` as a singular `FramePoint3D`, instead of in the list, and you could also store it as `FixedFramePoint3DBasics` so that you know it's frame is always invariant within the scope of this class.
- [ ] Line 125-128, from Robert:
> You could just pass in the `soleFrame` itself here, you don't need the side information. Then you could do `soleFrame.getTransformToRoot().getTranslationZ() - groundHeight`;

#### `FootReferencedYawCorrector.java`
- [ ] Lines 81-84, from Robert:
> We tend to try and always use the side prefix. So wrap these in a loop, and do `anchored.put(side, new YoBoolean(side.getLowerCaseName() + "_YawAnchored, registry);` This makes it less prone to naming mistakes between left and right.
- [ ] Lines 130-131 should use `AngleTools.angleMinusPiToPi(...)`.
- [ ] From Robert, lines 136-139:
> Is this going to result in the right thing? Your weighted sum is getting wrapped. Should you instead normalize the weighting value by the total weight, so that you do that normalization within the sum itself?

### Priority 3: Quality of Life Stuff
- [ ] CI Integration for unit tests for future use?
- [ ] Joint KF should support quadrupeds.
- [ ] Schur complement should be converted to the **centroidal mass matrix** formulation, as described in [this paper](https://ieeexplore.ieee.org/document/4650772)


ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariant_estimator/ConstantContactMeasurementNoiseProvider.java
```diff
@@ -0,0 +1,28 @@
+package us.ihmc.stateEstimation.invariant_estimator;
```
rjgriffin42: we always use camelCase for directories.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariant_estimator/SEK3_Utils.java
```diff
@@ -0,0 +1,262 @@
...
+ * <p>k is inferred from the argument sizes: {@code ξ.length = 3 + 3k} and
+ * {@code X} is {@code (3+k)×(3+k)}.</p>
+ *
+ * <p>All methods are static; {@code ToPack} parameters are the outputs — inputs are never
+ * modified. Each of {@link #exp}, {@link #log} and {@link #adjoint} has two forms: a convenience
+ * form that allocates its own scratch (fine for tests and other non-real-time callers) and an
+ * allocation-free overload that takes caller-owned {@link Vector3D}/{@link RotationMatrix}/{@link Matrix3D}
+ * scratch. Use the latter on hard real-time loops — the estimator's per-tick predict/update path does.</p>
+ */
+public class SEK3_Utils
```
rjgriffin42: We always use SnakeCase for classes.

rjgriffin42:
I would strongly prefer this be implemented as additional option mode, rather than a replacement. At the moment, you're replacing modules (e.g. the imuBiasStateEstimator). The estimator itself already has running modes baked into it (FROZEN and NORMAL). You could introduce a third (EKF) mode, and a fourth (BOTH) mode, and resolve what is going on between the two that way. Specifically for what I pointed out, the interfaces enable this quite easily (both the imuBiasStateEstimator and the preFilter implement IMUBiasProvider)

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/ProprioceptivePreFilterFactory.java
```diff
@@ -0,0 +1,66 @@
...
+         case NONE -> new PassThroughPreFilter();
+         case ALPHA_COMPLEMENTARY -> AlphaComplementaryPreFilter.createForKinematicsEstimator(sensorOutputMap,
+                                                                                              stateEstimatorParameters,
+                                                                                              imuProcessedOutputs,
+                                                                                              feet,
+                                                                                              gravitationalAcceleration,
+                                                                                              cancelGravityFromAccelerationMeasurement,
+                                                                                              estimatorDT,
+                                                                                              parentRegistry);
+         case JOINT_KF -> JointLevelKFPreFilter.createForKinematicsEstimator(sensorOutputMap,
```
rjgriffin42, line 87: I'd really like to move away from factories. I don't see how this style of code provides any advantage over directly calling the constructor.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+               }
+            }
+            massMatrixNuisanceColumns = new int[nuisanceColumns.size()];
+            nuisanceRotorInertiaDiag = new double[nuisanceColumns.size()];
+            for (int i = 0; i < massMatrixNuisanceColumns.length; i++)
+            {
+               massMatrixNuisanceColumns[i] = nuisanceColumns.get(i);
+               nuisanceRotorInertiaDiag[i] = nuisanceRotor.get(i);
+            }
+            numNuisanceDoF = massMatrixNuisanceColumns.length; // 6 + gap joints
```
rjgriffin42: don't abbreviate words, that provides no advantage and goes against our programming style.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+   // Lambda_eff = Lambda + diag(n_i^2 J_rotor,i). Physics: the rotor spins behind the gearbox about its own
+   // axis and does not couple through the floating base, so the post-Schur diagonal add is simultaneously the
+   // EXACT drivetrain term and a principled regularizer (it floors lambda_min(Lambda_eff) by Weyl). Distal
+   // joints' link-side apparent inertia is as low as ~8e-4 kg*m^2 while their drivetrains reflect 0.05-0.07 —
+   // without this term Lambda^-2 has diagonal outliers up to ~1.6e6 and Qa blows up for proximal/light joints.
+   private static final String[] ROTOR_INERTIA_JOINT_KEYS = {
+         "HIP_X", "HIP_Z", "HIP_Y", "KNEE", "ANKLE_Y", "ANKLE_X", "SPINE",
+         "SHOULDER_Y", "SHOULDER_X", "SHOULDER_Z", "ELBOW",
+         "WRIST_Z", "WRIST_X", "GRIPPER_Z", "NECK_Z", "NECK_Y"};
+   private static final double[] ROTOR_INERTIA_VALUES = {
```
rjgriffin42: This should absolutely not exist in here.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+   private static final double SIGMA_GYRO_FLOOR = 1.0e-6;        // (0.001 rad/s)^2 per axis (safety net)
+   private static final double SIGMA_GYRO_FLOOR_TRACE = 3.0e-6;  // 3 * (0.001 rad/s)^2
+   // Active innovation-covariance conditioning gate (Part B item 5). cond(S) is estimated from the Cholesky
+   // factor diagonal ((max L_ii / min L_ii)^2 — no eigendecomposition); above this the whole stacked update is
+   // skipped, because a finite-but-ill-conditioned S inverts to a huge gain that the Joseph K R K^T loop squares
+   // each tick (the P divergence mechanism the LU/isFinite guards were blind to).
+   private static final double COND_S_MAX = 1.0e9;
+   private static final double INIT_POS_VAR = 1.0e-6; // encoders trusted at initialization
+   private static final double INIT_VEL_VAR = 1.0; // velocity unknown at initialization
+   private static final double INIT_BIAS_VAR = 2.5e-3; // (0.05 rad/s)^2
```
rjgriffin42: All this stuff should be extracted into a parameters class, and then used to create yo variables that can be tuned on the fly within this filter.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+   private final DMatrixRMaj PBackup = new DMatrixRMaj(0, 0);
+   // One-shot triage: names the first predict/update stage whose output goes non-finite, so a hardware run
+   // prints exactly where the NaN enters. Latches after the first report (the KF stays NaN anyway).
+   private boolean nonFiniteStateReported = false;
+
+
+   // Package-private (not private) so the allocation/behavior tests in this package can build it directly
+   // from a synthetic IMU-pair setup without standing up a full StateEstimatorParameters.
+   /** Scalar-CWNA process-noise overload: no robot model, so Qa = SIGMA_ACCEL^2 I (the pre-mass-matrix behavior). */
+   JointLevelKFPreFilter(SensorOutputMapReadOnly sensorMap,
```
rjgriffin42, line 81: Because of the way this whole thing is written, a lot of your allocations occur in the `allocate()` function. This means that the objects themselves aren't declared as final, and they can be overriden. While it makes the constructor smaller, it provides more vulnerability to the codebase. All objects should be final, which means they should be declared eitehr in the constructor or the class definition. In the case of most of these, you could declare them in the class definition and then call reshape in the allocation block, which would satisfy this.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+            MNN.set(a, b, massMatrix.get(ca, massMatrixNuisanceColumns[b])); // M_NN (NxN)
+         for (int j = 0; j < n; j++)
+            MNf.set(a, j, massMatrix.get(ca, massMatrixColumn[j]));          // M_Nf (Nxn) = M_jN^T
+      }
+      for (int i = 0; i < n; i++)
+      {
+         int ci = massMatrixColumn[i];
+         for (int j = 0; j < n; j++)
+            Mff.set(i, j, massMatrix.get(ci, massMatrixColumn[j]));          // M_ff (nxn)
+      }
```
rjgriffin42: Is there a reason this isn't using block operations? I see the thing saying you shouldn't assume ordering, but you're literally constructing it, you could construct it such that ordering is enforced. Even if you don't, you could still do column-wise block operations, which are written into `CommonOps_DDRM.extract(..)` or `MatrixTools.setMatrixBlock(...)`. You could also write a convenience function in `MatrixTools.setMatrixColumn` that handles some of that for you, or even provide an array of columns to set, which is these for loops, which would make the code more readable, testable, and reliable.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+
+      // Gap-joint reflected rotor inertia on the nuisance-block diagonal (Part B item 1): 0 on the 6 base rows,
+      // the joint's reflected rotor inertia on each unfiltered gap joint (none on Alex; kept general). Floors
+      // M_NN the same way the filtered rotor diagonal floors Lambda_eff below.
+      if (nuisanceRotorInertiaDiag != null)
```
rjgriffin42: I don't find the naming "nuisance" as intuitive at all. What does that mean?

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+         for (int a = 0; a < nN; a++)
+            MNN.add(a, a, nuisanceRotorInertiaDiag[a]);
```
rjgriffin42: The name of this object leavea  lot to be desired.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+      if (containsNonFinite(MNNInvMNf))
+      {
+         warnMassMatrixFailureOnce("non-finite M_NN^-1 M_Nf (near-singular nuisance block)");
+         return;
+      }
+
+      // Lambda = M_ff - M_jN X = M_ff - M_Nf^T X (M symmetric => M_jN = M_Nf^T). SPEC §3.2.
+      CommonOps_DDRM.multTransA(MNf, MNNInvMNf, Lambda); // Lambda <- M_Nf^T X
+      CommonOps_DDRM.changeSign(Lambda);                 // Lambda <- -M_jN X
+      CommonOps_DDRM.addEquals(Lambda, Mff);             // Lambda <- M_ff - M_jN X
```
rjgriffin42: Replace these three lines with
```
Lambda.set(Mff);
CommonOps_DDRM.multAddTransA(-1.0, MNf, MNNInvMNf, Lambda);
```

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+         return;
+      }
+      nuisanceMassMatrixSolver.solve(MNf, MNNInvMNf); // X (Nxn); solve leaves MNf unmodified
+      if (containsNonFinite(MNNInvMNf))
+      {
+         warnMassMatrixFailureOnce("non-finite M_NN^-1 M_Nf (near-singular nuisance block)");
+         return;
+      }
+
+      // Lambda = M_ff - M_jN X = M_ff - M_Nf^T X (M symmetric => M_jN = M_Nf^T). SPEC §3.2.
```
rjgriffin42: `Lambda` should be named `lambda`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+      CommonOps_DDRM.multTransA(MNf, MNNInvMNf, Lambda); // Lambda <- M_Nf^T X
+      CommonOps_DDRM.changeSign(Lambda);                 // Lambda <- -M_jN X
+      CommonOps_DDRM.addEquals(Lambda, Mff);             // Lambda <- M_ff - M_jN X
+      // Symmetrize before the Cholesky invert: the block extraction is exact but the matrix products leave
+      // Lambda symmetric only to round-off, and Cholesky assumes exact symmetry.
+      symmetrize(Lambda);
```
rjgriffin42: Is this ever actually a problem? I would be surprised if this isn't handled by the back-end EJML solver.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+
+      // Lambda_eff = Lambda + diag(reflected rotor inertia) (Part B item 1, the primary fix). The rotor spins
+      // behind the gearbox about its own axis and does not couple through the floating base, so this post-Schur
+      // diagonal add is the EXACT drivetrain term — and simultaneously a principled regularizer: by Weyl it
+      // floors lambda_min(Lambda_eff) >= min_i rotorInertiaDiag[i], so Lambda_eff^-1 has no proximal-joint
+      // outliers and Qa cannot blow up the way the un-floored sigma_tau^2 Lambda^-2 did on Alex002.
+      for (int i = 0; i < n; i++)
+         Lambda.add(i, i, rotorInertiaDiag[i]);
```
rjgriffin42: I'd prefer this be done in another `MatrixTools` method. There's already an `addDiagonal` for singular values, so this is a natural extension.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+      }
+
+      // Per-joint Sigma_tau via the Gram form (Part B item 3): Y = Lambda_eff^-1 with column j scaled by
+      // sigma_tau,j, then Qa = Y Y^T = Lambda_eff^-1 diag(sigma_tau^2) Lambda_eff^-T. Qa is PSD AND exactly
+      // symmetric by construction, so the Van Loan fill below reads it directly (no symmetrized read). This
+      // replaces the scalar Qa = sigma_tau^2 Lambda^-2; per-joint sigma_tau = ALPHA * tau_max,i scales the
+      // process noise with each actuator's own torque capacity instead of one number across hip and wrist.
+      for (int i = 0; i < n; i++)
+         for (int j = 0; j < n; j++)
+            Ytau.set(i, j, LambdaInv.get(i, j) * sigmaTauPerJoint[j]); // scale column j by sigma_tau,j
```
rjgriffin42: Replace this with 
```
Ytau.set(LambdaInv);
for (int col = 0; col < n; col++)
   CommonOps_DDRM.scaleCol(sigmaPerTauJoint[col], Ytau, col);
```

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+      CommonOps_DDRM.multTransB(Ytau, Ytau, Qa); // Qa = Y Y^T
```
rjgriffin42: Replace with `CommonOps_DDRM.multOuter(Ytau, Qa);`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+
+   private final LinkedHashMap<OneDoFJointBasics, Integer> jointToIndex = new LinkedHashMap<>();
+   private final LinkedHashMap<IMUSensorReadOnly, Integer> imuToOrdinal = new LinkedHashMap<>();
+   private final List<Pair> pairs = new ArrayList<>();
+   private final List<FootAnchor> footAnchors = new ArrayList<>();
+   private IMUSensorReadOnly baseIMU;
+
+   private int baseBiasCol;
+   private int n; // number of distinct joints
+   private int m; // number of IMUs
```
rjgriffin42: change this to `numberOfIMUs`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+   private final double dt;
+
+   private final LinkedHashMap<OneDoFJointBasics, Integer> jointToIndex = new LinkedHashMap<>();
+   private final LinkedHashMap<IMUSensorReadOnly, Integer> imuToOrdinal = new LinkedHashMap<>();
+   private final List<Pair> pairs = new ArrayList<>();
+   private final List<FootAnchor> footAnchors = new ArrayList<>();
+   private IMUSensorReadOnly baseIMU;
+
+   private int baseBiasCol;
+   private int n; // number of distinct joints
```
rjgriffin42: change this to `numberOfJoints`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/KinematicContactDetector.java
```diff
@@ -0,0 +1,145 @@
...
+   {
+      FramePoint3D soleWorld = soleInWorld.get(side);
+      soleWorld.setToZero(soleFrames.get(side));
+      soleWorld.changeFrame(ReferenceFrame.getWorldFrame());
+      return soleWorld.getZ() - groundHeight;
+   }
+
+   private static double logistic(double x)
+   {
+      return 1.0 / (1.0 + Math.exp(-x));
```
rjgriffin42, line 138: Math.log?

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/KinematicContactDetector.java
```diff
@@ -0,0 +1,145 @@
...
+   }
+
+   private static double clamp(double value)
+   {
+      return value < 0.0 ? 0.0 : (value > 1.0 ? 1.0 : value);
```
rjgriffin42, line 143: `MathTools.clamp()`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/KinematicContactDetector.java
```diff
@@ -0,0 +1,145 @@
...
+   {
+      return probability[side.ordinal()];
+   }
+
+   /** @return the sole-origin height above the support plane (m), in world. */
+   private double soleHeightAboveGround(RobotSide side)
+   {
+      FramePoint3D soleWorld = soleInWorld.get(side);
+      soleWorld.setToZero(soleFrames.get(side));
+      soleWorld.changeFrame(ReferenceFrame.getWorldFrame());
```
rjgriffin42, line 132:
```
FixedFramePoint3DBasics soleWorld = soleInWorld.get(side);
soleWorld.setMatchingFrame(soleFrames.get(side));
```

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/KinematicContactDetector.java
```diff
@@ -0,0 +1,145 @@
...
+
+         double previous = previousHeight[side.ordinal()];
+         double verticalSpeed = Double.isNaN(previous) ? 0.0 : (height - previous) / dt;
+         previousHeight[side.ordinal()] = height;
+
+         double pHeight = logistic((heightThreshold - height) / heightWidth);
+         double pSpeed = logistic((speedThreshold - Math.abs(verticalSpeed)) / speedWidth);
+         double pRaw = pHeight * pSpeed;
+
+         double pSmoothed = smoothingAlpha * probability[side.ordinal()] + (1.0 - smoothingAlpha) * pRaw;
```
rjgriffin42, line 116: `EuclidCoreTools.interpolate()`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/KinematicContactDetector.java
```diff
@@ -0,0 +1,145 @@
...
+
+      for (RobotSide side : RobotSide.values)
+         soleInWorld.put(side, new FramePoint3D());
+   }
+
+   @Override
+   public void update()
+   {
+      if (frameUpdater != null)
+         frameUpdater.run();
```
rjgriffin42, line 102: THis looks to only be used in the walking track, and it may not actually serve a purpose because of that.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/KinematicContactDetector.java
```diff
@@ -0,0 +1,145 @@
...
+   private final double dt;
+
+   private final double groundHeight;
+   private final double heightThreshold;   // h0: half-transition height (m)
+   private final double heightWidth;       // wHeight: logistic width in height (m)
+   private final double speedThreshold;    // v0: half-transition vertical speed (m/s)
+   private final double speedWidth;        // wSpeed: logistic width in speed (m/s)
+   private final double smoothingAlpha;    // exponential smoother in [0, 1); 0 = no smoothing
+
+   private final SideDependentList<FramePoint3D> soleInWorld = new SideDependentList<>();
```
rjgriffin42, line 49: This is used only as a temp object. Because of htat, you could instead create `tempSoleInWorld` as a singular FramePoint3D, instead of in the list, and you could also store it as `FixedFramePoint3DBasics` so that you know it's frame is always invariant within the scope of this class.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/KinematicContactDetector.java
```diff
@@ -0,0 +1,145 @@
...
+   }
+
+   @Override
+   public double getContactProbability(RobotSide side)
+   {
+      return probability[side.ordinal()];
+   }
+
+   /** @return the sole-origin height above the support plane (m), in world. */
+   private double soleHeightAboveGround(RobotSide side)
```
rjgriffin42, line 128: You could just pass in the soleFrame itself here, you don't really need the side information. Then, you could do 
`soleFrame.getTransformToRoot().getTranslationZ() - groundHeight;`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/FootReferencedYawCorrector.java
```diff
@@ -0,0 +1,171 @@
...
+      this.soleFrames = soleFrames;
+      this.contactProbability = contactProbability;
+      this.correctionGain = correctionGain;
+      this.highThreshold = highThreshold;
+      this.lowThreshold = lowThreshold;
+
+      anchored = new SideDependentList<>(new YoBoolean("yawAnchoredLeft", registry), new YoBoolean("yawAnchoredRight", registry));
+      anchorPelvisYaw = new SideDependentList<>(new YoDouble("yawAnchorPelvisLeft", registry), new YoDouble("yawAnchorPelvisRight", registry));
+      anchorRelativeFootYaw = new SideDependentList<>(new YoDouble("yawAnchorRelFootLeft", registry),
+                                                      new YoDouble("yawAnchorRelFootRight", registry));
```
rjgriffin42, line 95: we tend to try and always use the side prefix. So wrap these in a loop, and do 
`anchored.put(side, new  YoBoolean(side.getLowerCaseName() + "_YawAnchored, registry);` 

This makes it less prone to naming mistakes between left and right.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/FootReferencedYawCorrector.java
```diff
@@ -0,0 +1,171 @@
...
+      ekf.getRotation(filterRotation);
+      double pelvisYaw = filterRotation.getYaw();
+
+      double weightedDeltaSum = 0.0;
+      double weightSum = 0.0;
+
+      for (RobotSide side : RobotSide.values)
+      {
+         double relativeFootYaw = relativeFootYaw(side);
+         double contactP = clamp(contactProbability.applyAsDouble(side));
```
rjgriffin42: `MathTools.clamp(...)`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/FootReferencedYawCorrector.java
```diff
@@ -0,0 +1,171 @@
...
+         else if (contactP <= lowThreshold)
+         {
+            anchored.get(side).set(false);
+            isAnchored = false;
+         }
+
+         if (isAnchored)
+         {
+            double referenceYaw = anchorPelvisYaw.get(side).getDoubleValue()
+                                  + wrapToPi(anchorRelativeFootYaw.get(side).getDoubleValue() - relativeFootYaw);
```
rjgriffin42: `AngleTools.angleMinusPiToPi(...)`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/FootReferencedYawCorrector.java
```diff
@@ -0,0 +1,171 @@
...
+            weightedDeltaSum += contactP * wrapToPi(referenceYaw - pelvisYaw);
+            weightSum += contactP;
+         }
+      }
+
+      if (weightSum > 1.0e-6)
+      {
+         double correction = correctionGain * (weightedDeltaSum / weightSum);
```
rjgriffin42, line 157: Is this going to result int he right thing? Your weighted sum is getting wrapped. Should you instead normalize the weighting value by the total weight, so taht you do that normalization wthin the summing?

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/ContactUpdater.java
```diff
@@ -0,0 +1,250 @@
...
+    * @param contactIndex   the contact index i in [0, N).
+    * @param jacobianToPack the 3×m Jacobian to fill; reshaped and zeroed. Modified.
+    */
+   public void computeJacobian(InvariantState state, int contactIndex, DMatrixRMaj jacobianToPack)
+   {
+      jacobianToPack.reshape(MEASUREMENT_SIZE, state.getTangentSize());
+      jacobianToPack.zero();
+
+      setScaledIdentityBlock(jacobianToPack, state.basePositionTangentIndex(), +1.0);
+      setScaledIdentityBlock(jacobianToPack, state.contactTangentIndex(contactIndex), -1.0);
```
rjgriffin42, line 172: Are these two sets guaranteed to populate the full span of hte Jacobian? If so, you can remove the `zero()`. I would also check thte documentation, I think that reshape by default zeros unless you tell it not to.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/ContactUpdater.java
```diff
@@ -0,0 +1,250 @@
...
+   {
+      state.getRotation(rotation);
+
+      rotatedCovariance.set(bodyCovariance);
+      rotation.transform(rotatedCovariance);
+
+      covarianceToPack.reshape(MEASUREMENT_SIZE, MEASUREMENT_SIZE);
+      for (int r = 0; r < 3; r++)
+         for (int c = 0; c < 3; c++)
+            covarianceToPack.set(r, c, rotatedCovariance.getElement(r,c));
```
rjgriffin42: can be replaced with `rotatedCovariance.get(covarianceToPack)`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/ContactUpdater.java
```diff
@@ -0,0 +1,250 @@
...
+
+   /**
+    * Maps joint-encoder noise into a body-frame contact-position covariance: Nᵢ = J_Cᵢ·Σ_q·J_Cᵢᵀ,
+    * realizing the {@code J_Cᵢ(q)·w_q} term of the measurement model as a covariance.
+    *
+    * @param contactJacobian      the body-frame contact Jacobian J_Cᵢ (3×nJoints). Not modified.
+    * @param jointCovariance      the joint-encoder covariance Σ_q (nJoints×nJoints). Not modified.
+    * @param bodyCovarianceToPack the 3×3 body-frame covariance Nᵢ to fill. Modified.
+    */
+   public static void mapEncoderNoise(DMatrixRMaj contactJacobian, DMatrixRMaj jointCovariance, Matrix3D bodyCovarianceToPack)
```
rjgriffin42, line 202: This is only used in the test for this function. Why include it?

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/ContactUpdater.java
```diff
@@ -0,0 +1,250 @@
...
+      CommonOps_DDRM.multTransB(jSigma, contactJacobian, nBody);
+
+      bodyCovarianceToPack.set(nBody.get(0,0), nBody.get(0,1), nBody.get(0,2),
+                               nBody.get(1,0), nBody.get(1,1), nBody.get(1,2),
+                               nBody.get(2,0),nBody.get(2,1),nBody.get(2,2));
+
+   }
+
+   /** Writes scale·I into the 3×3 block of {@code dest} whose first column/row is at {@code blockStart}. */
+   private static void setScaledIdentityBlock(DMatrixRMaj dest, int blockStart, double scale)
```
rjgriffin42, line 222: This isn't correct, as it doesn't guarantee the matrix is zero first, within the scope of the function. I would also put this in matrix tools

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/InvariantState.java
```diff
@@ -0,0 +1,238 @@
...
+   }
+
+   /** @return the number of contact columns N. */
+   public int getNumberOfContacts()
+   {
+      return numberOfContacts;
+   }
+
+   /** @return the group-element size n = 5 + N. */
+   public int getGroupSize()
```
rjgriffin42, line 73: In the doc, I would say `5 + #getNumberOfContacts`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/InvariantState.java
```diff
@@ -0,0 +1,238 @@
...
+   }
+
+   /**
+    * Packs the rotation R (top-left 3×3 block of X) into {@code rotationToPack}.
+    *
+    * @param rotationToPack the rotation to pack the result into. Modified.
+    */
+   public void getRotation(RotationMatrixBasics rotationToPack)
+   {
+      rotationToPack.set(groupElement.get(0,0),groupElement.get(0,1),groupElement.get(0,2),
```
rjgriffin42: equivalent to calling `rotationToPack.set(groupElement)`, except this is less efficient because of all the size checks.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+      imu.getMeasurementFrame().getTransformToDesiredFrame(tmpTransform, jacFrame);
+      RotationMatrixReadOnly r = tmpTransform.getRotation();
+      out.reshape(3,3);
+
+      set_matrix(out, r);
+   }
+
+   public static void set_matrix(DMatrixRMaj out, RotationMatrixReadOnly r)
+   {
+      out.set(0, 0, r.getM00());
```
rjgriffin42: This is the same as saying `r.get(out);`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2584 @@
...
+   // because the foot-switch contact-probability source seeds to 1.0 on the assumption feet are planted at
+   // init, so a naive "both == 1" would false-pass on the first tick(s) precisely while hanging.
+   private static final double ON_GROUND_INIT_DEBOUNCE = 0.05; // s of continuous ground contact before seeding
+
+   // Declaration of all pre-allocated variables
+   private final SensorOutputMapReadOnly sensorMap;
+   private final double dt;
+
+   private final LinkedHashMap<OneDoFJointBasics, Integer> jointToIndex = new LinkedHashMap<>();
+   private final LinkedHashMap<IMUSensorReadOnly, Integer> imuToOrdinal = new LinkedHashMap<>();
```
rjgriffin42: For all this stuff, I would look at how the `JointIndexHandler` was written, as it does a lot of similar things.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/GravityLevelingUpdater.java
```diff
@@ -0,0 +1,397 @@
...
+    * complementary filter over the RAW gyro and the accelerometer. It deliberately reads NO filter state —
+    * see {@link #isQuasiStatic} for why that is the whole point.
+    */
+   private final Vector3D gravityReferenceBody = new Vector3D();
+   private final Vector3D referenceRate = new Vector3D();         // ω × ĝ_ref
+   private boolean gravityReferenceInitialized = false;
+
+   private static final Vector3D E_X = new Vector3D(1.0, 0.0, 0.0);
+   private static final Vector3D E_Y = new Vector3D(0.0, 1.0, 0.0);
+   private static final Vector3D E_Z = new Vector3D(0.0, 0.0, 1.0);
```
rjgriffin42: these can all be replaced by `Axis3D.X, Y, Z`

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/GravityLevelingUpdater.java
```diff
@@ -0,0 +1,397 @@
...
+      // residual = ĝ_ref − ĝ_body ≈ −R̂ᵀ[e_z]ₓ δφ.
+      residual.set(0, 0, measuredGravityBody.getX() - predictedGravityBody.getX());
+      residual.set(1, 0, measuredGravityBody.getY() - predictedGravityBody.getY());
+      residual.set(2, 0, measuredGravityBody.getZ() - predictedGravityBody.getZ());
+
+      // H_φ = −R̂ᵀ[e_z]ₓ = [ −R̂ᵀe_y | +R̂ᵀe_x | 0 ]; zeros on v, p, and all contact blocks.
+      rotation.inverseTransform(E_X, rTransposeEx);
+      rotation.inverseTransform(E_Y, rTransposeEy);
+      measurementJacobian.reshape(MEASUREMENT_SIZE, tangentSize);
+      measurementJacobian.zero();
```
rjgriffin42, line 239: again, I could be wrong, but I think the zero call is done automatically.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/InvariantEKF.java
```diff
@@ -0,0 +1,356 @@
...
+ */
+public class InvariantEKF
+{
+   private final InvariantState state;
+   private final InvariantPropagator propagator;
+   private final InvariantUpdater updater;
+   private final GravityLevelingUpdater gravityUpdater;
+
+   /** |g| (m/s²), used only by the gravity-leveling quasi-static gate. */
+   private static final double GRAVITY_MAGNITUDE = 9.81;
```
rjgriffin42: We pass this in in the robot model, so that there's a consistent magnitude used throughout the controller.

lucaslibshutz: This is done now, was merged in #1417 to this PR.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/GravityLevelingUpdater.java
lucaslibshutz: Don't think so, as I think the only place this could happen would be in `reshape`, and the zeroing call is only done when the data is short, i.e.:

```java
if (data.length < numRows * numCols) {
    double[] d = new double[numRows * numCols];
    if (saveValues) System.arraycopy(...);
    this.data = d;
}
this.numRows = numRows; 
this.numCols = numCols;
```

The only reason that I have this here is to ensure that the yaw observability about $\delta \varphi_z$ is maintained at zero, which should be done by the filter here anyways but is just a safety check essentially. This would only happen if we have dynamic contact points being added and removed from the filter (which is usually done if not training the network that we are). If you want I can get rid of it if you think it causes significant computational overhead.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/AvatarEstimatorThreadFactory.java
```diff
@@ -497,6 +529,156 @@ public StateEstimatorController createDRCKinematicsStateEstimator()
...
+                                                                                estimatorDT,
+                                                                                preFilterRegistry);
+
+      // 2026-07-16 (12 Hz investigation): when the JointKF is selected, ALSO build the alpha-
+      // complementary filter and wrap both in a live-switchable source (jointLevelSourceSelection
+      // YoEnum). Same InEKF base, joint q/qd source swappable mid-session; biases stay pinned to
+      // the JointKF inside the wrapper so the switch is bumpless for the base state.
+      if (preFilter instanceof us.ihmc.stateEstimation.jointLevel.JointLevelKFPreFilter)
+      {
+         ProprioceptivePreFilter alpha = us.ihmc.stateEstimation.jointLevel.AlphaComplementaryPreFilter.createForKinematicsEstimator(getProcessedSensorOutputMap(),
```
Beomyeong, line 501: You should call method by using import not fully qualified ?
i.e. 
PropriocepivePrefilter  alpha = AlphaComplementaryPreFilter.createForKinematicsEstimatior( blah, 
blah,
...)

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/AvatarEstimatorThreadFactory.java
```diff
@@ -497,6 +529,156 @@ public StateEstimatorController createDRCKinematicsStateEstimator()
...
+                                                                                                                                     imuProcessedOutputs,
+                                                                                                                                     feet,
+                                                                                                                                     getGravity(),
+                                                                                                                                     () -> cancelGravityFromAccelerationMeasurement,
+                                                                                                                                     estimatorDT,
+                                                                                                                                     preFilterRegistry);
+         preFilter = new us.ihmc.stateEstimation.jointLevel.SwitchableJointLevelSource(preFilter, alpha, preFilterRegistry);
```
Beomyeong: Same

ihmc-sensor-processing/src/main/java/us/ihmc/sensorProcessing/stateEstimation/StateEstimatorParameters.java
```diff
@@ -26,6 +26,26 @@ public List<IMUBasedJointStateEstimatorParameters> getIMUBasedJointStateEstimato
...
+    * (kinematics-based) pipeline, {@code NONE} disables IMU-pair joint fusion AND IMU bias
+    * compensation, which is NOT that pipeline's historical behavior.
+    */
+   public enum JointLevelEstimatorType
+   {
+      NONE, ALPHA_COMPLEMENTARY, JOINT_KF
+   }
+
+   /** Selected once at construction time; the default preserves the historical behavior for all robots. */
+   public JointLevelEstimatorType getJointLevelEstimatorType()
```
Beomyeong, line 44: We do not have a setter, do we need setter to set jointlevelEstimatotype in more high layer?

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/humanoid/kinematicsBasedStateEstimation/DRCKinematicsBasedStateEstimator.java
```diff
@@ -61,7 +63,7 @@ public class DRCKinematicsBasedStateEstimator implements StateEstimatorControlle
    private final PelvisRotationalStateUpdaterInterface pelvisRotationalStateUpdater;
    private final PelvisLinearStateUpdater pelvisLinearStateUpdater;
    private final MomentumStateUpdater momentumStateUpdater;
-   private final IMUBiasStateEstimator imuBiasStateEstimator;
+   private final ProprioceptivePreFilter preFilter;
```
Beomyeong, line 66: We should left the old one. 
Replacing new one with old one should be done with a lot of verfication. 
So until that, we should keep both and have the option to choose which estimation method we can use

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/humanoid/kinematicsBasedStateEstimation/DRCKinematicsBasedStateEstimator.java
```diff
@@ -151,12 +153,14 @@ public DRCKinematicsBasedStateEstimator(FullInverseDynamicsStructure inverseDyna
                                                                                       registry,
                                                                                       stateEstimatorParameters.cancelGravityFromAccelerationMeasurement());
 
-      imuBiasStateEstimator = new IMUBiasStateEstimator(imuProcessedOutputs,
```
Beomyeong, line 155: need to keep this

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/humanoid/kinematicsBasedStateEstimation/DRCKinematicsBasedStateEstimator.java
```diff
@@ -151,12 +153,14 @@ public DRCKinematicsBasedStateEstimator(FullInverseDynamicsStructure inverseDyna
...
+      preFilter = ProprioceptivePreFilterFactory.create(sensorOutputMap,
```
Beomyeong, line 157: Is this variable before implementing KF ? Isnt' it resulf after KF?

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/humanoid/kinematicsBasedStateEstimation/DRCKinematicsBasedStateEstimator.java
```diff
@@ -296,7 +302,7 @@ public void initialize()
       if (momentumStateUpdater != null)
          momentumStateUpdater.initialize();
 
-      imuBiasStateEstimator.initialize();
+      preFilter.initialize();
```
Beomyeong, line 306: Same, keep the old one and make to choose both as option

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/humanoid/kinematicsBasedStateEstimation/JointStateUpdater.java
```diff
@@ -1,26 +1,24 @@
 package us.ihmc.stateEstimation.humanoid.kinematicsBasedStateEstimation;
```
Beomyeong, line 1: It should be better to make separate KFJointStateUpdate r for using KF. And then we choose which joint update we use in the high level layer

lucaslibshutz: this is already implemented in #1421 , going to merge as soon as I get to it.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/humanoid/kinematicsBasedStateEstimation/DRCKinematicsBasedStateEstimator.java
lucaslibshutz: working on fix for this right now.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/ContactMeasurementNoiseProvider.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+public interface ContactMeasurementNoiseProvider
+{
+   /**
+    * Packs the per-contact measurement covariance in the body (pelvis) frame.
+    *
+    * @param side             which foot's contact measurement.
+    * @param covarianceToPack the 3x3 covariance. Modified.
+    */
+   void packContactCovariance(RobotSide side, Matrix3D covarianceToPack);
```
Beomyeong, line 22: Not the DMatrixRMaj type?

lucaslibshutz: good point, will check out the discrepancy, as I think I may have used them interchangeably here throughout the codebase.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointLevelKFPreFilter.java
```diff
@@ -0,0 +1,2586 @@
...
+                                       parentRegistry);
+   }
+
+   /**
+    * Finds the 6-DoF free-flyer joint at the root of the filtered joints' tree (SPEC §3.2 partitions the mass
+    * matrix over a 6-DoF base). In an IHMC model this is the elevator's single child joint (the SixDoFJoint /
+    * root joint). Returns {@code null} if no 6-DoF joint sits directly below the tree root (fixed-base model),
+    * which drops the filter to the scalar-CWNA fallback. Construction-only; no hot-path use.
+    */
+   private static JointReadOnly findFloatingBaseJoint(RigidBodyBasics treeRoot)
```
Beomyeong: We may need to have this as universal, just in case when estimator is needed for the qaudrupedal. I mean if we try to use this estimator for the quadrupedal, it's child joint number isn't 6.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/ContactMeasurementNoiseProvider.java
lucaslibshutz: checked, this is because the lie group tools Euclid uses are defined on manifolds, which means that the covariances need to be defined as Matrix3D so its compatible. This is probably worth a more thorough investigation in the future, but I don't think it makes much of a difference at the moment.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/FootSwitchContactProbabilityProvider.java
```diff
@@ -0,0 +1,221 @@
...
+   {
+      /** Debounced Schmitt trigger (2026-07-17): load hysteresis + dwell. Production default. */
+      SCHMITT,
+      /** Pre-2026-07-17 behavior: raw switch quantizer {0, 0.5, 1} + EMA, no debounce. A/B arm. */
+      LEGACY,
+      /** Contact detection off: both feet pinned fully trusted (p = 1, no EMA lag). Control arm — DRC-style permanent foothold trust. */
+      NONE
+   }
+
+   /** Probability when the foot is not trusted but the sensitive (unconfirmed) detector reports contact. */
```
Beomyeong, line 80: Make them can be tuned in real time until we find the proper gains.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/FootSwitchContactProbabilityProvider.java
```diff
@@ -0,0 +1,221 @@
...
+         boolean switchFiltered = footSwitch.hasFootHitGroundFiltered();
+         double load = footSwitch.getFootLoadPercentage();
+
+         // Schmitt trigger with dwell: the state-flipping condition must hold trustDwellTicks
+         // consecutive ticks. NaN load (switch cannot compute it) degrades to switch-only trust:
+         // !(NaN < threshold) passes both load tests. Advanced every tick in ALL modes so a live
+         // mode switch never resumes from stale debounce state.
+         boolean flipCondition = trusted[index] ? !(switchFiltered && !(load < loadStayThreshold))
+                                                : switchFiltered && !(load < loadEnterThreshold);
+         if (flipCondition)
```
Beomyeong, line 170: Isn't it too strick? 
What if the tick went to 39 and 0 several times in sequence, do we have back up for these cases?

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/GravityLevelingUpdater.java
```diff
@@ -0,0 +1,398 @@
...
+    * {@link #GRAVITY_REFERENCE_TIME_CONSTANT}. Since the world up-axis in body coordinates evolves as
+    * {@code ĝ⁺ = exp(−ω Δt)·ĝ ≈ ĝ − Δt (ω × ĝ)}, the gyro supplies the high-frequency content and the
+    * accelerometer pins the low-frequency level, so the reference neither lags a rotating body nor inherits the
+    * accelerometer's transient noise.</p>
+    *
+    * @param specificForceBody   the body-frame specific force a. Not modified.
+    * @param rawAngularVelocity  the RAW body-frame gyro ω (no bias correction). Not modified.
+    * @param dt                  the estimator timestep (s).
+    */
+   public void updateGravityReference(Vector3DReadOnly specificForceBody, Vector3DReadOnly rawAngularVelocity, double dt)
```
Beomyeong, line 356: updateGravityReference's guard accelNorm < 1.0e-9 doesn't catch NaN (NaN < x is false), so a single non-finite IMU sample gets blended into gravityReferenceBody and poisons it permanently (IIR never recovers), silently disabling gravity leveling.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/InvariantEKF.java
```diff
@@ -0,0 +1,357 @@
...
+    * @param contactIndex              the contact index i in [0, N).
+    * @param bodyMeasurement           the body-frame forward-kinematics measurement y = h_Cᵢ(q). Not modified.
+    * @param bodyMeasurementCovariance the body-frame measurement covariance Nᵢ = J_Cᵢ·Σ_q·J_Cᵢᵀ (3×3). Not modified.
+    */
+   public void update(int contactIndex, Tuple3DReadOnly bodyMeasurement, Matrix3DReadOnly bodyMeasurementCovariance)
+   {
+      updater.update(state, contactIndex, bodyMeasurement, bodyMeasurementCovariance, false); // learned module not wired yet
+   }
+
+   private final us.ihmc.euclid.matrix.RotationMatrix reseedRotation = new us.ihmc.euclid.matrix.RotationMatrix();
```
Beomyeong, line 152: These should be import on the top. 
We do not define the variable in this way

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/InvariantEKF.java
```diff
@@ -0,0 +1,357 @@
...
+    *         (snap avoided), for logging.
+    */
+   public double reseedContact(int contactIndex, Tuple3DReadOnly bodyMeasurement, Matrix3DReadOnly bodyMeasurementCovariance)
+   {
+      state.getRotation(reseedRotation);
+      state.getBasePosition(reseedPosition);
+      state.getContactPosition(contactIndex, reseedContactPosition);
+
+      // pre-reseed residual r = R*y - (d - p), world frame
+      us.ihmc.euclid.tuple3D.Vector3D rotated = new us.ihmc.euclid.tuple3D.Vector3D();
```
Beomyeong, line 177: No need to make new variable. 
We can rotate the reseedRotation, maybe we need to define this value as someone which includes the frame information so that we can update its frame

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/InvariantEKFStateEstimator.java
```diff
@@ -0,0 +1,1001 @@
...
+                                                                                                   .map(IMUSensorReadOnly::getSensorName)
+                                                                                                   .toList()));
+
+      contactMeasurementNoiseProvider = new ConstantContactMeasurementNoiseProvider(contactMeasurementVariance);
+
+      this.baseContactVariance = contactVariance;
+      yoContactProbability = new SideDependentList<>(new YoDouble("invariantContactProbabilityLeft", registry),
+                                                     new YoDouble("invariantContactProbabilityRight", registry));
+
+      yoContactNIS = new SideDependentList<>(new YoDouble("invariantContactNISLeft", registry),
```
Beomyeong, line 412: Below lines are repeated.

Let's replace with this

private SideDependentList<YoDouble> createSideDependentYoDouble(String namePrefix)
{
return new SideDenpedentList<>(new YoDouble(namePrifix + "Left"), new YoDouble(namePrefix + "Right", registry)
}

then call the variables in this way
yoContactCondSProxyLog10 = createSideDependentYoDouble("invariantContactCondsProxyLog10");

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/InvariantEKFStateEstimator.java
```diff
@@ -0,0 +1,1001 @@
...
+                                             new YoDouble("invariantContactNISRight", registry));
```
Beomyeong, line 413: I"ll move all the asiggnemet including logging in compute to the separate class, just to make this class simple

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/InvariantMainStateEstimator.java
```diff
@@ -0,0 +1,496 @@
...
+
+      centerOfMassUpdater = new InvariantCenterOfMassUpdater(rootJoint, centerOfMassDataHolder);
+      registry.addChild(centerOfMassUpdater.getRegistry());
+   }
+
+   @Override
+   public void doControl()
+   {
+      if (preFilter != null)
+         preFilter.computeJointState(); // phase 1: before joint outputs are consumed
```
Beomyeong, line 306: THis is called three time for the first tick.
Is this intended one?
Otherwise, we need to have initialize checker instead (prefilter != null)

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/SEK3Utils.java
```diff
@@ -0,0 +1,262 @@
+package us.ihmc.stateEstimation.invariantEstimator;
```
Beomyeong, line 1: Should it be in open-robotics ? It looks good to be with other lie group classes in Euclid

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/AlphaComplementaryPreFilter.java
```diff
@@ -0,0 +1,212 @@
+package us.ihmc.stateEstimation.jointLevel;
```
Beomyeong, line 1: Is this class for run the old estimatior with your new state estimation pipeline?

lucaslibshutz: yep, thats it!

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/invariantEstimator/SEK3Utils.java
lucaslibshutz: Dont think so, as SEK3 is InEKF specific - you really wouldnt use the group (at least I think) for anything else besides the estimator.

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointKFPrediction.java
```diff
@@ -0,0 +1,596 @@
...
+      // lambda symmetric only to round-off, and Cholesky assumes exact symmetry.
+      JointLevelKFPreFilter.symmetrize(lambda);
+      //TODO: check if this is done internally by the EJML solver
+
+      // lambda_eff = lambda + diag(rotor inertia). The rotor does not couple through the floating base, so this
+      // post-Schur diagonal add is the EXACT drivetrain term and simultaneously a principled regularizer: by
+      // Weyl it floors lambda_min(lambda_eff), which is what stops Qa blowing up as it did on Alex002.
+      for (int i = 0; i < n; i++)
+//         MatrixTools.addDiagonal(lambda, rotorInertiaDiag[i]);
+         //TODO: tests fail when addDiagonal is used - need to figure out why.
```
Beomyeong, line 481: This might be adding the input to all diagonal terms

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointKFState.java
```diff
@@ -0,0 +1,549 @@
...
+      baseIMU = pairs.isEmpty() ? null : pairs.get(0).parent;
+      baseBiasCol = baseIMU == null ? -1 : 2 * numberOfJoints + 3 * requireImuOrdinal(baseIMU);
+      if (baseIMU == null)
+         throw new RuntimeException("Base IMU is null, check the kinematic tree.");
+      LogTools.info("Base IMU initialized as " + baseIMU.getSensorName());
+
+      // 4) Precompute base->foot chains for the stance anchoring measurement update
+      if (feet != null)
+      {
+         double sigmaQdUnfiltered = parameters.sigmaQdUnfiltered.getValue();
```
Beomyeong, line 223: Is sigmaQdUnfiltered able to update in real time?

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointKFUpdate.java
```diff
@@ -0,0 +1,654 @@
...
+   {
+      this.state = state;
+      this.parameters = parameters;
+      this.dt = state.dt;
+      int n = state.numberOfJoints;
+
+      // Per-joint encoder position variance, resolved once (lookup is by joint NAME; state order thereafter).
+      // The S-pivot floor for the encoder block follows the smallest wired variance — see encVarFloorMin.
+      encVarPerJoint = new double[n];
+      double encoderVarFallback = parameters.encoderVar.getValue();
```
Beomyeong, line 179: Seems not to be as live,

ihmc-state-estimation/src/main/java/us/ihmc/stateEstimation/jointLevel/JointKFUpdate.java
```diff
@@ -0,0 +1,654 @@
...
+         this.describeRowsAsEncoder = describeRowsAsEncoder;
+      }
+
+      /** Maps the legacy label strings the package tests still pass to {@code josephUpdate}. */
+      static Channel fromLabel(String label)
+      {
+         for (Channel c : values())
+            if (c.label.equals(label))
+               return c;
+         return TEST;
```
Beomyeong: Should return exception

PR #1424, opened 2026-08-14
Change timestamp for the thread

PotatoPeeler3000:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -303,11 +303,10 @@ public void run()
 
       // Copy measured robot data to master context
       hardwareCommunicationInterface.read(masterContext.getSensorDataContext());
+      masterContext.setTimestamp(monotonicTimeProvider.getTimestamp());
```
PotatoPeeler3000, line 306: We used to only do this when the `hardwareCommunicationInterface.hasReceivedFirstState()` was true, but the yovariable server updates regardless. So the timestamp should be updated regardless as the master thread (`masterContext`) has no regard if hardware is working. This way the log has correct timestamps even if hardware is down or not communicating.

PotatoPeeler3000: So pretty much we should be setting the timestamp all the time whenever the yovariable server is running, which is what this change does.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingManager.java
```diff
@@ -333,7 +332,11 @@ private void updateYoVariableServer()
       if (yoVariableServer == null)
          return;
 
-      yoVariableServer.update(monotonicTimeProvider.getTimestamp(), rootRegistry);
+      // Reuse masterContext's timestamp (set once per cycle, before doControl() dispatches the estimator
+      // and controller) instead of re-reading the clock here. The estimator and controller copy that same
+      // masterContext timestamp into their own context, so matching it here lets their independent
+      // yoVariableServer.update() calls merge into this row instead of always writing a different ones.
+      yoVariableServer.update(masterContext.getTimestamp(), rootRegistry);
```
PotatoPeeler3000, line 339: Calling `monotonicTimeProvider.getTimestamp()` gets a new timestamp at that moment. But we want to reuse the timestamp we already asked for earlier because its still the same tick. So calling `masterContext.getTimestamp()` does this so that all the threads are using the same timestamp for the given tick.
This was causing an issue where because the old `monotonicTimeProvider.getTimestamp()` got a new timestamp the log would say its a different tick and duplicate a lot of entries as a new tick.

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -60,6 +60,7 @@
 import us.ihmc.yoVariables.registry.YoRegistry;
 import us.ihmc.yoVariables.variable.YoDouble;
 import us.ihmc.yoVariables.variable.YoEnum;
+import us.ihmc.yoVariables.variable.YoLong;
```
rjgriffin42, line 63: Looks like this may not have been needed

rjgriffin42 (approved):
If you test it on hardware and things work, I approve

ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
PotatoPeeler3000: Mhmmm good catch, left over from debugging

reesep5 (approved):

stefanfasano (approved):

PR #1425, opened 2026-08-17
Feature/standup

LuigiPenco93:
setting the trajectory duration for ground prep


ihmc-interfaces-jros2/messages/ihmc_interfaces/controller_msgs/msg/HighLevelStateMessage.msg
```diff
@@ -21,3 +21,5 @@ uint64 sequence_id
 
 # Specifies the which state the controller should transition into.
 byte high_level_controller_name 255
+# Optional spline duration for GROUND_PREP_STATE. Values <= 0 keep the controller default.
+float64 trajectory_time 0.0
```
rjgriffin42, line 26: This is an unhelpful comment, as I can't identify why I don't like this, but I don't. I also don't have a suggestion for how to do it differently.  I think it's because it's only being used for a single state, but it's encoded into this thing for all the states. 

Again, don't know how to do it differently, but thought it was worth noting, in case someone had an idea.

LuigiPenco93: changed the comment:
Optional trajectory duration for transitions into joint setpoint states (for example, stand prep or ground prep).
Values less than or equal to 0 use the controller’s default duration.

can add support for stand prep

ds58 (approved):

PR #1429, opened 2026-08-21
hand configuration for mimic

LuigiPenco93:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/kinematicsSimulation/AbilityHandKinematicsSimulation.java
```diff
@@ -2,10 +2,13 @@
 
 import static us.ihmc.handsros2.abilityHand.AbilityHand.*;
```
LuigiPenco93, line 4: edits needed to use the kin sim also with mujoco

PR #1433, opened 2026-08-26
Feature/transition to rl

reesep5:


ihmc-avatar-interfaces/src/main/java/us/ihmc/avatar/wholeBodyHardwareControl/AvatarMultiThreadingFactory.java
```diff
@@ -762,6 +762,16 @@ public void addStandPrepStateTransition(HighLevelControllerName nextControlState
       avatarControllerFactory.addCustomStateTransition(createStandTransitionState(nextControlStateEnum, avatarControllerFactory, true));
```
stefanfasano: i'd have this call the method below

stefanfasano (approved):

PR #1442, opened 2026-09-04
Remove the hard coded value assuming the pelvis height, and use the lowest food for the cells that don't have data

PotatoPeeler3000:
Screenshot from 2026-09-04 09-22-50
^^^ BUG
Screenshot from 2026-09-04 09-21-52
^^^ FIX


TomaszTB (approved):
Seems good to me. Wasn't this how it was done before some changes by @rjgriffin42? What was the reason it got switched to using the pelvis?

PotatoPeeler3000:
Well it never used to update on every update, but ya if you did reset the height it would go to the feet before

